PR #10, opened 2022-07-25
feat: add redstone node

cehali:


cehali:
Github actions for monorepo based on https://github.com/int128/typescript-actions-monorepo
Typescript tsconfig.json files based on https://dev.to/mxro/the-ultimate-guide-to-typescript-monorepos-5ap7

packages/protocol/src/data-package/SignedDataPackage.ts
```diff
@@ -14,6 +14,14 @@ import { DataPackage, DataPackagePlainObj } from "./DataPackage";
 
 export interface SignedDataPackagePlainObj extends DataPackagePlainObj {
   signature: string; // base64-encoded joined signature
+  signerAddress: string;
```
hatskier, line 17: Can we please create this extended interface with non-core data (`signerAddress`, `permawebTx`, `provider`, etc.) in the redstone-node code as it shouldn't be a part of the protocol?

packages/protocol/src/data-package/SignedDataPackage.ts
```diff
@@ -59,9 +67,12 @@ export class SignedDataPackage extends Serializable {
 
   toObj(): SignedDataPackagePlainObj {
     const signatureHex = this.serializeSignatureToHex();
+    const signerAddressAsBytes = this.recoverSignerPublicKey();
```
hatskier, line 70: Let's not recoverSignerPublicKey in this function, it is not neccessary always for serialization to plain object, but significantly consumes CPU resources

packages/protocol/src/data-point/StringDataPoint.ts
```diff
@@ -7,4 +7,11 @@ export class StringDataPoint extends DataPoint {
     const valueBytes = toUtf8Bytes(value);
     super(symbol, valueBytes);
   }
+
+  serializeToBroadcast() {
```
hatskier, line 11: Shouldn't we remove it?

hatskier (changes requested):

packages/protocol/src/data-package/SignedDataPackage.ts
cehali: Let's talk about it tomorrow

hatskier (approved):

PR #15, opened 2022-08-04
feat: remove notifier job and replace with uptime-kuma

cehali:


hatskier (approved):
Wow, what a cleaning 😄

PR #23, opened 2022-08-16
draft-cache-service-pr

hatskier:


packages/cache-service/README.md
```diff
@@ -1,5 +1,11 @@
 # redstone-cache-service
 
+## Run mongodb locally
+
+Don't forget to replace alex with your username
+
+- mongod --dbpath=/Users/alex/data/db
```
cehali, line 7: Maybe delete comment and change `/Users/alex/data/db` to `{path-to-mongodb}`

packages/cache-service/package.json
```diff
@@ -8,7 +8,7 @@
   "license": "MIT",
   "scripts": {
     "prebuild": "rimraf dist",
-    "build": "nest build",
+    "build": "tsc --build && nest build",
```
cehali, line 11: `tsc --build` shouldn't be here. Nest have TS support so something must be wrong here.

packages/cache-service/src/config.ts
```diff
@@ -0,0 +1,14 @@
...
+    throw new Error(`Env variable not found: ${envName}`);
+  }
+  return process.env[envName] as string;
+};
+
+export default {
+  mongoDbUrl: getEnv("MONGO_DB_URL"),
+  enableStreamrListening: getEnv("ENABLE_STREAMR_LISTENING") === "true",
+  streamrUnsubscribeDelayMilliseconds: 120 * 1000,
+};
```
cehali, line 14: I think we should use nest built in configuartion for env variables: https://docs.nestjs.com/techniques/configuration

packages/cache-service/src/data-packages/data-packages.model.ts
```diff
@@ -0,0 +1,38 @@
+import mongoose from "mongoose";
```
cehali, line 1: Nest has built-in support for mongo: https://docs.nestjs.com/techniques/mongodb

packages/cache-service/src/shared/db.ts
```diff
@@ -0,0 +1,32 @@
+import mongoose from "mongoose";
```
cehali, line 1: The same as above. Nest has support for mongo: https://docs.nestjs.com/techniques/mongodb.
This will make easier to handle mongo connection

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
```diff
@@ -0,0 +1,91 @@
...
+    )) {
+      const nodeIsRegistered = nodeEvmAddresses.some(
+        (address) =>
+          address.toLowerCase() === subscribedNodeEvmAddress.toLowerCase()
+      );
+      if (!nodeIsRegistered) {
+        this.cancelListeningToNodeStream(subscribedNodeEvmAddress);
+      }
+    }
+  }
```
cehali, line 53: I am not sure but don't we have a concurrency problem here? If node address is not in `subscriptionState` `listenToNodeStream` function will be called but without any waiting and it is possible that `cancelListeningToNodeStream` will be called for the same node address

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  cancelListeningToNodeStream(nodeEvmAddress: string) {
+    const subscription = this.subscriptionsState[nodeEvmAddress];
+    if (!!subscription) {
+      delete this.subscriptionsState[nodeEvmAddress];
+      setTimeout(
+        () => this.streamrClient.unsubscribe(subscription),
+        config.streamrUnsubscribeDelayMilliseconds
+      );
```
cehali, line 88: Why `setTimeout` here?

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -1,19 +1,56 @@
...
+    };
+
+    return await this.dataPackagesService.getDataPackages(requestParams);
   }
 
   @Post("bulk")
-  async addBulk(): Promise<string> {
-    return "TODO: implement. Will add many data packages";
-  }
+  async addBulk(@Body() body: BulkPostRequestBody) {
```
cehali, line 43: If we don't have `addOne` maybe we should drop naming it bulk and many data packages

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -1,8 +1,124 @@
...
 
 @Injectable()
 export class DataPackagesService {
-  getHello(): string {
-    return "Hello World! I was updated :)";
+  async saveManyDataPackagesInDB(dataPackages: CachedDataPackage[]) {
+    await DataPackage.insertMany(dataPackages);
+  }
+
+  // TODO: try to replace current implementation using only one aggregation call
```
cehali, line 29: It is done already?

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -1,8 +1,124 @@
...
+  async getDataPackages(
+    requestConfig: DataPackagesRequestParams
+  ): Promise<DataPackagesResponse> {
+    const fetchedPackagesPerDataFeed: {
+      [dataFeedId: string]: CachedDataPackage[];
+    } = {};
+
+    const getDataPackagesForDataFeed = async (dataFeedId: string) => {
```
cehali, line 37: `DataFeed` -> `DataService` ?

packages/cache-service/package.json
hatskier: Yes, but nest build doesn't work correctly with monorepos and composite building

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
hatskier: Hmm, I don't really see how it would be possible. If `listenToNodeStream` is called then the address is included in `nodeEvmAddresses`, which means that `nodeIsRegistered` variable will be `true`, and `cancelListeningToNodeStream` will not be called

packages/cache-service/package.json
cehali: There is some monorepo option: https://docs.nestjs.com/cli/monorepo#monorepo-mode. I have never used it but maybe it could helpful

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
hatskier: To handle few latest data packages for the removed node. But actually, it doesn't have any added value. I will remove it :)

packages/cache-service/src/data-packages/data-packages.controller.ts
hatskier: We don't have it now, but I would prefer to stay with the current name, because in REST api standard `POST /some-resources` usually mean adding only one instance of the resource

packages/cache-service/src/data-packages/data-packages.service.ts
hatskier: Nope, we still have dataFeed concept, but it has a different meaning (e.g. ETH is data feed id)

packages/cache-service/src/config.ts
hatskier: We do some magic in app.module (e.g. conditional ScheduleModule importing_ based on env variables, so the standard nest approach would not be a good fit for us here

PR #28, opened 2022-08-20
feat: prepare oracle node for the new ecosystem

hatskier:


packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -1,10 +1,5 @@
 import { Injectable } from "@nestjs/common";
-import {
-  computeAddress,
-  keccak256,
-  recoverPublicKey,
-  toUtf8Bytes,
-} from "ethers/lib/utils";
+import { RequestSigner } from "redstone-protocol";
```
cehali, line 2: I am wondering if `RequestSigner` is a good name, specially the request word. A request is when someone requests from me or I am requested to do something. Here is more like without any request, nodes want to broadcast to cache. Maybe something with broadcast would be better/

packages/oracle-node/src/NodeRunner.ts
```diff
@@ -277,6 +269,7 @@ export default class NodeRunner {
         enableStreamrBroadcaster &&
         !disableSinglePricesBroadcastingInStreamr
```
cehali, line 270: Maybe we want to make it default false and delete it from manifest?

packages/oracle-node/src/broadcasters/http/HttpBroadcaster.ts
```diff
@@ -24,30 +36,4 @@ export class HttpBroadcaster implements Broadcaster {
 
     await Promise.allSettled(promises);
   }
-
-  async broadcastPricePackage(
```
cehali, line 28: Do we want to delete the broadcasting price package? Or we implement it in another way?

packages/oracle-node/src/signers/EvmPriceSigner.ts
```diff
@@ -1,161 +1,23 @@
...
-];
-
-const serializePriceValue = (value: number) => Math.round(value * 10 ** 8);
+  NumericDataPoint,
+  DataPackage,
+  SignedDataPackage,
+} from "redstone-protocol";
+import { PricePackage } from "../types";
 
 export default class EvmPriceSigner {
```
cehali, line 8: I wondering if we could delete it completely without breaking old nodes

cehali, line 8: Do we want to keep it as a separate service? It seems that it could be small util if most of the functionality is in the protocol.

packages/oracle-node/src/broadcasters/http/HttpBroadcaster.ts
cehali: I believe I understand how it works, it is covered in StreamrBrodcaster. But don't we want it also in direct caches?

packages/cache-service/src/data-packages/data-packages.service.ts
hatskier: Renamed to UniversalSigner

cehali (approved):

PR #31, opened 2022-08-25
feat: add sAVAX token

cehali:


hatskier (approved):
Good job! Please test if all the previous avlanche data feeds work correctly

PR #34, opened 2022-08-26
feat: add sav2 token

cehali:


packages/oracle-node/src/fetchers/evm-chain/AvalancheEvmFetcher.ts
```diff
@@ -56,37 +62,58 @@ export class AvalancheEvmFetcher extends BaseFetcher {
...
-      switch (id) {
-        case "YYAV3SA1": {
-          const price = await this.extractPriceForYieldYak(response);
-          pricesObject[id] = Number(price);
-          break;
-        }
-        default:
-          throw new Error("Invalid id for Avalanche EVM fetcher");
-      }
+      const price = await this.extractPriceForYieldYak(response, id);
```
hatskier, line 65: Note for future: let's use `await Promise.all` to make requests at the same time instead of sequential requests

hatskier (approved):

PR #41, opened 2022-09-05
feat: arweave backup feature implemented in cache-service

hatskier:


packages/cache-service/src/streamr-listener/streamr-listener.service.ts
```diff
@@ -65,12 +70,17 @@ export class StreamrListenerService {
...
           console.log(
             `Saved ${dataPackagesToSave.length} data packages for node: ${nodeEvmAddress}`
           );
+
+          if (config.enableArchivingOnArweave) {
+            await this.bundlrService.saveDataPackages(dataPackagesToSave);
+          }
         } catch (e) {
-          console.error(e);
+          console.error("Error occured ", e);
```
cehali, line 83: Maybe `error.stack` or `error.message`?

packages/cache-service/src/bundlr/bundlr.service.ts
```diff
@@ -0,0 +1,61 @@
...
+@Injectable()
+export class BundlrService {
+  private bundlrClient: Bundlr;
+
+  constructor() {
+    if (config.enableArchivingOnArweave) {
+      this.bundlrClient = new Bundlr(
+        config.bundlrNodeUrl!,
+        "arweave",
+        config.arweaveJwkKey!
```
cehali, line 17: We should think about solutions to not use exclamation when using config

ppedziwiatr: https://frontlive.pl/blog/typescript-modelowanie-domeny illegal states?

cehali (approved):

PR #80, opened 2022-10-12
fix: fixes after the evm-connector audit by ABDK

hatskier:


packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -11,23 +13,39 @@ import "./RedstoneConstants.sol";
  * and the ProxyConnector contracts
  */
 contract CalldataExtractor is RedstoneConstants {
+  using SafeMath for uint256;
+
   function _extractByteSizeOfUnsignedMetadata() internal pure returns (uint256) {
     // Using uint24, because unsigned metadata byte size number has 3 bytes
     uint24 unsignedMetadataByteSize;
+    uint256 calldataSize = msg.data.length;
```
3sGgpQ8H, line 21: This variable is redundant, as access to `msg.data.length` is cheaper than access to a local variable.

packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -11,23 +13,39 @@ import "./RedstoneConstants.sol";
...
+    require(
+      REDSTONE_MARKER_BS + STANDARD_SLOT_BS <= calldataSize,
```
3sGgpQ8H, line 23: Should be `REDSTONE_MARKER_BS_PLUS_STANDARD_SLOT_BS` instead pf `REDSTONE_MARKER_BS + STANDARD_SLOT_BS`.

packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -11,23 +13,39 @@ import "./RedstoneConstants.sol";
...
-    returns (uint256)
+    returns (uint16 dataPackagesCount)
   {
-    // Using uint16, because unsigned metadata byte size number has 2 bytes
-    uint16 dataPackagesCount;
+    require(calldataNegativeOffset + STANDARD_SLOT_BS <= msg.data.length,
+      ERR_CALLDATA_OVER_OR_UNDER_FLOW);
     assembly {
       let calldataOffset := sub(calldatasize(), calldataNegativeOffset)
       dataPackagesCount := calldataload(sub(calldataOffset, STANDARD_SLOT_BS))
```
3sGgpQ8H, line 51: The expression `calldataNegativeOffset + STANDARD_SLOT_BS` is implicitly calculated twice.  Consider calculating once and reusing.

packages/evm-connector/contracts/core/RedstoneDefaultsLib.sol
```diff
@@ -9,28 +9,31 @@ import "../libs/NumericArrayLib.sol";
...
-  uint256 constant DEFAULT_MAX_DATA_TIMESTAMP_DELAY_IN_SECONDS = 3 * 60;
-  uint256 constant DEFAULT_MAX_DATA_TIMESTAMP_AHEAD_IN_SECONDS = 60;
+  uint256 constant DEFAULT_MAX_DATA_TIMESTAMP_DELAY_SECONDS = 3 minutes;
+  uint256 constant DEFAULT_MAX_DATA_TIMESTAMP_AHEAD_SECONDS = 1 minutes;
 
-  function isTimestampValid(uint256 receivedTimestamp) internal view returns (bool) {
+  string internal constant ERR_TIMESTAMP_FROM_TOO_LONG_FUTURE = "Data with too future timestamps not allowed";
+  string internal constant ERR_TIMESTAMP_IS_TOO_OLD = "Timestamp is too old";
+
+  function validateTimestamp(uint256 receivedTimestampMilliseconds) internal view {
```
3sGgpQ8H, line 18: Ethereum don't have access to time with millisecond precision.  Using milliseconds looks like unnecessary overcomplication.  Consider using seconds instead.

packages/evm-connector/contracts/samples/SampleProxyConnector.sol
```diff
@@ -10,6 +10,12 @@ import "./SampleRedstoneConsumerNumericMock.sol";
  * @dev An example of a contract that makes a call to a SampleRedstoneConsumerMock contract
  */
 contract SampleProxyConnector is ProxyConnector {
+  string internal constant ERR_UNEXPECTED_ORACLE_VALUE = "Received an unexpected oracle value";
+  string internal constant ERR_WRONG_VALUE = "Wrong value!";
+  string internal constant ERR_EXPECTED_MSG_VALUE_TO_BE_PASSED = "Expected msg.value to be passed";
+  string internal constant ERR_EXPECTED_MSG_VALUE_NOT_TO_BE_PASSED =
+    "Expected msg.value not to be passed";
```
3sGgpQ8H, line 17: String error message are inefficient.  Consider using named errors.

packages/evm-connector/contracts/samples/Benchmark.sol
```diff
@@ -1,15 +1,20 @@
...
 import "../mocks/RedstoneConsumerNumericMock.sol";
 
 contract Benchmark is RedstoneConsumerNumericMock {
   // We use this storage variable to avoid annoying compilation
   // warnings about function state mutability
   uint256 someStorageVar = 0;
+  uint8 uniqueSignersThreshold = 1;
 
-  function updateUniqueSignersThreshold(uint256 newUniqueSignersThreshold) external {
+  function getUniqueSignersThreshold() public view virtual override returns (uint8) {
```
3sGgpQ8H, line 13: This getter function is declared as virtual, while the corresponding setter function is not virtual.  Consider declaring this getter as not virtual for consistency.

packages/evm-connector/contracts/mocks/RedstoneConsumerBytesMock.sol
```diff
@@ -1,32 +1,28 @@
...
-  function isTimestampValid(uint256 _receivedTimestamp)
-    public
-    view
-    virtual
-    override
-    returns (bool)
-  {
-    return _receivedTimestamp >= 1654353400000;
+  function validateTimestamp(uint256 receivedTimestampMilliseconds) public view virtual override {
+    require(receivedTimestampMilliseconds >= 1654353400000, ERR_TIMESTAMP_IS_NOT_VALID);
```
3sGgpQ8H, line 26: The maximum valid timestamp value should be a named constant.

packages/evm-connector/contracts/mocks/RedstoneConsumerBytesMock.sol
```diff
@@ -1,32 +1,28 @@
...
-pragma solidity ^0.8.4;
+pragma solidity ^0.8.0;
 
 import "../core/RedstoneConsumerBytesBase.sol";
 import "./AuthorisedMockSignersBase.sol";
 
 contract RedstoneConsumerBytesMock is RedstoneConsumerBytesBase, AuthorisedMockSignersBase {
-  constructor() {
-    uniqueSignersThreshold = 3;
+  string internal constant ERR_TIMESTAMP_IS_NOT_VALID = "Timestamp is not valid";
```
3sGgpQ8H, line 9: String error messages are suboptimal.  Consider using named errors instead.

packages/evm-connector/contracts/mocks/RedstoneConsumerNumericMock.sol
```diff
@@ -1,32 +1,28 @@
...
-pragma solidity ^0.8.4;
+pragma solidity ^0.8.0;
 
 import "../core/RedstoneConsumerNumericBase.sol";
 import "./AuthorisedMockSignersBase.sol";
 
 contract RedstoneConsumerNumericMock is RedstoneConsumerNumericBase, AuthorisedMockSignersBase {
-  constructor() {
-    uniqueSignersThreshold = 10;
+  string internal constant ERR_TIMESTAMP_IS_NOT_VALID = "Timestamp is not valid";
```
3sGgpQ8H, line 9: String error messages are suboptimal.  Consider using named error instead.

packages/evm-connector/contracts/mocks/RedstoneConsumerNumericMock.sol
```diff
@@ -1,32 +1,28 @@
...
-  function isTimestampValid(uint256 _receivedTimestamp)
-    public
-    view
-    virtual
-    override
-    returns (bool)
-  {
-    return _receivedTimestamp >= 1654353400000;
+  function validateTimestamp(uint256 receivedTimestampMilliseconds) public view virtual override {
+    require(receivedTimestampMilliseconds >= 1654353400000, ERR_TIMESTAMP_IS_NOT_VALID);
```
3sGgpQ8H, line 26: The maximum valid timestamp value should be a named constant.

packages/evm-connector/contracts/mocks/AuthorisedMockSignersBase.sol
```diff
@@ -1,71 +1,71 @@
...
+    } else if (signerAddress == 0x2546BcD3c84621e976D8185a91A922aE77ECEc30) {
       return 16;
-    } else if (_signerAddress == 0xbDA5747bFD65F08deb54cb465eB87D40e51B197E) {
+    } else if (signerAddress == 0xbDA5747bFD65F08deb54cb465eB87D40e51B197E) {
       return 17;
-    } else if (_signerAddress == 0xdD2FD4581271e230360230F9337D5c0430Bf44C0) {
+    } else if (signerAddress == 0xdD2FD4581271e230360230F9337D5c0430Bf44C0) {
       return 18;
-    } else if (_signerAddress == 0x8626f6940E2eb28930eFb4CeF49B2d1F2C9C1199) {
+    } else if (signerAddress == 0x8626f6940E2eb28930eFb4CeF49B2d1F2C9C1199) {
```
3sGgpQ8H, line 68: A linear chain of conditional statements is suboptimal.  Consider using the bisection method like this:

    if (signerAddress < 0x...) {
      // Handle lower addresses recursively
    } else {
      // Handle upper addresses recursively
    }

packages/evm-connector/contracts/libs/NumericArrayLib.sol
```diff
@@ -14,9 +11,11 @@ library NumericArrayLib {
   uint256 constant BYTES_ARR_LEN_VAR_BS = 32;
   uint256 constant UINT256_VALUE_BS = 32;
 
-  // This functions modifies the array
+  string internal constant ERR_MEDIAN_IN_EMPTY_ARRAY = "Can't pick a median of an empty array";
```
3sGgpQ8H, line 14: String error messages are suboptimal.  Consider using named errors instead.

packages/evm-connector/contracts/libs/NumericArrayLib.sol
```diff
@@ -87,13 +57,14 @@ library NumericArrayLib {
...
+  // We've commented the unused code below to optimise gas cost for deployment
+  // function solidityBubbleSort(uint256[] memory arr) internal pure {
+  //   for (uint256 i = 0; i < arr.length; i++) {
+  //     for (uint256 j = 0; j < i; j++) {
+  //       if (arr[i] < arr[j]) {
+  //         (arr[i], arr[j]) = (arr[j], arr[i]);
+  //       }
+  //     }
+  //   }
+  // }
```
3sGgpQ8H, line 69: This commented code should be removed.

packages/evm-connector/contracts/core/RedstoneDefaultsLib.sol
hatskier: We have milliseconds in our protocol as we will also launch on other chains, which may support millisecond precision

packages/evm-connector/contracts/mocks/AuthorisedMockSignersBase.sol
hatskier: This is a very interesting idea. The only disadvantage I see is that it will be a bit more difficult to add new signers and  that it will require more testing to ensure the bisection is correct and we don't have an address that can't be reached. Also, we call the `getAuthorisedSignerIndex` only once for each data package so the gas gains would not be very high.
@jakub-wojciechowski what do you think?

packages/evm-connector/contracts/samples/SampleProxyConnector.sol
hatskier: We have considered using named errors, but decided to stay with the "old" way of reverting to make it easier for consumers to integrate with the older version of compilers (as the named errors are supported only starting from 0.8.4)

packages/evm-connector/contracts/mocks/AuthorisedMockSignersBase.sol
jakub-wojciechowski: Because of the gas costs we don't expect more than a few and around ~10 signers in the most complex scenarios. I'm afraid at this size of n, the gas savings from logarithmic complexity won't be minimal if any, but the solution will be less readable.

packages/evm-connector/contracts/libs/NumericArrayLib.sol
jakub-wojciechowski: We've got some issues with this approach before (errors haven't been properly propagated in delegated calls). Also, the testing tools (like hardhat) lacked full support. We're going to revisit this and if it works correctly

packages/evm-connector/contracts/samples/SampleProxyConnector.sol
hatskier: After internal discussion we decided to add the named errors :)

packages/evm-connector/contracts/libs/NumericArrayLib.sol
hatskier: After internal discussion and gas benchmarking, we decided to implement named errors

jakub-wojciechowski (approved):

PR #83, opened 2022-10-13
feat: implemented required functions in redstone-sdk 

hatskier:


packages/sdk/package.json
```diff
@@ -11,12 +11,15 @@
   },
   "dependencies": {
     "axios": "^0.27.2",
+    "bluebird": "^3.7.2",
```
cehali, line 14: I don't think we should use `bluebird` 

image

packages/sdk/src/index.ts
```diff
@@ -1,6 +1,11 @@
...
 import axios from "axios";
+import bluebird, { AggregateError } from "bluebird";
+import { RedstoneOraclesState } from "redstone-oracles-smartweave-contracts/src/contracts/redstone-oracle-registry/types";
+import redstoneOraclesInitialState from "redstone-oracles-smartweave-contracts/src/contracts/redstone-oracle-registry/initial-state.json";
+import {
+  SignedDataPackage,
+  SignedDataPackagePlainObj,
+} from "redstone-protocol";
 
 export const DEFAULT_CACHE_SERVICE_URLS = [
```
cehali, line 10: Maybe update those urls? They seem to be invalid for now.

packages/sdk/src/index.ts
```diff
@@ -16,22 +21,15 @@ export interface DataPackagesRequestParams {
...
 export const getOracleRegistryState =
   async (): Promise<RedstoneOraclesState> => {
-    return {
-      dataServices: {},
-      nodes: {},
-      contractAdmins: ["hahah2"],
-      canEvolve: true,
-      evolve: null,
-    };
+    return redstoneOraclesInitialState;
```
cehali, line 30: Maybe we should think about moving it out of Arweave? We don't update it very often and current state is different than initial on Arweave and also  state on Arweave is different than real state.

packages/sdk/package.json
```diff
@@ -11,12 +11,15 @@
...
     "redstone-oracles-smartweave-contracts": "1.0.0",
     "redstone-protocol": "1.0.0"
   },
   "devDependencies": {
+    "@types/bluebird": "^3.5.37",
     "@types/jest": "^27.5.2",
     "jest": "^28.1.0",
+    "mws": "^2.0.8",
```
cehali, line 22: Good choice 👍

packages/sdk/src/index.ts
```diff
@@ -44,23 +42,56 @@ export const getDataServiceIdForSigner = (
...
+      params: {
+        "data-service-id": reqParams.dataServiceId,
+        "unique-signers-count": reqParams.uniqueSignersCount,
+        "data-feeds": reqParams.dataFeeds.join(","),
+      },
+    })
+  );
+
+  try {
+    const response = await bluebird.Promise.any(promises);
```
cehali, line 86: Any function is native: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any

packages/sdk/src/index.ts
```diff
@@ -44,23 +42,56 @@ export const getDataServiceIdForSigner = (
...
 
-// TODO: implement
-// This function will simply proxy requests to
-// the requested cache services (given urls)
-// And will return the first valid response
+const parseDataPackagesResponse = (dpResponse: {
+  [dataFeedId: string]: SignedDataPackagePlainObj[];
+}): DataPackagesResponse => {
+  const parsedResponse: DataPackagesResponse = {};
+  for (const dataFeedId of Object.keys(dpResponse)) {
```
cehali, line 49: `js
const [dataFeedId, dataPoint] of Object.entries(dpResponse)
`

packages/sdk/src/index.ts
```diff
@@ -44,23 +42,56 @@ export const getDataServiceIdForSigner = (
...
+    parsedResponse[dataFeedId] = dpResponse[dataFeedId].map(
+      (dataPackage: SignedDataPackagePlainObj) =>
+        SignedDataPackage.fromObj(dataPackage)
+    );
+  }
+  return parsedResponse;
```
cehali, line 55: We have similar code in different places. I think that when we need to take care of performance `reduce` or `map` and `reduce` could be faster.

packages/sdk/src/index.ts
hatskier: Let's do it in a separate PR

packages/sdk/src/index.ts
hatskier: I don't think map + reduce has better performance than for loops, and IMO for loops are more readable (but it's probably a matter of taste)
https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/

packages/sdk/src/index.ts
hatskier: But it doesn't work properly with Node.js 16 with typescript

cehali (approved):

PR #85, opened 2022-10-18
feat: add score by address route

cehali:


packages/oracle-node/src/routes/score-by-address.ts
```diff
@@ -0,0 +1,146 @@
...
+  return fetchedData as number;
+};
+
+const getSignedDataPackage = async (
+  address: string,
+  timestamp: number,
+  privateKey: string
+) => {
+  const score = await fetchData(address);
+  const scoreData = { symbol: address, value: score };
```
hatskier, line 137: Do we really need the `scoreData` variable? Can't we just pass address and score directly?

packages/oracle-node/src/routes/score-by-address.ts
```diff
@@ -0,0 +1,146 @@
...
+  currentTimestamp: number
+) => {
+  const timestampDifference = currentTimestamp - timestampFromMessage;
+  return (
+    timestampDifference > 0 &&
+    timestampDifference <= TEN_MINUTES_IN_MILLISECONDS
+  );
+};
+
+const fetchData = async (address: string): Promise<number> => {
```
hatskier, line 119: Let's change the `fetchData` function name to something more describing, e.g. `fetchScoreForAddress`

packages/oracle-node/src/routes/score-by-address.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+  });
+  return clearInterval;
+};
+
+const verifyPayload = (params: Payload, limitedAddresses: LimitedAddress[]) => {
+  if (!(params.timestamp && params.signature)) {
+    throw new Error("Invalid request, missing parameter");
+  }
+  const { timestamp, signature } = params;
```
hatskier, line 75: Let's move this line to the top to replace `params.timestamp && params.signature` with `timestamp && signature`

packages/oracle-node/src/routes/score-by-address.ts
```diff
@@ -0,0 +1,146 @@
...
+  const addressFromMessage = utils.verifyMessage(timestamp, signature);
+  const isAddressInLimited = checkIfAddressIsLimited(
```
hatskier, line 77: Do we also want to add some kind of address activity check? E.g. if it had any transactions on some blockchain

packages/oracle-node/src/routes/services/LimitedAddressService.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface LimitedAddress {
+  address: string;
+  timestamp: number;
+}
+
+const ONE_MINUTE_IN_MILLISECONDS = 60 * 1000;
+const ONE_HOUR_IN_MILLISECONDS = 60 * 60 * 1000;
+
+export const buildLimitedAddressesService = () => {
+  let limitedAddresses: LimitedAddress[] = [];
```
hatskier, line 14: What do you think about changing this array to an object `{ [address: string]: {numberOfRequests: number, lastRequestTimestamp: number} }`, so that we can allow few requests (e.g. 10) for the same address during one hour? I am a bit afraid of rate limiting to 1 request per hour, as it can make quite poor UX for some cases

hatskier (changes requested):

packages/oracle-node/src/routes/score-by-address.ts
cehali: Here we only want a rate limit. I would place checking address activity as fetching score. I think this is only a boilerplate which can be extended for the score/binary score of the activity or a mix of both.

packages/oracle-node/src/routes/services/LimitedAddressService.ts
cehali: Good idea 👍

packages/oracle-node/src/routes/services/LimitedAddressesService.ts
```diff
@@ -0,0 +1,65 @@
...
+export const getLimitedAddresses = () => limitedAddresses;
+
+export const checkIsAddressWithoutAccess = (addressFromMessage: string) =>
+  Object.entries(limitedAddresses).some(
+    ([address, addressStats]) =>
+      address === addressFromMessage &&
+      addressStats.numberOfRequests > MAX_NUMBER_OF_REQUESTS
+  );
+
+export const pushToLimitedAddresses = (
```
hatskier, line 49: Can we change the naming a bit to make the code more readable? I would propose the following changes:
- LimitedAddressService -> RateLimitingService
- checkIsAddressWithoutAccess -> hasAddressReachedRateLimit
- pushToLimitedAddresses -> recordRequestSentByAddress

packages/oracle-node/src/routes/services/LimitedAddressesService.ts
```diff
@@ -0,0 +1,65 @@
...
+const logger = require("../../utils/logger")("score-by-address") as Consola;
+
+export interface LimitedAddresses {
+  [address: string]: { numberOfRequests: number; lastRequestTimestamp: number };
+}
+
+const ONE_MINUTE_IN_MILLISECONDS = 60 * 1000;
+const ONE_HOUR_IN_MILLISECONDS = 60 * 60 * 1000;
+
+const MAX_NUMBER_OF_REQUESTS = 1;
```
hatskier, line 12: Can we please set it to a number bigger than one and cover it with tests. I am a bit afraid of a poor UX with 1 request alloowed per hour. I would rather set 5 or 10

packages/oracle-node/src/routes/services/LimitedAddressesService.ts
```diff
@@ -0,0 +1,65 @@
...
+let limitedAddresses: LimitedAddresses = {};
+
+export const startCheckingLimitedAddresses = () => {
+  const intervalTimer = setInterval(() => {
+    const oldLimitedAddressesLength = Object.keys(limitedAddresses).length;
+    logger.info(
+      `Updating limited addresses. Current number of addresses: ${oldLimitedAddressesLength}`
+    );
+    const currentTimestamp = Date.now();
+    const filteredLimitedAddressesArray = Object.entries(
```
hatskier, line 23: Wouldn't it be more efficient to just iterate over the object keys and using `delete` operation when needed?

packages/oracle-node/src/routes/services/LimitedAddressesService.ts
```diff
@@ -0,0 +1,65 @@
+import { Consola } from "consola";
+
+const logger = require("../../utils/logger")("score-by-address") as Consola;
+
+export interface LimitedAddresses {
+  [address: string]: { numberOfRequests: number; lastRequestTimestamp: number };
```
hatskier, line 6: Sorry for changing the idea again, but what do you think about storing a list of timestamps (`number[]`) instead of `{ numberOfRequests: number; lastRequestTimestamp: number }`? Otherwise our rate limiter will work incorrectly, e.g. if you make 4 requests from 00:00 till 01:00 and then one request at 3:00, then you'll be rate limited till 4:00

hatskier (changes requested):

packages/oracle-node/src/routes/score-by-address.ts
```diff
@@ -0,0 +1,120 @@
...
+      registerRequestAndVerifyAddress(currentTimestamp, address);
+
+      const signedDataPackage = await getSignedDataPackage(
+        address,
+        currentTimestamp,
+        nodeConfig.privateKeys.ethereumPrivateKey
+      );
+
+      return res.json({
+        dataPoints: signedDataPackage.dataPackage.dataPoints.map((dataPoint) =>
```
hatskier, line 45: What do you think about using the following code instead?
```js
return res.json(signedDataPackage.toObj());
```

hatskier (changes requested):

packages/oracle-node/src/routes/score-by-address.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  return currentTimestamp;
+};
+
+const registerRequestAndVerifyAddress = (
+  timestamp: number,
+  address: string
+) => {
+  recordRequestSentByAddress(address, timestamp);
```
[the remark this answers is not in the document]
cehali: Now when I am thinking about it could provide to the situation when someone has no access at all if constantly requesting after requesting more than max times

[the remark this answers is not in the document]
cehali: Now when I am thinking about it could provide to the situation when someone has no access at all if constantly requesting after requesting more than max times

packages/oracle-node/src/routes/services/RateLimitingService.ts
```diff
@@ -0,0 +1,47 @@
...
+let recordedAddresses: RecordedAddresses = {};
+
+export const getRecordedAddresses = () => recordedAddresses;
+
+export const recordRequestSentByAddress = (
+  address: string,
+  timestamp: number
+) => {
+  logger.info(`Recorded request from ${address}`);
+  const addressTimestamps = recordedAddresses[address]?.timestamps ?? [];
```
hatskier, line 21: What do you think about replacing this code 
```js
const addressTimestamps = recordedAddresses[address]?.timestamps ?? [];
recordedAddresses[address] = {
  timestamps: [...addressTimestamps, timestamp],
};
```
with this code?
```js
if (!recordedAddresses[address]?.timestamps) {
  recordedAddresses[address] = {timestamps: []};
}
recordedAddresses[address]!.timestamps.push(timestamp);
```

to improve the complexity (O(1) instead of O(N))

packages/oracle-node/src/routes/score-by-address.ts
```diff
@@ -0,0 +1,111 @@
...
+    throw new Error("Invalid timestamp, it should be less than 10 minutes ago");
+  }
+
+  return currentTimestamp;
+};
+
+const verifyAddress = (address: string) => {
+  const isAddressWithoutAccess = hasAddressReachedRateLimit(address);
+  if (isAddressWithoutAccess) {
+    throw new Error(`Address ${address} checked its score in the last hour`);
```
hatskier, line 81: Can we please change the error message to a more generic one as we can use the rate limiting service later for other use cases? e.g.: "Address ${address} reached rate limit. Please try in an hour"

hatskier, line 81: Can we please change the error message to a more generic one as we can use the rate limiting service later for other use cases? e.g.: "Address ${address} reached rate limit. Please try in an hour"

hatskier (approved):

PR #86, opened 2022-10-19
chore: remove unnecessary dependency and env variable

cehali:


packages/oracle-node/src/NodeRunner.ts
```diff
@@ -296,13 +296,7 @@ export default class NodeRunner {
       promises.push(this.httpBroadcaster.broadcast(signedDataPackages));
       const enableStreamrBroadcaster =
         this.currentManifest?.enableStreamrBroadcaster ?? false;
```
hatskier, line 298: What do you think about replacing `this.currentManifest?.enableStreamrBroadcaster ?? false` with `!!this.currentManifest?.enableStreamrBroadcaster`?

hatskier (approved):

PR #94, opened 2022-10-25
feat: prepare evm-connector for npm publishing

hatskier:


cehali (approved):

packages/evm-connector/README.md
```diff
@@ -59,7 +59,7 @@ You need to apply a minium change to the source code to enable smart contract to
 We strongly recommend you to have some upgradability mechanism for your contracts (it can be based on multisig, DAO, or anything else). This way, you can quickly swtich to the latest trusted data providers in case of changes or problems with the current providers.
 
 ```js
-import "redstone-evm-connector/lib/contracts/data-services/AvalancheDataServiceConsumerBase.sol";
+import "@redstone-finance/evm-connector/dist/contracts/data-services/AvalancheDataServiceConsumerBase.sol";
```
cehali, line 62: Maybe export all contracts in the way using them doesn't require to import them from `dist` folder?

packages/sdk/package.json
```diff
@@ -1,16 +1,20 @@
...
-    "build": "rm -rf dist && tsc --build"
+    "build": "rm -rf dist && tsc --build",
+    "preversion": "yarn build",
+    "postversion": "git commit -a -m \"build: $npm_package_name@$npm_package_version\" && git push"
   },
   "dependencies": {
-    "axios": "^0.27.2"
+    "axios": "^0.27.2",
+    "redstone-oracles-smartweave-contracts": "latest",
+    "redstone-protocol": "latest"
```
cehali, line 17: Maybe not in this PR but I wonder if this is not a problem that `evm-connector` depends on `redstone-protocol`, `redstone-sdk` and `redstone-sdk` depends on `redstone-protocol`

cehali (approved):

packages/sdk/package.json
hatskier: This is not a big problem, but let's try to finid a time later and simplify (or ideally remove) internal dependencies later. But maybe in a separate PR :)

PR #103, opened 2022-10-30
feat: prepared deployment scripts for token, locking, vesting contracts

hatskier:


packages/eth-contracts/README.md
```diff
@@ -5,6 +5,12 @@
 
 RedStone main business contracts
 
+## Deployment
+
+- Create a `.env` file with your PRIVATE_KEY and ALCHEMY_API_KEY
```
cehali, line 10: Maybe instead of `ALCHEMY_API_KEY` we should have `RPC_URL` or `NETWORK_ENDPOINT_URL` and be able to operate not only on alchemy

packages/eth-contracts/src/deploy-all.ts
```diff
@@ -0,0 +1,160 @@
...
+  );
+  const locking = await upgrades.deployProxy(LockingRegistryFactory, [
+    tokenAddress,
+    AUTHORISED_SLASHER,
+    DELAY_FOR_UNLOCKING_SECONDS,
+  ]);
+  return locking.address;
+}
+
+async function prepareAllVestingContracts(
```
cehali, line 79: Split the content of this function into three functions. One for deploying, one for transferring tokens, and the main one with the loop that runs two other tasks.

packages/eth-contracts/README.md
hatskier: Actually, it's a good idea!

cehali (approved):

PR #113, opened 2022-11-04
feat: update oracle monitoring ecr deploy script

cehali:


.github/workflows/oracle-monitoring-deploy-to-ecr-public.yml
```diff
@@ -3,15 +3,15 @@
...
 # Learn more: https://github.com/pahud/ecr-public-action
 
-name: Deploy to Amazon ECR Public
+name: Deploy monitoring to Amazon ECR Public
 on:
   push:
-    branches:
-      - main
-      - push-docker-ecr-public
+    # branches:
```
hatskier, line 9: Let's implement it in the same way as we have in `deploy-to-ecr-public.yml`. Otherwise it would consume too much GH resources

cehali: It is only temporary, not to merge each time to main in order to deploy to ecr

hatskier: In the `deploy-to-ecr-public.yml` you also don't need to merge it to main all the time. You can trigger the workflow using a button in Github UI thanks to the `workflow_dispatch:` trigger

hatskier (approved):

PR #119, opened 2022-11-09
feat: add BTC.b token

cehali:


hatskier (commented):
I don't think it's a good idea to add this token, as it's volume is veery low. Let's use BTC price instead

hatskier (changes requested):
IMO it has too low volume to support this data feed

PR #120, opened 2022-11-09
feat: add WETH.e token

cehali:


hatskier (changes requested):
I also don't think we should add this token, it's value is not very reliable and we can use ETH price instead

PR #151, opened 2022-11-22
feat: local db and improved deviation checks

hatskier:


packages/oracle-node/package.json
```diff
@@ -10,9 +10,9 @@
     "start": "ts-node index.ts",
     "start:prod": "node dist/index.js",
     "start:dev": "yarn start > tmp.out",
-    "test": "jest",
+    "test": "jest --runInBand",
```
cehali, line 13: Do we need this? Are there problems with tests without it?

packages/oracle-node/src/aggregators/median-aggregator.ts
```diff
@@ -5,67 +5,13 @@ import {
   PriceDataBeforeAggregation,
 } from "../types";
 
-const logger = require("../utils/logger")(
-  "aggregators/median-aggregator"
-) as Consola;
-
 const medianAggregator: Aggregator = {
```
cehali, line 8: Why we don't want to filter values from sources which give values way different than others?

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -142,38 +157,149 @@ export default class PricesService {
...
+    const { value, deviationConfig } = args;
+    const { deviationWithRecentValues } = deviationConfig;
+
+    let isValid = false;
+    let reason = "";
+
+    if (isNaN(value)) {
+      reason = "Value is not a number";
+    } else if (value <= 0) {
+      reason = "Value is less or equal 0";
```
cehali, line 260: Should we exclude value 0?

cehali, line 260: image

hatskier: But it is PriceService, so it's used only for prices. Binary values should not go through it

packages/oracle-node/package.json
hatskier: Yes, tests do not work properly without this flag because they rely on the same local level db

packages/oracle-node/src/fetchers/PricesService.ts
cehali: It is called `PriceService` but we use it for all fetching, aggregating, grouping, etc values from all fetchers. F.g. `custom-urls` could potentially return 0 as a proper value in some scenarios.

hatskier: Good point, I will think how to enable this check only for prices

packages/oracle-node/src/NodeRunner.ts
```diff
@@ -215,28 +215,38 @@ export default class NodeRunner {
...
+      // "Helpful" prices (e.g. AVAX_SPOT) can be used to calculate TWAP values
+      const pricesForSigning =
+        this.pricesService!.filterPricesForSigning(aggregatedPrices);
+
       // Signing
       const signedDataPackages = this.signPrices(
-        pricesReadyForSigning,
-        pricesReadyForSigning[0].timestamp
+        pricesForSigning,
+        pricesForSigning[0].timestamp
```
jakub-wojciechowski, line 234: There is no right answer here, but maybe it'll be better to take a timestamp of a moment when all the prices are fetched. Otherwise, we could take the timestamp of the first returned value and artificially extend the update period (which may result into errors or force us to extend the window for price acceptance)

packages/oracle-node/src/aggregators/median-aggregator.ts
```diff
@@ -74,6 +20,11 @@ export function getMedianValue(arr: number[]): number {
   if (arr.length === 0) {
     throw new Error("Cannot get median value of an empty array");
   }
+  if (arr.some(isNaN)) {
```
jakub-wojciechowski, line 23: I'm not sure if that isn't too strict. A single NaN could ruin the whole aggregation process.
How about checking other odd js values (undefined, null void 0, etc...)

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -1,9 +1,90 @@
...
+    PriceValueInLocalDB[]
+  >[] = [];
+
+  for (const price of prices) {
+    const priceForSymbolToAdd: PriceValueInLocalDB = {
+      value: price.value,
+      timestamp: price.timestamp,
+    };
+
+    const currentTimestamp = Date.now();
```
jakub-wojciechowski, line 69: Should we cut off by price.timestamp or currentTime?

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -1,9 +1,90 @@
...
+    const filteredPricesForSymbol = pricesFromDB[price.symbol].filter(
+      (p) => p.timestamp > currentTimestamp - PRICES_TTL_MILLISECONDS
```
jakub-wojciechowski, line 71: Should we cut off by price.timestamp or currentTime?

packages/oracle-node/src/fetchers/PricesService.ts
jakub-wojciechowski: I'll keep checking 0 for prices only, as it's a good signal that something might be wrong (e.g. no data).

hatskier: Actually, I think we can safely remove the 0-values-check, as for prices the zero values will be filtered out using our deviation checker

jakub-wojciechowski: Hmm, do we have a step where we sanitiza arguments. During that phase we should translate all the strings with numeric values to just numbers.

packages/oracle-node/src/aggregators/median-aggregator.ts
hatskier: We already do it in another place (PriceService). This function is supposed to receive valid values and give the right answer (the median value). If we want to extend aggregators in future (e.g. add mean-aggregator), it doesn't make sense to repeat the same filtering logic in each aggregator

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -1,9 +1,90 @@
+import { AbstractBatchPutOperation } from "abstract-level";
 import { Level } from "level";
 import { config } from "../../config";
+import { PriceDataAfterAggregation } from "../../types";
 
-export const db = new Level(config.levelDbLocation, {
+const PRICES_TTL_MILLISECONDS = 15 * 60 * 1000; // 15 minutes
```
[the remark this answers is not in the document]
hatskier: We already keep similar values there. But that's a good point, we can simplify the logic by removing "old" prices only in the moment of adding a new one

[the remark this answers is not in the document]
hatskier: Currently, we do it, but we also filter them in the deviation checking part

packages/oracle-node/src/NodeRunner.ts
hatskier: Let's not change it for now, as we already have a task that will help with the timestamp determination: https://github.com/redstone-finance/redstone-oracles-monorepo/issues/161

packages/oracle-node/src/fetchers/PricesService.ts
hatskier: Now we have :)

packages/oracle-node/src/db/local-db/index.ts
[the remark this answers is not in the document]
hatskier: Adding it to config for now, as it's much easier to handle it from config. Will add a separate issue to move it to manifest later

cehali (approved):
LGTM

PR #155, opened 2022-11-23
feat: cast to number in XtFetcher

cehali:


packages/oracle-node/src/fetchers/xt/XtFetcher.ts
```diff
@@ -33,7 +33,7 @@ export class XtFetcher extends BaseFetcher {
     for (const result of results) {
       const symbol = result.s;
       const prices = result.p;
-      pricesObj[symbol] = prices;
+      pricesObj[symbol] = Number(prices);
```
hatskier, line 36: Can you rename `prices` to `price`?

PR #156, opened 2022-11-23
feat: introduce KYD address levels

cehali:


packages/oracle-node/src/on-demand/CoinbaseKyd.ts
```diff
@@ -0,0 +1,207 @@
...
+
+export const determineAddressLevelByCoinbaseData = async (
+  address: string
+): Promise<number> => {
+  const transactions = await fetchTransactionForAddress(address);
+  const transactionsFromCoinbase = transactions.filter((transaction) =>
+    COINBASE_ADDRESSES.includes(transaction.from)
+  );
+  const sumFromCoinbaseTransactions = transactionsFromCoinbase.reduce(
+    (sum, transaction) => (sum += Number(transaction.value)),
```
hatskier, line 72: Shouldn't we use BigInt or BigNumber here to avoid overflows?

packages/oracle-node/src/on-demand/CoinbaseKyd.ts
```diff
@@ -0,0 +1,207 @@
...
+  if (transactionsSumInUSD > LEVEL_2_MIN_USD_AMOUNT) {
+    return 2;
+  }
+  if (transactionsFromCoinbaseCount > 0) {
+    return 1;
+  }
+  return 0;
+};
+
+const retryRequestIfFailedOrRateLimited = async ({
```
hatskier, line 172: What do you think about using some ready-made library to not reinvent the wheel? E.g. https://www.npmjs.com/package/axios-retry

cehali: The problem with libraries like your example or https://www.npmjs.com/package/async is that they retry only failed transitions. I checked both of them and there isn't an option to retry a request which is not failed but with specific status or message. Below image with the interface of the config parameter called `retryCondition`. It receives `AxiosError` only and I tested it manually if it could intercept requests without error but unfortunately not.

image

packages/oracle-node/src/on-demand/CoinbaseKyd.ts
cehali: fixed

packages/oracle-node/src/on-demand/CoinbaseKyd.ts
```diff
@@ -0,0 +1,212 @@
...
+import { Consola } from "consola";
+import { config } from "../config";
+import { Transaction } from "../db/remote-mongo/models/Transaction";
+
+const logger = require("../utils/logger")("score-by-address") as Consola;
+
+const BIG_END_BLOCK = 99999999;
+const TXS_PER_PAGE = 10000;
+const LEVEL_2_MIN_USD_AMOUNT = 100;
+const WEI_TO_ETH_MULTIPLIER = 10 ** -18;
```
hatskier, line 13: Is it used anywhere?

packages/oracle-node/src/on-demand/CoinbaseKyd.ts
```diff
@@ -0,0 +1,212 @@
...
+) => {
+  const response = await retryRequestIfFailedOrRateLimited({
+    request: () => getEtherscanRequest(address, page),
+    retryCount: RETRY_COUNT,
+    retryInterval: RETRY_INTERVAL,
+  });
+
+  let transactions = (response.data as QueryResponse).result;
+  validateEtherscanResponse(transactions);
+  const transactionFromNextPage = await fetchMoreTransactionsIfPaginated(
```
hatskier, line 104: I think we can get rid of the `fetchMoreTransactionsIfPaginated` function. Its call here has the same number of lines as its source code xD

packages/oracle-node/src/on-demand/CoinbaseKyd.ts
```diff
@@ -0,0 +1,212 @@
...
+  page: number
+): Promise<RawTx[]> => {
+  if (transactionsCount >= TXS_PER_PAGE) {
+    return await fetchTransactionForAddress(address, page + 1);
+  }
+  return [];
+};
+
+const assignAddressLevel = (
+  transactionsCount: number,
```
hatskier, line 157: What do you think about using an object here (smth like `{transactionsCount, transactionsFromCoinbaseCount, transactionsSumInUSD}`) to make it harder to mix up passed arguments? As all of them have the same `number` type

hatskier (approved):

PR #157, opened 2022-11-25
feat: add handling exceptions

cehali:


cehali:
Test `beforeExit` is probably impossible right now: https://github.com/facebook/jest/issues/10927

packages/oracle-node/src/utils/handle-exceptions.ts
```diff
@@ -0,0 +1,39 @@
+import process from "node:process";
+import { Consola } from "consola";
+
+const logger = require("./logger")("index") as Consola;
+
+export const handleExceptions = async (
```
hatskier, line 6: Can you move it to some common module in monorepo and also reuse in the cache service code?

hatskier: Also, I am not sure we should run Promises again (it can be even quite dangerous), the whole point of this task was just not to break the whole process execution :)

hatskier (changes requested):

cehali:
Closing for now, will come back later to this

PR #165, opened 2022-11-27
feat: implemented stats endpoint in cache-service

hatskier:


packages/cache-service/package.json
```diff
@@ -15,7 +15,7 @@
     "start:debug": "nest start --debug --watch",
     "start:prod": "node dist/src/main",
     "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
-    "test": "jest",
+    "test": "jest --runInBand",
```
cehali, line 18: What is the motivation behind this? I think it is only useful for debugging?

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -25,10 +25,24 @@ export interface GetLatestDataPackagesQuery {
   "data-feeds": string;
 }
 
+export interface GetDataPackagesStatsQuery {
+  "from-timestamp": number;
+  "to-timestamp": number;
```
cehali, line 30: Maybe make `to-timestamp` optional if it not required

cehali (approved):

packages/cache-service/package.json
cehali, line 18: What is the motivation behind this? I think it is only useful for debugging or improving test performance in limited resource environment

hatskier: Yes, I noticed that it improves tests performance locally and in GH actions. Also in oracle-node tests this param becomes required because of local level db, so just wanted to keep it consistent in cache-service as well

PR #166, opened 2022-11-28
fix: issue with big numbers, handle address without transactions

cehali:


packages/oracle-node/src/on-demand/CoinbaseKyd.ts
```diff
@@ -79,9 +80,10 @@ export const determineAddressLevelByCoinbaseData = async (
     constants.Zero
   );
   const lastEthPriceInUsd = (await redstone.getPrice("ETH")).value;
-  const ethPriceAsNumber = BigNumber.from(lastEthPriceInUsd.toString());
-  const transactionsSumInUSD =
-    sumFromCoinbaseTransactions.mul(ethPriceAsNumber);
+  const ethPriceAsBigNumber = utils.parseUnits(lastEthPriceInUsd.toString(), 8);
```
hatskier, line 83: What is the reason of converting the withdrawn eth amount to USD value?
IMO to correctly calculate withdrawn USD value you would need to query multiple price values of ETH at each given point of time for each withdrawal

hatskier (approved):

packages/oracle-node/src/on-demand/CoinbaseKyd.ts
cehali: I agree, we discussed an option to handle it based on ETH value but decided to convert it to USD. I believe for the demo purpose this estimation is enough. We can think of a better solution when going with a more "production" version.

PR #176, opened 2022-12-02
feat: add dia-fetcher

Lukasz2891:


packages/oracle-node/src/fetchers/dia/DiaFetcher.ts
```diff
@@ -0,0 +1,41 @@
...
+      .map(resp => resp.data as Quotation))
+      .filter(quotation => quotation != undefined)
+      .reduce((result, quotation) => {
+        result[quotation.Symbol] = quotation.Price;
+
+        return result;
+      }, input);
+  }
+
+  private async quotationResponse(id: string): Promise<any> {
```
hatskier, line 33: Please use the verbs for the function names (e.g. `getQuotationResponse` instead of `quotationResponse`)

packages/oracle-node/src/fetchers/dia/DiaFetcher.ts
```diff
@@ -0,0 +1,41 @@
...
+const DIA_BASE_URL = "https://api.diadata.org/v1";
+const DIA_QUOTATION_PATH = "/quotation"
+
+export class DiaFetcher extends BaseFetcher {
+  constructor() {
+    super("dia");
+  }
+
+  async fetchData(ids: string[]): Promise<any> {
+    return await Promise.all(
```
hatskier, line 15: I would use `Promise.allSettled` instead of `Promise.all` to not throw exception for all symbols if only a subset of requests fails

packages/oracle-node/src/fetchers/dia/dia-quotation-AAVE.json
```diff
@@ -0,0 +1,11 @@
+{
```
hatskier, line 1: Can you move data from `dia-quotation-XXX.json` to `example-response.json` to be consistent with other fetchers?

packages/oracle-node/src/fetchers/dia/dia-quotation-XX11.json
```diff
@@ -0,0 +1 @@
+{"errorcode":404,"errormessage":"no quotation available"}
```
hatskier, line 1: Newline symbol is missing

packages/oracle-node/test/fetchers/dia.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+];
+
+jest.mock("axios");
+
+describe("dia fetcher", () => {
+  const sut = fetchers["dia"];
+
+  it("should properly fetch data", async () => {
+    // Given
+    tokens.forEach(token => mockFetcherResponseOnce(exampleResponsePathFormat.replace("{id}", token)));
```
hatskier, line 25: I would create a separate test for each token. In general, please try to use more standard `for` loops instead of `forEach` to be consistent with the rest of the oracle-node code

hatskier (changes requested):

packages/oracle-node/test/fetchers/dia.spec.ts
Lukasz2891: But the single-token invocation never happens :)

packages/oracle-node/src/fetchers/dia/DiaFetcher.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  async fetchData(ids: string[]): Promise<any> {
+    return await Promise.allSettled(ids.map(this.getQuotationResponse));
+  }
+
+  async extractPrices(responses: any): Promise<PricesObj> {
+    let result: PricesObj = {};
+
+    for (const response of responses) {
+      const quotation = response.value.data as Quotation;
```
hatskier, line 27: I think reading `data` prop from the `response.value` will fail in the case when response.value is undefined (e.g. when respone.status is "rejected")

packages/oracle-node/src/fetchers/dia/DiaFetcher.ts
```diff
@@ -0,0 +1,41 @@
...
+      if (quotation == undefined) {
```
hatskier, line 28: It's a better practice to use `===` instead of `==` if you want to compare with exact value (without type conversion) https://stackoverflow.com/a/359509/12205723

packages/oracle-node/src/fetchers/dia/DiaFetcher.ts
```diff
@@ -0,0 +1,41 @@
...
+        continue;
```
hatskier, line 29: What do you think about simplifying this code to smth like?

```js
for (const response of responses) {
  if (quotation !== undefined) {
    result[quotation.Symbol] = quotation.Price;
  }
}
```

packages/oracle-node/src/fetchers/dia/example-response-AAVE.json
```diff
@@ -0,0 +1,11 @@
+{
```
hatskier, line 1: In future we'll add much more symbols to the DIA fetcher, so let's prevent too many mixed files in the dia folder. I see 2 options how to achieve this:
1. Create a folder named `example-response` and put AAVE.json, ETH.json, ... files there
2. Create a single file example-response.json with the symbol->response mapping, like
```js
{
  "AAVE": { "Symbol": "AAVE", "Name": "Aave Token", .... },
  "ETH": { ... },
  ...
}
```

packages/oracle-node/src/fetchers/dia/example-response-XX11.json
```diff
@@ -0,0 +1 @@
+{"errorcode":404,"errormessage":"no quotation available"}
```
hatskier, line 1: Can you please doublecheck if you also have a JSON formatting configured?

packages/oracle-node/test/fetchers/_helpers.ts
```diff
@@ -59,6 +59,13 @@ export function mockFetcherResponse(pathToResponseFile: string) {
   mockedAxios.post.mockResolvedValue({ data: exampleResponse });
 }
 
+export function mockFetcherResponseOnce(pathToResponseFile: string) {
```
hatskier, line 62: Just curious, what was the problem with the `mockFetcherResponse` function?

packages/oracle-node/src/fetchers/dia/DiaFetcher.ts
```diff
@@ -0,0 +1,41 @@
...
+  constructor() {
+    super("dia");
+  }
+
+  async fetchData(ids: string[]): Promise<any> {
+    return await Promise.allSettled(ids.map(this.getQuotationResponse));
+  }
+
+  async extractPrices(responses: any): Promise<PricesObj> {
+    let result: PricesObj = {};
```
hatskier, line 24: This `let` can be changed to `const`

packages/oracle-node/src/fetchers/dia/DiaFetcher.ts
hatskier: Let's communicate here please ;)

packages/oracle-node/test/fetchers/_helpers.ts
hatskier: Yep, let's do it in a separate task, but later. Can you please create an issue for it on GitHub?

hatskier (approved):

PR #189, opened 2022-12-07
feat: Implement payload endpoint

Lukasz2891:


packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -66,6 +91,21 @@ export class DataPackagesController {
     return await this.dataPackagesService.getDataPackages(requestParams);
   }
 
+  private makePayload(
```
Lukasz2891, line 94: Static? Utils?

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -53,7 +58,25 @@ export class DataPackagesController {
...
+
+  @Get("payload")
+  @Header("Content-Type", "application/octet-stream")
+  async getPayload(
+    @Query() query: GetLatestDataPackagesQuery,
+    @Res() res: Response
+  ) {
+    await this.getDataPackages(query).then((cachedDataPackagesResponse) => {
+      readableStream(
+        RedstonePayloadFactory.makePayload(cachedDataPackagesResponse).toBytes()
```
cehali, line 72: Using `await` and `then` is not good practice. It is a mix of two different principles. In most cases it is better to use `await`

` js
    const cachedDataPackagesResponse = await this.getDataPackages(query)
      readableStream(RedstonePayloadFactory.makePayload(cachedDataPackagesResponse).toBytes()
      ).pipe(res);
    });
`

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -53,7 +58,25 @@ export class DataPackagesController {
...
+      ).pipe(res);
```
cehali, line 73: I don't think we need to use `pipe` here. We can just return the result of `toBytes()` function

packages/cache-service/src/utils/RedstonePayloadFactory.ts
```diff
@@ -0,0 +1,19 @@
+import { RedstonePayload, SignedDataPackage } from "redstone-protocol";
+import { DataPackagesResponse } from "../data-packages/data-packages.controller";
+
+export class RedstonePayloadFactory {
```
cehali, line 4: It doesn't need to be a class if it has only one static function. It could be folder with one or more exported functions

packages/cache-service/src/utils/RedstonePayloadFactory.ts
```diff
@@ -0,0 +1,19 @@
...
+  static makePayload(
+    cachedDataPackagesResponse: DataPackagesResponse,
+    unsignedMetadataMsg?: string
+  ): RedstonePayload {
+    const cachedDataPackages = [];
+    for (const packages of Object.values(cachedDataPackagesResponse)) {
+      cachedDataPackages.push(...packages);
+    }
```
cehali, line 12:
```js
const cachedDataPackages = Object.values(cachedDataPackagesResponse).flatMap(packages => packages);
```

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -53,7 +58,25 @@ export class DataPackagesController {
...
+  }
+
+  @Get("payload")
+  @Header("Content-Type", "application/octet-stream")
+  async getPayload(
+    @Query() query: GetLatestDataPackagesQuery,
+    @Res() res: Response
+  ) {
+    await this.getDataPackages(query).then((cachedDataPackagesResponse) => {
+      readableStream(
```
cehali, line 71: Why you want to use stream here?

packages/cache-service/test/e2e/data-packages.spec.ts
```diff
@@ -163,6 +166,34 @@ describe("Data packages (e2e)", () => {
...
+    const testResponse = await request(httpServer)
+      .get("/data-packages/payload")
+      .query({
+        "data-service-id": "service-2",
+        "unique-signers-count": 4,
+        "data-feeds": "ETH,BTC",
+      })
+      .expect(200);
+
+    expect(testResponse.body.length).toBe(1662);
```
cehali, line 179: Please move those numbers to separate variables with proper names.

packages/cache-service/test/e2e/data-packages.spec.ts
```diff
@@ -163,6 +166,34 @@ describe("Data packages (e2e)", () => {
...
+    const payload = new RedstonePayloadParser(testResponse.body).parse();
+    expect(payload.signedDataPackages.length).toBe(8); // 4*2
```
cehali, line 181: Maybe add a test that will check some value in the response, not only lengths?

cehali (changes requested):
Details in comments

packages/cache-service/src/data-packages/data-packages.controller.ts
Lukasz2891: It doesn’t work. The response is automaticaly transformed to a json except StreamableFile. I’ve spent on it some time. The other way is to use custom Intercepror but that solutions seems to be simplier.

packages/cache-service/test/e2e/data-packages.spec.ts
Lukasz2891: It’d need to have the mocks upgraded. In that case all of mocks consist of the same data, so we can check it, but it doesn’t have a value. There is one check below (with the length of datapackages field). The most important thing is to have a parsable stream returned, for me.

packages/cache-service/src/utils/RedstonePayloadFactory.ts
Lukasz2891: I love of using high level functions but I though we don’t want ;)

packages/cache-service/src/data-packages/data-packages.controller.ts
cehali: Ok, I see. I have to talk with Alex about it, what are the exceptions regarding this. Maybe we could use the `toBytesHex` function on the `makePayload` result and send JSON with a string value as a result.

cehali: I talked about it with Alex and we could do both. Either create a separate endpoint or the same with some parameter that distinguishes if the return value should be a string with the result of `toBytesHex` function or `readableStream` with the result of the `toBytes` function.

Lukasz2891: A separate endpoint will be better because of different content-types. But the naming... Do you accept ```payload``` (with bytes/```octet-stream```) and ```payloadHex``` (with hex/```text/html```) or you prefer ```payloadBytes``` instead?

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -53,7 +58,36 @@ export class DataPackagesController {
...
+  async getPayload(
+    @Query() query: GetLatestDataPackagesQuery,
+    @Res() res: Response
+  ) {
+    const payload = await this.makePayload(query);
+
+    readableStream(payload.toBytes()).pipe(res);
+  }
+
+  @Get("payloadHex")
```
cehali, line 75: I think for route name better would be kebab-case: `payload-hex`

cehali (approved):
Approved with 2 minor changes

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -62,8 +66,38 @@ export class DataPackagesController {
...
+  ) {
+    const payload = await this.dataPackagesService.getPayload(
+      this.prepareDataPackagesRequestParams(query)
+    );
+
+    readableStream(payload.toBytes()).pipe(res);
+  }
+
+  @Get("payload-hex")
+  async getPayloadHex(@Query() query: GetLatestDataPackagesQuery) {
```
hatskier, line 95: What do you think about using query param to specify the expected format of the payload (e.g. `?format=hex` or `?format=raw`)?

packages/cache-service/src/utils/make-redstone-payload.ts
```diff
@@ -0,0 +1,16 @@
+import { RedstonePayload, SignedDataPackage } from "redstone-protocol";
+import { DataPackagesResponse } from "../data-packages/data-packages.controller";
+
+export function makePayload(
```
hatskier, line 4: I would move this function to the redstone-sdk package

packages/cache-service/src/utils/streams.ts
```diff
@@ -0,0 +1,11 @@
+import { Duplex } from "stream";
+import * as Stream from "stream";
+
+export function readableStream(array: Uint8Array): Stream {
+  const stream = new Duplex();
+
+  stream.push(array);
```
hatskier, line 7: Can you please add a comment about what happening here? It's not clear at first sight?

packages/cache-service/src/utils/make-redstone-payload.ts
hatskier: Probably there is already a similar one there, so you can refactor it using the fancy `flatMap` method :)

hatskier: Actually, what do you think about using the `flat` method instead? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

hatskier (changes requested):

packages/cache-service/src/data-packages/data-packages.controller.ts
Lukasz2891: I’ve written it above - the reason were different response types. Less of ’’’ifs’’’

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -115,4 +147,25 @@ export class DataPackagesController {
       await this.bundlrService.safelySaveDataPackages(dataPackagesToSave);
     }
   }
+
+  private sendResponse(
```
hatskier, line 151: Let's change the name of this function to smth like `sendPayloadResponse`

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -6,23 +6,33 @@ import {
...
+import type { Response } from "express";
+import { duplexStream } from "../utils/streams";
+import { Serializable } from "redstone-protocol";
 
 export interface BulkPostRequestBody {
   requestSignature: string;
   dataPackages: ReceivedDataPackage[];
 }
 
+export enum ResponseFormat {
```
hatskier, line 26: Let's use `type ResponseFormat = "raw" | "hex"` instead

packages/cache-service/test/e2e/data-packages.spec.ts
hatskier: Frankly, I see a value in checking the response value, as it will allow us to catch potential problems with payload generation

packages/cache-service/src/utils/make-redstone-payload.ts
hatskier: Ok

hatskier (changes requested):

packages/cache-service/src/data-packages/data-packages.controller.ts
Lukasz2891: That's more ```sendStreamableResponse``` but I don't know if we'd like to reuse it.

packages/cache-service/src/data-packages/data-packages.controller.ts
Lukasz2891: but then we cannot use ResponseFormat.raw but we need to write just "raw" :)

hatskier (approved):

PR #197, opened 2022-12-08
chore: Replace Promise.all with Promise.allSettled by using MultiRequ…

Lukasz2891:
…estFetcher

### DIFF
--- a/packages/oracle-node/src/fetchers/MultiRequestFetcher.ts
+++ b/packages/oracle-node/src/fetchers/MultiRequestFetcher.ts
@@ -1,0 +1,49 @@
+import { BaseFetcher } from "./BaseFetcher";
+import { PricesObj } from "../types";
+
+export abstract class MultiRequestFetcher extends BaseFetcher {
+  abstract processData(
+    data: any,
+    pricesObj: PricesObj,
+    context?: any
+  ): PricesObj;
+  abstract makeRequest(id: string, context?: any): Promise<any>;
+
+  getProcessingContext(): any {
+    return undefined;
+  }
+
+  getRequestContext(): any {
+    return undefined;
+  }
+
+  async fetchData(ids: string[]): Promise<any> {
+    const context = await this.getRequestContext();
+    const promises: Promise<any>[] = [];
+
+    for (const id of ids) {
+      promises.push(this.makeRequest(id, context));
+    }
+
+    return Promise.allSettled(promises);
+  }
+
+  async extractPrices(responses: any): Promise<PricesObj> {
+    let result: PricesObj = {};
+    let context = await this.getProcessingContext();
+
+    for (const response of responses) {
+      if (
+        response.status === "rejected" ||
+        response.value === undefined ||
+        response.value.data === undefined
+      ) {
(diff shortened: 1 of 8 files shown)


Lukasz2891:
I'll create a task for implementing unit tests for particular parts of the MultiRequestFetcher and their implementations.

packages/oracle-node/src/fetchers/MultiRequestFetcher.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface DataProcessor {
+  process(data: any, pricesObj: PricesObj): PricesObj;
+}
+
+export interface RequestFactory {
+  makeRequest(id: string): Promise<any>;
+}
+
+export abstract class MultiRequestFetcher extends BaseFetcher {
+  protected abstract getRequestFactory(): RequestFactory;
```
hatskier, line 13: What is the added value to have the factory and the processor instance here comparing to just having 2 new abstract functions in this class called `makeRequest` and `processResponse`/

hatskier: I like the idea of this refactoring, but let's please try to not create additional abstractions if they are not needed

packages/oracle-node/src/fetchers/MultiRequestFetcher.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  async extractPrices(responses: any): Promise<PricesObj> {
+    const valueProcessor = await this.getDataProcessor();
+    let result: PricesObj = {};
+
+    for (const response of responses) {
+      if (
+        response.status === "rejected" ||
+        response.value === undefined ||
+        response.value.data === undefined
```
hatskier, line 35: Are you sure that each redsponse.value should have the `data` field? It looks a bit axios-specific

packages/oracle-node/src/fetchers/MultiRequestFetcher.ts
```diff
@@ -0,0 +1,45 @@
+import { BaseFetcher } from "./BaseFetcher";
```
hatskier, line 1: Just another question. Have you tested this new code version with the main manifest? If not, ca you please doublecheck if everything works as expected before the merge?

hatskier (changes requested):

packages/oracle-node/src/fetchers/MultiRequestFetcher.ts
Lukasz2891: sometimes I need more variables to create the request. The interfaces in a great way hide that. I like my solution :(

packages/oracle-node/src/fetchers/MultiRequestFetcher.ts
Lukasz2891: Why it couldn't work? Could you help me in that?

Lukasz2891: Now it seems to work properly.

hatskier (approved):

PR #199, opened 2022-12-08
feat: add mock fetcher

rucinpk:


packages/oracle-node/src/fetchers/mock-fetcher/mock-fetcher.ts
```diff
@@ -0,0 +1,68 @@
...
+    this.maxRandomPrice = 100;
+  }
+
+  loadPrices(prices: PricesObj[]) {
+    this.mockedPrices = prices;
+    this.counter = 0;
+  }
+  loadPricesFromFile(path: string) {
+    if (fs.existsSync(path)) {
+      this.mockedPrices = JSON.parse(fs.readFileSync(path, "utf8"));
```
hatskier, line 25: Let's use fetching from external file (using axios), so we don't need to redeploy nodes to enable other mock prices

packages/oracle-node/src/fetchers/mock-fetcher/mock-fetcher.ts
```diff
@@ -0,0 +1,68 @@
+import fs from "fs";
```
hatskier, line 1: It looks quite interesting, but let's pleas simplify this file for now to jest return prices fetched from a given URL

packages/oracle-node/src/types.ts
```diff
@@ -154,6 +154,7 @@ export interface NodeConfig {
   ttlForPricesInLocalDBInMilliseconds: number;
   avalancheRpcUrl: string;
   enableStreamrBroadcasting: boolean;
+  mockedPricesDataPath: string;
```
hatskier, line 157: Let's change it to `mockedPricesDataUrl`

packages/oracle-node/src/config.ts
```diff
@@ -144,4 +145,8 @@ export const config: NodeConfig = Object.freeze({
       DEFAULT_ENABLE_STREAMR_BROADCASTING
     )
   ),
+  mockedPricesDataPath: getFromEnv(
+    "MOCKED_PRICES_DATA_PATH",
```
hatskier, line 149: Let's change it to MOCKED_PRICES_DATA_URL please

hatskier (changes requested):

packages/oracle-node/src/fetchers/mock-fetcher/mock-fetcher.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class MockFetcher extends BaseFetcher {
+  constructor() {
+    super("mock-fetcher");
+  }
+
+  async fetchData(ids: string[]) {
+    let mockedPrices = await axios.get(config.mockPricesUrl);
+    let result = {};
+    ids.forEach((id) => {
```
hatskier, line 14: Let's move the logic with the `result` object preparation to the `extractPrices` method

packages/oracle-node/src/fetchers/mock-fetcher/mock-fetcher.ts
```diff
@@ -0,0 +1,32 @@
...
+      let price;
+      if (id in mockedPrices.data) {
+        price = mockedPrices.data[id];
+      } else {
+        price = mockedPrices.data.__DEFAULT__;
+      }
+      result = {
```
hatskier, line 21: What is the added value of this code comparing to just `result[id] = price;`?

packages/oracle-node/src/fetchers/mock-fetcher/mock-fetcher.ts
hatskier, line 14: Please use `for .. of` loops instead of `forEach` to be consistent with the rest of the oracle-node code. There are some other reasons for it: https://stackoverflow.com/questions/50844095/should-one-use-for-of-or-foreach-when-iterating-through-an-array

packages/oracle-node/src/fetchers/index.ts
```diff
@@ -21,12 +21,14 @@ import { ChainlinkFetcher } from "./chainlink/ChainlinkFetcher";
...
 import { CoinMarketCapFetcher } from "./coinmarketcap/CoinMarketCapFetcher";
 import { config } from "../config";
+import { MockFetcher } from "./mock-fetcher/mock-fetcher";
 
 export default {
   "yf-unofficial": new YfUnofficialFetcher(),
   "custom-urls": new CustomUrlsFetcher(),
   "trader-joe": new TraderJoeFetcher(),
   "twelve-data": new TwelveDataFetcher(),
+  "mock-fetcher": new MockFetcher(),
```
hatskier, line 31: Let's call it just `mock` to be consistent with other fetcher names, which do not have the word `fetcher` in the end :)

packages/oracle-node/manifests/single-source/mock-fetcher.json
```diff
@@ -0,0 +1,25 @@
+{
+  "interval": 15000,
+  "priceAggregator": "median",
+  "defaultSource": ["mock-fetcher"],
```
hatskier, line 4: Let's use the work `mock` instead of `mock-fetcher`

hatskier (changes requested):

packages/oracle-node/src/fetchers/mock-fetcher/mock-fetcher.ts
```diff
@@ -0,0 +1,32 @@
...
+import axios from "axios";
+
+export class MockFetcher extends BaseFetcher {
+  constructor() {
+    super("mock-fetcher");
+  }
+
+  async fetchData(ids: string[]) {
+    let mockedPrices = await axios.get(config.mockPricesUrl);
+    let result = {};
```
hatskier, line 13: `let` can be changed to `const`

packages/oracle-node/src/fetchers/mock-fetcher/mock-fetcher.ts
hatskier, line 14: I think this code can be simplified to
```js
const mockedPrices = response.data;
for (const id of ids) {
  result[id] = mockedPrices[id] ?? mockedPrices.__DEFAULT__;
}

hatskier (approved):

PR #202, opened 2022-12-09
feat: replace redstone-api with local DB

cehali:


packages/oracle-node/manifests/dev/on-demand.json
```diff
@@ -0,0 +1,51 @@
+{
```
hatskier, line 1: Why have you added this manifest?

packages/oracle-node/src/NodeRunner.ts
```diff
@@ -241,7 +236,7 @@ export default class NodeRunner {
 
   private async savePricesInLocalDB(prices: PriceDataAfterAggregation[]) {
     logger.info(`Saving ${prices.length} prices in local db`);
-    await localDB.savePrices(prices);
```
hatskier, line 244: Frankly, I liked the previous version more, as it was more readable in terms of what happens with prices (liek where they are saved)

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -22,11 +22,20 @@ interface LastPrices {
...
-  DEFAULT_LEVEL_OPTS
-);
+let db: Level<string, string>;
+let pricesSublevel: AbstractSublevel<
+  Level<string, string>,
+  string | Buffer | Uint8Array,
+  string,
+  PriceValueInLocalDB[]
+>;
+export const setupLocalDb = () => {
```
hatskier, line 32: Can you add a comment that this function should be called once in order to make this module usable?

hatskier (changes requested):

packages/oracle-node/manifests/dev/on-demand.json
cehali: Because we fetched the ETH price from `redstone-api` before in order to calculate the sum of transactions received from Coinbase in USD. If we want to get rid of `redstone-api` we need to fetch ETH in an "on-demand" type of node ourselves and store it in the local db.

hatskier: Makes sense

hatskier (approved):

Issue #168, opened 2022-11-28
[cache-service] Prepare a new endpoint for getting redstone payload


hatskier:
## Description
We would like to implement a new endpoint in the [redstone-cache-service web app](https://github.com/redstone-finance/redstone-oracles-monorepo/tree/main/packages/cache-service), which will allow to request redstone payload (in hex format).

Currently, it's only possible to request signed data packages in JSON format using `/data-packages/latest` endpoint and then it can be serialized to redstone payload on the client side. Within this task, we would like to extend this endpoint (e.g. by adding an additional query param `format=[hex, json]`) or create a new endpoint, which will be used for redstone payload requesting, so that there will be no need to serialize signed data packages to the hex format on the client side.

## Why should we do it
It will make the integration with non-JS environments (e.g. liquidation job written in RUST) way easier. This feature has already been requested by the AlephZero team.

## Steps
I would propose the following steps for getting into the context
1. Read code of the [redstone-protocol](https://github.com/redstone-finance/redstone-oracles-monorepo/tree/main/packages/protocol)
2. Check available functions in the [redstone-sdk](https://github.com/redstone-finance/redstone-oracles-monorepo/blob/main/packages/sdk/src/index.ts)
3. Implement the feature of requesting payload in the [cache-service](https://github.com/redstone-finance/redstone-oracles-monorepo/tree/main/packages/cache-service)
4. Implement automated tests for the new feature in the cache-service

PR #204, opened 2022-12-09
feat: implementation of signature verification

Lukasz2891:


Lukasz2891:
Contains changes of #168 so that branch is temporary the base .

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -105,20 +109,30 @@ export class DataPackagesService {
...
+    } of signersStats) {
       const nodeDetails = Object.values(state.nodes).find(
         ({ evmAddress }) => evmAddress === signerAddress
       );
 
       stats[signerAddress] = {
         dataPackagesCount,
+        verifiedDataPackagesCount,
+        verifiedDataPackagesPercentage:
+          (100 * verifiedDataPackagesCount) / Math.max(dataPackagesCount, 1),
```
cehali, line 135: Move to separate variable: 
```js 
const (100 * verifiedDataPackagesCount) / Math.max(dataPackagesCount, 1);
```

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -163,4 +183,16 @@ export class DataPackagesService {
 
     return dataPackagesForSaving;
   }
+
+  private isSignatureValid(
```
cehali, line 187: This function doesn't return `isSignatureValid` if `recoverDeserializedSignerAddress` doesn't throw an error

packages/protocol/src/data-package/SignedDataPackageDeserializer.ts
```diff
@@ -0,0 +1,52 @@
...
+  base64,
+  computeAddress,
+  recoverPublicKey,
+  splitSignature,
+} from "ethers/lib/utils";
+import { DataPackage } from "./DataPackage";
+import { SignedDataPackagePlainObj } from "./SignedDataPackage";
+import { Signature } from "ethers";
+
+export interface SignedDataPackageLike {
```
cehali, line 12: I believe this interface is the same as `SignedDataPackagePlainObj`

packages/cache-service/src/data-packages/data-packages.service.ts
Lukasz2891: Why? the return statement is below the try{}catch{}

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -163,4 +185,17 @@ export class DataPackagesService {
...
+    receivedDataPackage: ReceivedDataPackage,
+    signerAddress: string
+  ): boolean {
+    try {
+      const address = recoverDeserializedSignerAddress(receivedDataPackage);
+
```
cehali, line 195: I feel a little bit like we could go with 
```js 
const singedDataPackage = SignedDataPackage.fromObj(receivedDataPackage);
const address = singedDataPackage.recoverSignerAddress();
```
This would save us a lot of additional code and despite creating new instances of the class we don't have to introduce unnecessary computation.

Lukasz2891: I wanted to avoid of using SignedDataPackage here. Creating of an object is not needed, just functions used to create fields inside that object are. So, that was the reason of detaching and moving the functions to a separate file.

Lukasz2891: I can imagine that the SignedDataPackage object would need more fields to be created in the constructor in the future.

Lukasz2891: And also using that object as a proxy would break SRP. The existence of the SignedDataPackage should be implied by itself, not by deserialising signatures in another part of the system, for me :) but that's your decision ;p

packages/protocol/src/data-package/signed-package-deserializing.ts
```diff
@@ -0,0 +1,52 @@
...
+import { Signature } from "ethers";
+
+export interface SignedDataPackageLike {
+  signature: Signature;
+  dataPackage: DataPackage;
+}
+
+export function deserializeSignedPackage(
+  plainObject: SignedDataPackagePlainObj
+): SignedDataPackageLike {
```
hatskier, line 19: Why can't we just use `SignedDataPackage.fromObj` instead?

packages/protocol/src/data-package/signed-package-deserializing.ts
```diff
@@ -0,0 +1,52 @@
...
+  object: SignedDataPackageLike
+): Uint8Array {
+  const digest = object.dataPackage.getSignableHash();
+  const publicKeyHex = recoverPublicKey(digest, object.signature);
+
+  return arrayify(publicKeyHex);
+}
+
+export function recoverSignerAddress(object: SignedDataPackageLike): string {
+  const signerPublicKeyBytes = recoverSignerPublicKey(object);
```
hatskier, line 43: Why can't we just use `recoverSignerAddress` method on the `SignedDataPackage` object?

packages/protocol/src/data-package/SignedDataPackage.ts
```diff
@@ -2,20 +2,27 @@ import { Signature } from "ethers";
...
   base64,
-  computeAddress,
   concat,
   joinSignature,
-  recoverPublicKey,
   splitSignature,
 } from "ethers/lib/utils";
 import { Serializable } from "../common/Serializable";
 import { DataPackage, DataPackagePlainObj } from "./DataPackage";
+import {
```
hatskier, line 11: Why have you decided to move a solid part of this file to a new one?

Lukasz2891: I’ve explained it below @cehali comments. SOLID/SRP.

Lukasz2891:
Do you preffer to have restored the old version of SignedDataPackage? @hatskier

hatskier:
> Do you preffer to have restored the old version of SignedDataPackage? @hatskier

No, let's have this separate file as it is

hatskier (approved):

PR #206, opened 2022-12-10
feat: add evm-examples test to monitoring

cehali:


packages/oracle-monitoring/backend/src/modules/cron/cron.service.ts
```diff
@@ -320,7 +339,48 @@ export class CronService {
...
+  checkPayloadsFromCacheLayer = async (
+    dataServiceId: string,
+    urls: string[]
+  ) => {
+    Logger.log(`Checking payloads: ${dataServiceId} from cache layers`);
+    if (!fs.existsSync("redstone-evm-examples")) {
+      execSync(
+        "git clone https://github.com/redstone-finance/redstone-evm-examples.git"
+      );
+      execSync("cd redstone-evm-examples && yarn");
```
hatskier, line 357: AFAIK `execSync` stops execution of everything else in the app. Can you please investigate it and check if using a standard `exec` would not be a better option?

cehali: Changed to `exec`

packages/oracle-monitoring/backend/src/modules/cron/cron.service.ts
```diff
@@ -320,11 +339,51 @@ export class CronService {
...
         dataServiceId,
         dataFeedId,
-        "invalid-signers-number"
+        "invalid-signers-number",
+        url
       );
     }
   };
 
+  checkPayloadsFromCacheLayer = async (
```
hatskier, line 348: It's a bit misleading that this function is async, because looks like it doesn't return Promise

packages/oracle-monitoring/backend/src/modules/cron/cron.service.ts
```diff
@@ -320,11 +339,51 @@ export class CronService {
...
+    dataServiceId: string,
+    urls: string[]
+  ) => {
+    Logger.log(`Checking payloads: ${dataServiceId} from cache layers`);
+    if (!fs.existsSync("redstone-evm-examples")) {
```
hatskier, line 353: This condition looks a bit weird (and empty) ;)

cehali: 🤦 Fixed

hatskier (approved):

PR #214, opened 2022-12-13
feat: streamr listener allowed evm addresses config

Lukasz2891:


packages/cache-service/src/streamr-listener/streamr-listener.service.ts
```diff
@@ -24,11 +24,14 @@ export class StreamrListenerService {
   private readonly logger = new Logger(StreamrListenerService.name);
   private readonly streamrClient: StreamrClient = new StreamrClient();
   private subscriptionsState: StreamrSubscriptions = {};
+  private readonly allowedNodeEnvAddresses: string[];
```
hatskier, line 27: What is the added value of this private field if we can get this value from the config directly?

packages/cache-service/.env.example
```diff
@@ -3,3 +3,7 @@ ENABLE_STREAMR_LISTENING=true
 ENABLE_DIRECT_POSTING_ROUTES=false
 JWK_KEY_FOR_ARCHIVING_ON_ARWEAVE=
 API_KEY_FOR_ACCESS_TO_ADMIN_ROUTES=
+
+# array of strings in json format: ["0xaaab", "0xbccc", ...];
+# if no value is specified or the array is empty, all nodes will be allowed and no filter will be applied
+ALLOWED_STREAMR_NODE_EVM_ADDRESSES=
```
hatskier, line 9: We would like to filter based on dataServiceIds, not on the nodes, to make the configuration simpler. There is no value in listening on only a subset of nodes from a data service

packages/cache-service/src/config.ts
```diff
@@ -45,6 +46,16 @@ const config = {
     getEnv("MOCK_DATA_SERVICE_ID_FOR_PACKAGES", false) === "true",
 } as CacheServiceConfig;
 
+if (config.enableStreamrListening) {
```
hatskier, line 49: It looks a bit inconsistent to me, that the `allowedStreamrNodeEnvAddresses` field is required in the `CacheServiceConfigRequiredFields` type, but we set it only if  `config.enableStreamrListening` is `true`. I would remove this condition and would replace the code below with something like:
```js
const envValue = getEnv("ALLOWED_STREAMR_NODE_EVM_ADDRESSES", false) || "[]";
config.allowedStreamrNodeEnvAddresses = JSON.parse(envValue);
```

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
```diff
@@ -109,4 +111,39 @@ export class StreamrListenerService {
...
+  getAllowedNodeEnvAddresses(): string[] {
+    if (this.allowedNodeEnvAddresses === undefined) {
+      return [];
+    }
+
+    return this.allowedNodeEnvAddresses.map((obj) => obj.toLowerCase());
+  }
+
+  private prepareActiveNodeEvmAddresses(oracleRegistryState: any): string[] {
+    let nodeEvmAddresses = Object.values(oracleRegistryState.nodes).map(
```
hatskier, line 124: Please change this `let` to `const`

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
```diff
@@ -109,4 +111,39 @@ export class StreamrListenerService {
       this.streamrClient.unsubscribe(subscription);
     }
   }
+
+  getAllowedNodeEnvAddresses(): string[] {
+    if (this.allowedNodeEnvAddresses === undefined) {
+      return [];
+    }
+
+    return this.allowedNodeEnvAddresses.map((obj) => obj.toLowerCase());
```
hatskier, line 120: Why `obj` if it's a string?

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
```diff
@@ -109,4 +111,39 @@ export class StreamrListenerService {
...
+    const allowedNodeEnvAddresses = this.getAllowedNodeEnvAddresses();
+
+    if (allowedNodeEnvAddresses.length == 0) {
+      this.logger.log(
+        `Filter is empty - allowing all of the node evm addresses`
+      );
+      return nodeEvmAddresses;
+    }
+
+    nodeEvmAddresses = nodeEvmAddresses.filter((obj) =>
```
hatskier, line 139: Why `obj`?

hatskier (changes requested):

packages/cache-service/.env.example
Lukasz2891: I don’t like to have mixed external and internal dependencies outside the Constructor. Generally, I was refectoring in my life a lot lot of classes and having it mixed in diffrerent parts of code leads to something not nice. 

But also I maybe don’t feel the importance of performance and memory efficiency in js with so small amount of data.

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
Lukasz2891: I don’t like to have mixed external and internal dependencies outside the Constructor. Generally, I was refectoring in my life a lot lot of classes and having it mixed in diffrerent parts of code leads to something not nice.

But also I maybe don’t feel the importance of performance and memory efficiency in js with so small amount of data.

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
Lukasz2891: it's modified (filtered) below

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
```diff
@@ -12,6 +12,7 @@ import {
 import { DataPackage } from "../data-packages/data-packages.model";
 import { DataPackagesService } from "../data-packages/data-packages.service";
 import { BundlrService } from "../bundlr/bundlr.service";
+import { RedstoneOraclesState } from "redstone-oracles-smartweave-contracts/dist/src/contracts/redstone-oracle-registry/types";
```
Lukasz2891, line 15: hmmm ;/

hatskier (changes requested):

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
```diff
@@ -109,4 +113,41 @@ export class StreamrListenerService {
...
+      dataServiceId.toLowerCase()
+    );
+  }
+
+  private prepareActiveNodeEvmAddresses(oracleRegistryState: any): string[] {
+    const nodes: NodeLike[] = Object.values(oracleRegistryState.nodes);
+    this.logger.log(`Found ${nodes.length} node evm addresses`);
+
+    const allowedDataServiceIds = this.getAllowedDataServiceIds();
+    if (allowedDataServiceIds.length == 0) {
```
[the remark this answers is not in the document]
Lukasz2891: still forgetting about it, especially for numbers, sorry, that's the last time ;p ;)

packages/cache-service/src/streamr-listener/streamr-listener.service.ts
```diff
@@ -109,4 +113,41 @@ export class StreamrListenerService {
       this.streamrClient.unsubscribe(subscription);
     }
   }
+
+  getAllowedDataServiceIds(): string[] {
```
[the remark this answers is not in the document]
Lukasz2891: the body of the function -> right
But I wanted to have a function to have it mockable in tests. I can't do it for a variable - also changing values in process env didn't work. Can we leave the function returning only ```config.allowedStreamrDataServiceIds```?
Or do you have another tip for mocking the config here, working for tests separately?

hatskier (approved):

PR #215, opened 2022-12-14
feat: node schedule and rounded timestamps

rucinpk:


packages/oracle-node/src/NodeRunner.ts
```diff
@@ -130,7 +131,10 @@ export default class NodeRunner {
 
     try {
       await this.runIteration(); // Start immediately then repeat in manifest.interval
-      setInterval(this.runIteration, this.currentManifest!.interval);
+      schedule.scheduleJob(
```
hatskier, line 134: Can you add tests for it and check some edge cases (e.g. case when interval divided by 10000 is not an integer number, or when it's bigger than 60)

packages/protocol/src/data-package/DataPackage.ts
```diff
@@ -109,6 +110,14 @@ export class DataPackage extends Serializable {
     );
   }
 
+  //Rounds timestamp for better node synchronization and querying historical data
```
hatskier, line 113: Let's add a space between `//` and the comment text please. `//Rounds ` -> `// Rounds `

packages/protocol/src/data-package/DataPackage.ts
```diff
@@ -109,6 +110,14 @@ export class DataPackage extends Serializable {
...
+  protected serializeRoundedTimestamp(): Uint8Array {
```
hatskier, line 114: I would move this logic to the oracle-node code

hatskier (changes requested):

packages/protocol/src/data-package/DataPackage.ts
hatskier: Because we will also use the `timestamp` from the oracle node to save some diagnostci info in future, so it's easier to unify the logic of timestamp rounding in the node (and leave the protocol code more flexible)

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -101,7 +102,8 @@ const lastPrices: LastPrices = {};
 const setLastPrices = (prices: PriceDataAfterAggregation[]) => {
   for (const price of prices) {
     const { symbol, value, timestamp } = price;
-    lastPrices[symbol] = { value, timestamp };
+    const roundedTimestamp = roundTimestamp(timestamp);
```
hatskier, line 105: Do we really need to call the `roundTimestamp` function twice? IMO, the call in the NodeRunner should be enough

packages/oracle-node/test/integration/NodeRunner.spec.ts
```diff
@@ -139,6 +139,29 @@ describe("NodeRunner", () => {
...
+      const sut = await NodeRunner.create({
+        ...nodeConfig,
+        overrideManifestUsingFile: manifest,
+      });
+      await expect(sut.run()).rejects.toThrowError(
+        "Interval needs to be divisible by 1000"
+      );
+    });
+
+    it("should create node when interval greater than 60 seconds", async () => {
```
hatskier, line 153: Have you checked how it works in this case?

PR #216, opened 2022-12-14
feat: add price deviation to monitoring

cehali:


packages/oracle-monitoring/backend/src/modules/cron/cron.service.ts
```diff
@@ -320,7 +353,53 @@ export class CronService {
...
+      );
+    }
+  };
+
+  checkDataFeedDeviationInTime = async (
+    dataFeedId: string,
+    dataFeedDetails: DataFeedToCheckDeviationDetails,
+    dataServiceId: string
+  ) => {
+    const apiUrl = "https://api.redstone.finance/prices";
```
hatskier, line 367: What do you think about using `redstone-api` here?

cehali: I didn't use `redstone-api` on purpose. I just get rid of this and I don't want to introduce it once anywhere in monorepo 😄

hatskier (approved):

PR #227, opened 2022-12-18
feat: node schedule and rounded timestamps

rucinpk:


packages/oracle-node/src/utils/timestamps.ts
```diff
@@ -0,0 +1,4 @@
+// Rounds timestamp for better node synchronization and querying historical data
+export const roundTimestamp = (timestamp: number): number => {
+  return Math.floor(timestamp / 10000) * 10000;
```
cehali, line 3: Two questions here. Why do we want to round the timestamp to the nearest 10 seconds? Shouldn't it be somehow connected to the interval in the manifest?
Why should we round with `floor` and not `ceil` or `round`?

rucinpk: 1. We want to make Arweave querying easier. I think it is not connected to the interval in the manifest. 
2. It could be either floor or ceil. I don't see advantages of using one instead of another.

hatskier: 2. It should be floor. Rounding is used here to handle the potential delay between iteration starting and timestamp measurement (e.g. we can start at timestamp X, but at the moment of evaluating `Date.now()` it can become `X + 1` or `X + 2`

packages/oracle-node/src/types.ts
```diff
@@ -81,6 +81,7 @@ export interface PriceDataBeforeAggregation {
   symbol: string;
   source: { [sourceName: string]: any };
   timestamp: number;
+  roundedTimestamp: number;
```
hatskier, line 84: We don't need this field (`roundedTimestamp`)

packages/oracle-node/src/NodeRunnerOld.ts
```diff
@@ -222,13 +223,15 @@ export default class NodeRunnerOld {
     const fetchingAllTrackingId = trackStart("fetching-all");
 
     const fetchTimestamp = Date.now();
```
hatskier, line 225: Let's please use `const fetchTimestamp = roundTimestamp(Date.now());`

hatskier: We don't need to have fetchTimestamp and roundedTimestamp separately

packages/oracle-node/src/utils/timestamps.ts
hatskier: 1. IMO rounding to the nearest 1 second (not 10 seconds) should be enough

hatskier (changes requested):

hatskier (changes requested):

hatskier (approved):

PR #237, opened 2022-12-22
feat: set up efficient caching

hatskier:


packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -77,6 +88,39 @@ export class DataPackagesController {
     return requestParams;
   }
 
+  @Get("latest/:DATA_SERVICE_ID")
+  async getAllLatest(
```
cehali, line 92: Maybe we could move most of this function to the service. I think the controller should only have routes definition and basic functionality.

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -77,6 +88,39 @@ export class DataPackagesController {
...
+          error: "Data service id is invalid",
+        },
+        HttpStatus.BAD_REQUEST
+      );
+    }
+
+    // Checking if data packages for this data service are
+    // presented in the application memory cache
+    const cacheKey = `data-packages/latest/${dataServiceId}`;
+    const dataPackagesFromCache = await this.cacheManager.get(cacheKey);
```
cehali, line 110: Maybe `get` is generic function and we can pass `DataPackagesResponse` there?

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -52,14 +59,18 @@ export interface DataPackagesStatsResponse {
   };
 }
 
+const CACHE_TTL = 5000;
```
cehali, line 62: Could add a comment why we want it 5 seconds? I understand that because our interval is 10 seconds we don't want to cache old prices.

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -33,6 +34,61 @@ export class DataPackagesService {
...
+      },
+      {
+        $sort: { timestampMilliseconds: -1 },
+      },
+    ]);
+
+    // Parse DB response
+    for (const dataPackage of groupedDataPackages) {
+      const { _id, __v, ...rest } = dataPackage;
+      __v;
```
cehali, line 76: Why is it necessary?

packages/evm-connector/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@redstone-finance/evm-connector",
+  "name": "@redstone-finance/evm-connector-exp",
```
cehali, line 2: Why do we want to change the name?

packages/evm-connector/package.json
```diff
@@ -46,4 +46,4 @@
     "dist/**/*",
     "contracts/**/*"
   ]
-}
+}
```
cehali, line 49: Add new line

packages/cache-service/src/data-packages/data-packages.controller.ts
hatskier: Added a comment

cehali (approved):
LGTM

PR #266, opened 2022-12-29
feat: implemented comparison with avg value instead of max deviation

hatskier:


packages/oracle-node/src/utils/numbers.ts
```diff
@@ -12,3 +12,27 @@ export const safelyConvertAnyValueToNumber = (value: any): number => {
...
+  if (nums.length === 0) {
+    throw new Error("Can not calculate an average value for an empty array");
+  }
+
+  return calculateSum(nums) / nums.length;
+};
+
+export const calculateDeviationPercent = (args: {
+  measuredValue: number;
+  trueValue: number;
```
cehali, line 29: I don't think `trueValue` is the proper name here. Maybe something like `valueToCompare`?

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -276,24 +280,28 @@ export default class PricesService {
     };
   }
 
-  // Calculates max deviation from all recent values
-  getDeviationPercentWithRecentValues(args: PriceValidationArgs): number {
+  // Calculates max deviation from average of recent values
+  getDeviationPercentWithRecentValuesAVG(args: PriceValidationArgs): number {
```
cehali, line 284: Maybe `getDeviationPercentWithRecentValuesAVG` -> `getDeviationWithRecentValuesAverage` ?

Percent is not superly informative in this name

cehali (approved):
Approved with minor naming comments

PR #268, opened 2022-12-30
feat: kaiko fetcher

rucinpk:


packages/oracle-node/src/fetchers/kaiko/KaikoFetcher.ts
```diff
@@ -0,0 +1,41 @@
+import { MultiRequestFetcher } from "../MultiRequestFetcher";
+import { PricesObj } from "../../types";
+import { config } from "../../config";
+
+const axios = require("axios");
```
cehali, line 5: Always use `import` if possible:

```js
import axios from "axios";
```

packages/oracle-node/src/fetchers/kaiko/KaikoFetcher.ts
```diff
@@ -0,0 +1,41 @@
...
+const KAIKO_PRICES_URL =
+  "https://eu.market-api.kaiko.io/v2/data/trades.v1/spot_exchange_rate";
+const KAIKO_CONFIG = {
+  headers: {
+    "X-Api-Key": config.kaikoApiKey,
+    Accept: "application/json",
+  },
+  params: {
+    page_size: 1,
+    interval: "1m",
```
cehali, line 16: I am wondering if interval `1m` is not providing us with old data. I am afraid that it could give us data with a granularity of 1 minute and with our node interval of 10 seconds this would lead to stale prices. Could you double-check this? Maybe we need to pass here interval `1s` in order to get the freshest data

packages/oracle-node/src/fetchers/kaiko/KaikoFetcher.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const axios = require("axios");
+
+const KAIKO_PRICES_URL =
+  "https://eu.market-api.kaiko.io/v2/data/trades.v1/spot_exchange_rate";
+const KAIKO_CONFIG = {
+  headers: {
+    "X-Api-Key": config.kaikoApiKey,
+    Accept: "application/json",
+  },
```
cehali, line 13: Please split it into two separate objects. One would be `KAIKO_REQUEST_HEADERS` and other would be `KAIKO_REQUEST_PARAMS`

packages/oracle-node/src/fetchers/kaiko/KaikoFetcher.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export class KaikoFetcher extends MultiRequestFetcher {
+  constructor() {
+    super("kaiko");
+  }
+
+  makeRequest(id: string): Promise<any> {
+    return axios.get(
+      `${KAIKO_PRICES_URL}/${id.toLowerCase()}/usd`,
```
cehali, line 28: Please put the URL string building in a separate variable, just for better readability.

packages/oracle-node/src/fetchers/kaiko/KaikoFetcher.ts
```diff
@@ -0,0 +1,41 @@
...
+      KAIKO_CONFIG
+    );
+  }
+
+  processData(data: any, pricesObj: PricesObj): PricesObj {
+    if (data.result === "error") {
+      return pricesObj;
+    }
+
+    pricesObj[data.query.base_asset.toUpperCase()] = Number(data.data[0].price);
```
cehali, line 38: Please move `data.query.base_asset.toUpperCase()` and `Number(data.data[0].price)` to separate variables with descriptive names

packages/oracle-node/src/fetchers/kaiko/example-responses/AAVE.json
```diff
@@ -0,0 +1,44 @@
+{
```
cehali, line 1: If those response files don't have significant differences (I don't see them at first sight) please leave only one of them.

packages/oracle-node/test/fetchers/kaiko.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+jest.mock("axios");
+
+describe("kaiko fetcher", () => {
+  const sut = fetchers["kaiko"];
+
+  it("should properly fetch data with proper tokens", async () => {
+    // Given
```
cehali, line 36: Those comments are unnecessary

packages/oracle-node/test/fetchers/kaiko.spec.ts
```diff
@@ -0,0 +1,69 @@
+import fetchers from "../../src/fetchers/index";
+import { mockFetcherResponseOnce } from "./_helpers";
+
+const exampleResponsePathFormat =
+  "../../src/fetchers/kaiko/example-responses/{id}.json";
```
cehali, line 5: I would do this differently. Leave one file with an example response, it is good to know how the response looks like. In `mock` function, you can use `mockFetcherResponseWithFunction` to which you can pass the function which returns mocked response (only necessary fields) for the proper token.

cehali (changes requested):
Details in comments

packages/oracle-node/src/fetchers/kaiko/KaikoFetcher.ts
rucinpk: 1s is too small interval for kaiko - oftentimes if I pass 1s to API it doesn't give any results

packages/oracle-node/test/fetchers/kaiko.spec.ts
rucinpk: I think the current solution is cleaner - we separate data from code. With solution you mention we would hardcode every response for all tokens.

packages/oracle-node/src/fetchers/kaiko/KaikoFetcher.ts
```diff
@@ -0,0 +1,44 @@
...
+  page_size: 1,
+  interval: "10s",
+  sort: "desc",
+};
+const KAIKO_CONFIG = {
+  headers: KAIKO_REQUEST_HEADERS,
+  params: KAIKO_REQUEST_PARAMS,
+};
+
+export class KaikoFetcher extends MultiRequestFetcher {
```
hatskier, line 22: Have you checked if it's possible to request multiple assets in a single API request?

packages/oracle-node/test/fetchers/kaiko.spec.ts
cehali: In both solutions, you separate data from code. Also in your solution, you hardcode every response as well. The only difference is that you will have a small file close to the test itself and not files far away from the test.

packages/oracle-node/src/fetchers/kaiko/KaikoFetcher.ts
rucinpk: Yes - it's not possible to request multiple assets in a single API request with available API versions.

packages/oracle-node/test/fetchers/kaiko.spec.ts
rucinpk: I think I don't get this - for now file is data, and with solution you are proposing the code would be data? With `mockFetcherResponseWithFunction` it would require a handler or different `if` for every asset

cehali: I meant something like this: https://github.com/redstone-finance/redstone-oracles-monorepo/pull/279

rucinpk: refactored

cehali (approved):

cehali (changes requested):
One more thing, we need to add manifest and update manifests for tokens which we want to fetch and update manifests for data services

rucinpk:
Unfortunately - interval for fetching from kaiko needs to be 1m. If interval is set to less than 1 minute then API returns null price oftentimes. We still always get latest value from kaiko even if interval is set to 1m.

cehali (approved):
LGTM

PR #285, opened 2023-01-04
284 oracle monitoring add support for the new urls with all data feeds

hatskier:


packages/oracle-monitoring/backend/yarn.lock
```diff
@@ -0,0 +1,6562 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
cehali, line 1: Is it necessary?

packages/oracle-monitoring/backend/src/modules/cron/cron.service.ts
```diff
@@ -420,35 +420,36 @@ export class CronService {
     }
   };
 
+  // TODO: Hmm, why do we have urls here?
```
cehali, line 423: Just for reporting errors, if you don't see any value in this we can remove it and make it optional in uptime-kuma and db

cehali (approved):
Approved with minor comments

PR #289, opened 2023-01-08
chore: few improvements in redstone-sdk function for data packages requesting

hatskier:
- Added storing by timestamp in desc order
- Added failing for many previously unhandled cases (missing data feed id, insufficient number of data packages)


cehali (approved):
LGTM

PR #297, opened 2023-01-11
feat: added cairo connector

Lukasz2891:


packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -59,14 +59,14 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
     return RedstoneDefaultsLib.aggregateValues(values);
   }
 
-  /* ========== FUNCTIONS WITH IMPLEMENTATION (CAN NOT BE OVERRIDEN) ========== */
+  /* ========== FUNCTIONS WITH IMPLEMENTATION (CAN NOT BE overridden) ========== */
```
hatskier, line 62: I would stay with the upper case version here

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -19,19 +19,19 @@ import "../libs/SignatureLib.sol";
 abstract contract RedstoneConsumerBase is CalldataExtractor {
   using SafeMath for uint256;
 
-  /* ========== VIRTUAL FUNCTIONS (MAY BE OVERRIDEN IN CHILD CONTRACTS) ========== */
+  /* ========== VIRTUAL FUNCTIONS (MAY BE overridden IN CHILD CONTRACTS) ========== */
```
hatskier, line 22: I would stay with the upper case version here

packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -40,7 +40,7 @@ contract CalldataExtractor is RedstoneConstants {
       )
     }
     uint256 calldataNegativeOffset = unsignedMetadataByteSize
-      + UNSGINED_METADATA_BYTE_SIZE_BS
```
hatskier, line 43: Can you please move these typo fixes to a separate Pull request?

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
Lukasz2891: it will be replaced :)

packages/evm-connector/contracts/core/CalldataExtractor.sol
Lukasz2891: yes

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -188,6 +188,9 @@ export class DataPackagesService {
       requestParams
     );
 
+    console.log("XXX");
```
Lukasz2891, line 191: wrr :)

hatskier (approved):

PR #320, opened 2023-01-16
feat: add min valid sources for single data feed

cehali:


packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -341,4 +347,24 @@ export default class PricesService {
...
   }
+
+  assertSourcesNumber(price: PriceDataAfterAggregation, manifest: Manifest) {
+    const { symbol, source } = price;
+    const sourcesFetchedCount = Object.keys(source).length;
+    const minValidSourcesPercentage =
+      ManifestHelper.getMinValidSourcesPercentage(manifest);
+    const allSourcesCount = ManifestHelper.getAllSourceCount(symbol, manifest);
+    const validSourcesPercentage =
+      (sourcesFetchedCount / allSourcesCount) * 100;
```
hatskier, line 358: I think this function can fail in manifests with only default sources, because if the division by 0

cehali: image

packages/oracle-node/tools/manifest/add-parameter-to-all-manifests.ts
```diff
@@ -0,0 +1,30 @@
+import fs from "fs";
+import { readJSON, saveJSON } from "../common/fs-utils";
+import {
+  SINGLE_SOURCE_MANIFESTS_FOLDER,
+  DATA_SERVICES_MANIFESTS_FOLDER,
+  DEV_MANIFESTS_FOLDER,
+} from "../common/paths";
+
+(() => {
+  const parameter = {};
```
hatskier, line 10: Can you please add a comment regarding the usage of this script? AFAIU the user should update the `parameter` object before running it

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -341,4 +347,24 @@ export default class PricesService {
...
+    const isSourcesNumberValid =
+      validSourcesPercentage >= minValidSourcesPercentage;
+    if (!isSourcesNumberValid) {
+      const requiredSourcesCount = Math.ceil(
```
hatskier, line 362: I would remove some additional logic for calculating the `requiredSourcesCount` value here. Let's just log valid sources count, and `minValidSourcesPercentage` here

hatskier (changes requested):

hatskier (approved):

packages/oracle-node/src/manifest/ManifestHelper.ts
```diff
@@ -61,4 +63,20 @@ export default class ManifestHelper {
...
+      manifest?.defaultSource?.length;
+    if (!allSourcesCount) {
+      throw new Error(`Cannot define all sources count for symbol ${symbol}`);
+    }
+    return allSourcesCount;
+  }
+
+  static getMinValidSourcesPercentage(manifest: Manifest) {
+    return (
+      manifest?.minValidSourcesPercentage ?? DEFAULT_MIN_VALID_SOURCE_PERCENTAGE
```
hatskier, line 79: Since we have mechanism of getting default `minValidSourcesPercentage`, maybe we can remove all these changes in manifests?

hatskier (approved):

PR #322, opened 2023-01-18
feat: adjust mongo pipeline query for payload

cehali:


packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -111,74 +118,39 @@ export class DataPackagesService {
...
-    };
+    const { dataFeeds, dataServiceId, uniqueSignersCount } = requestConfig;
 
     // Fetching data packages for each data feed
-    if (!!requestConfig.dataFeeds) {
-      const promises = requestConfig.dataFeeds.map(getDataPackagesForDataFeed);
-      await Promise.all(promises);
-    } else {
-      await getDataPackagesForDataFeed(ALL_FEEDS_KEY);
+    if (!!dataFeeds) {
```
hatskier, line 142: Can you please reuse the `getAllLatestDataWithCache` function instead of modifying `getAllLatestDataPackagesFromDB`? It will simplify the code and will help to improve DB performance thanks to mem caching

hatskier (changes requested):

packages/cache-service/src/data-packages/data-packages.service.ts
cehali: Changed

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -299,4 +256,29 @@ export class DataPackagesService {
       return false;
     }
   }
+
+  private filterDataPackages(
```
hatskier, line 260: Can we please reuse `parseDataPackagesResponse` from redstone-sdk here?

hatskier (approved):

PR #325, opened 2023-01-18
feat: add redstone txs analysis on avalanche

rucinpk:


packages/data-analytics/scripts/transactions-stats.js
```diff
@@ -0,0 +1,178 @@
...
+    progressBar.update(progressCounter);
+    i += blockPerPage;
+  }
+
+  progressBar.stop();
+  return matchingTransactions;
+}
+
+const countUniqueAddresses = (transactions) => {
+  let addresses = new HashSet();
```
hatskier, line 90: Please use const instead of let where possible

packages/data-analytics/scripts/transactions-stats.js
```diff
@@ -0,0 +1,178 @@
...
+  latestBlock,
+  blockCount,
+  callData,
+  blockPerPage = 50
+) {
+  let matchingTransactions = [];
+
+  let startingBlock = latestBlock - blockCount;
+
+  let i = startingBlock > 0 ? startingBlock : 0;
```
hatskier, line 57: Please rename `i`, `k`, and `j` variables to something with more helpful names ;)

packages/data-analytics/scripts/transactions-stats.js
```diff
@@ -0,0 +1,178 @@
...
+  if (typeof step == "undefined") {
+    step = 1;
+  }
+
+  if ((step > 0 && start >= stop) || (step < 0 && start <= stop)) {
+    return [];
+  }
+
+  var result = [];
+  for (var i = start; step > 0 ? i < stop : i > stop; i += step) {
```
hatskier, line 40: Please don't use `var`, use `const` or `let` (only if needed) instead

packages/data-analytics/scripts/transactions-stats.js
```diff
@@ -0,0 +1,178 @@
+const prompts = require("prompts");
```
hatskier, line 1: What do you think about switching to typescript in this script? It can require a bit more work now, but will may make it easier to read and extend in future

packages/data-analytics/scripts/transactions-stats.js
```diff
@@ -0,0 +1,178 @@
...
+const Web3 = require("web3");
+const cliProgress = require("cli-progress");
+const HashSet = require("hashset");
```
hatskier, line 4: Can't we just use native JS objects (or Sets) instead of adding a new dependency?

hatskier (changes requested):

packages/data-analytics/scripts/transactions-stats.js
hatskier, line 1: Shouldn't we remove this file?

packages/data-analytics/package.json
```diff
@@ -0,0 +1,18 @@
+{
```
hatskier, line 1: Please rename this PR to something more informative, e.g. `feat: add redstone txs analysis on avalanche`

packages/data-analytics/scripts/transactions-stats.ts
```diff
@@ -0,0 +1,181 @@
+import prompts from "prompts";
```
hatskier, line 1: Please rename this file to make it a bit more informative, e.g. `get-redstone-txs-stats-from-avalanche.ts`. Also please use verb for scripts names

hatskier (changes requested):

packages/data-analytics/scripts/get-redstone-txs-stats-from-avalanche.ts
```diff
@@ -0,0 +1,165 @@
...
+  ];
+};
+
+async function getTransactionsByCallData(
+  latestBlock: number,
+  blockCount: number,
+  callData: string,
+  blockPerPage = 50
+) {
+  let matchingTransactions: any[] = [];
```
hatskier, line 31: Please replace `let` with `const`, since you don't change this variable later (methods calling, e.g. `push` is allowed on constants).

Also, please replace `any` with a more informative type

hatskier (approved):

Issue #99, opened 2022-10-28
[evm-connector] Implement linear time median selection in Solidity


hatskier:
https://rcoh.me/posts/linear-time-median-finding/

PR #327, opened 2023-01-19
chore: benchmark linear median finding algorithm

jkurdek:
Implementation and benchmarks for linear time median finding algorihtm (#99).


hatskier:
Thank you very much @jkurdek for the implementation and benchmarks. Great job!

We definitely won't use arrays with more than 15 elements because it would cause huuuge gas costs on the signature verification and data package deserialization level. It's great that we checked this option, but it doesn't look promising because of the higher code complexity and no gas savings.

So, I am closing the PR :)

PR #331, opened 2023-01-19
feat: add fallback provider for Avalanche EVM fetcher

cehali:


packages/oracle-node/src/fetchers/evm-chain/avalanche/AvalancheEvmFetcher.ts
```diff
@@ -1,15 +1,24 @@
...
+import {
+  MulticallParsedResponses,
+  MulticallRequest,
+  PricesObj,
+} from "../../../types";
 import { extractPrice } from "./extract-price";
 import { oracleAdaptersContractsDetails } from "./contracts-details/oracle-adapters";
 import { glpManagerContractsDetails } from "./contracts-details/glp-manager";
 
+interface Providers {
```
hatskier, line 17: What do you think about using the built-in [FallbackProvider](https://docs.ethers.org/v5/api/providers/other/#FallbackProvider) or implementing another version of CutomProvider which would extend the BaseProvider from ethers?

cehali:
I checked `FallbackProvided` to see if it is possible to make it send one request to the first provider and if the request fails it will send it to the second provider. Unfornetally I couldn't tune the parameters priority, weight, and quorum to achieve the required result. Despite sending requests to both providers it also fails to get prices randomly (f.g. 2 times in a row it gets prices and then 3 times doesn't)

hatskier (changes requested):
Please check if the built-in ethers FallbackProvider wouldn't be sufficient for this case

packages/oracle-node/src/fetchers/evm-chain/avalanche/AvalancheEvmFetcher.ts
```diff
@@ -27,35 +36,55 @@ const MUTLICALL_CONTRACT_ADDRESS = "0x8755b94F88D120AB2Cc13b1f6582329b067C760d";
 
 export class AvalancheEvmFetcher extends BaseFetcher {
   protected retryForInvalidResponse: boolean = true;
+  private isRequestRetried = false;
```
hatskier, line 39: Why do we need this variable?
Also, IMHO it's not a good idea to store it as a fetcher field, because we have the same instance of evm fetcher that is used across different calls and some race condition is possible here

packages/oracle-node/src/fetchers/evm-chain/avalanche/AvalancheEvmFetcher.ts
```diff
@@ -27,35 +36,55 @@ const MUTLICALL_CONTRACT_ADDRESS = "0x8755b94F88D120AB2Cc13b1f6582329b067C760d";
...
+  }
+
+  async handleFallbackAfterRetry(error: any, requests: MulticallRequest[]) {
+    const fallbackProvider = this.fallbackProvider;
+    const isFallback = this.isRequestRetried && fallbackProvider;
+    if (!isFallback) {
+      this.isRequestRetried = true;
       return error;
     }
+    const evmMulticallService = new EvmMulticallService(
```
hatskier, line 82: What do you think about creating a `fallbackMulticallService` in the constructor instead?
I think that would make this function unnecessary

packages/oracle-node/src/fetchers/evm-chain/avalanche/AvalancheEvmFetcher.ts
```diff
@@ -27,35 +36,55 @@ const MUTLICALL_CONTRACT_ADDRESS = "0x8755b94F88D120AB2Cc13b1f6582329b067C760d";
...
-    const requests = [];
+    const requests: MulticallRequest[] = [];
     for (const id of ids) {
       const requestsPerId = prepareMulticallRequests(id);
       requests.push(...requestsPerId);
     }
     try {
       return await this.evmMulticallService.performMulticall(requests);
     } catch (error) {
+      return await this.handleFallbackAfterRetry(error, requests);
```
hatskier, line 71: Let's log a message here about the failed call with the main evm multicall service

hatskier (changes requested):

packages/oracle-node/src/fetchers/evm-chain/avalanche/AvalancheEvmFetcher.ts
```diff
@@ -28,34 +37,53 @@ const MUTLICALL_CONTRACT_ADDRESS = "0x8755b94F88D120AB2Cc13b1f6582329b067C760d";
...
+      this.logger.warn(
+        `[${this.name}]: multicall request failed, trying to use fallback provider`
+      );
+      return await this.tryToRunFallback(error, requests);
+    }
+  }
+
+  async tryToRunFallback(error: any, requests: MulticallRequest[]) {
+    if (!this.fallbackMulticallService) {
       return error;
```
hatskier, line 84: Shouldn't we throw it instead of returning?

hatskier (approved):

PR #332, opened 2023-01-23
feat: added historical prices querying from chainlink

rucinpk:


packages/oracle-node/package.json
```diff
@@ -17,7 +17,8 @@
     "version": "git add -A src",
     "postversion": "git push && git push --tags",
     "docker:build": "docker build -f Dockerfile.public -t redstone-node .",
-    "run-ts": "ts-node"
+    "run-ts": "ts-node",
+    "query-historical-prices": "ts-node scripts/query-historical-prices.ts"
```
hatskier, line 21: Please remove this script, as we have a lot of different scripts in the tools folder and adding shortcuts for all of them here would make the package.json file huge :)

packages/oracle-node/scripts/query-historical-prices.ts
```diff
@@ -0,0 +1,117 @@
+import { contracts, abi } from "../src/fetchers/chainlink/constants";
```
hatskier, line 1: Please move this script to the tools folder, you can create a chainlink subfolder inside

packages/oracle-node/scripts/query-historical-prices.ts
```diff
@@ -0,0 +1,117 @@
...
+
+const getPhaseId = (roundId: bigint) => {
+  const phaseId = Number(roundId >> 64n);
+  return BigInt(phaseId);
+};
+
+const getAllPrices = async (priceFeed: Contract, latestPhaseId: bigint) => {
+  let prices: PriceWithTimestamp[] = [];
+
+  for (let i = latestPhaseId; i > 0; i--) {
```
hatskier, line 32: Please name variables using a bit more letters (not just `i`), because it will help with the code readability

packages/oracle-node/scripts/query-historical-prices.ts
```diff
@@ -0,0 +1,117 @@
...
+
+const getAggregatorPrices = async (contract: Contract, phaseId: bigint) => {
+  let prices: PriceWithTimestamp[] = [];
+
+  let roundID = (phaseId << 64n) | BigInt("1");
+
+  let numErrors = 0;
+  while (true) {
+    try {
+      let promises = range(roundID, roundID + BigInt("100"), BigInt(1)).map(
```
hatskier, line 53: What do you think about replacing the custom `range` function with a [simple shorter javascript implementation](https://stackoverflow.com/questions/3895478/does-javascript-have-a-method-like-range-to-generate-a-range-within-the-supp) on numbers (instead of BigInts) and converting a number to BigInt later (in the map function param)

Range in javascript with normal numbers

packages/oracle-node/scripts/query-historical-prices.ts
```diff
@@ -0,0 +1,117 @@
...
+        (round: any) => contract.getRoundData(round)
```
hatskier, line 54: Please avoid using `any` types. They devalue our usage of typescript

packages/oracle-node/scripts/query-historical-prices.ts
```diff
@@ -0,0 +1,117 @@
...
+      let isReverted = false;
+      await Promise.all(promises).then((results) => {
+        results.forEach((answer) => {
+          const price = BigInt(answer.answer);
+          if (price === 0n) {
+            isReverted = true;
+          } else {
+            const startedAtTimestamp = answer.startedAt;
+            roundID++;
+            prices.push({
```
hatskier, line 65: Doesn't it make the result array unsorted? Since promises passed to the Promise.all function may resolve in random order

packages/oracle-node/scripts/query-historical-prices.ts
```diff
@@ -0,0 +1,117 @@
...
+        break;
+      }
+      numErrors++;
+    }
+  }
+  return prices;
+};
+
+const writeResults = (results: PriceWithTimestamp[]) => {
+  console.log("Saving results to file");
```
hatskier, line 93: I would reuse functions from [tools/common/fs-utils.ts module](https://github.com/redstone-finance/redstone-oracles-monorepo/blob/f6ebf697d61ae89977ee1f4dc0b29effbed62ea8/packages/oracle-node/tools/common/fs-utils.ts)

packages/oracle-node/scripts/query-historical-prices.ts
```diff
@@ -0,0 +1,117 @@
...
+};
+async function query() {
+  const priceFeed = new Contract(
+    PRICE_FEED_ADDR,
+    AGGREGATOR_V3_INTERFACE_ABI,
+    provider
+  );
+  const roundId = await getRoundId(priceFeed);
+  const phaseId = getPhaseId(roundId);
+  let prices = await getAllPrices(priceFeed, phaseId);
```
hatskier, line 105: Please use `const` instead of `let`

hatskier (changes requested):

packages/oracle-node/scripts/query-historical-prices.ts
rucinpk: Yes, that array is unsorted, but array can be sorted by timestamp later on.

packages/oracle-node/tools/chainlink/query-historical-prices.ts
```diff
@@ -0,0 +1,120 @@
...
+  "https://api.avax.network/ext/bc/C/rpc"
+);
+
+interface PriceWithTimestamp {
+  price: string;
+  timestamp: bigint;
+}
+
+const getRoundId = async (priceFeedContract: Contract) => {
+  return BigInt(
```
hatskier, line 19: I would split this into 2 instructions, e.g.
```js
const latestRoundData = await priceFeedContract.latestRoundData();
return BigInt(latestRoundData.roundId.toHexString());
```

packages/oracle-node/tools/chainlink/query-historical-prices.ts
```diff
@@ -0,0 +1,120 @@
...
+      if (numErrors > 10) {
+        break;
+      }
+      numErrors++;
+    }
+  }
+  return prices;
+};
+
+async function query() {
```
hatskier, line 100: Please add longer names to functions to make it easier to read the code later

packages/oracle-node/tools/chainlink/query-historical-prices.ts
```diff
@@ -0,0 +1,120 @@
...
+  return prices;
+};
+
+const getAggregatorPrices = async (contract: Contract, phaseId: bigint) => {
+  let prices: PriceWithTimestamp[] = [];
+
+  let roundID = (phaseId << 64n) | BigInt("1");
+
+  let numErrors = 0;
+  while (true) {
```
hatskier, line 58: This function is quite long. Consider splitting it into 2 functions

packages/oracle-node/tools/chainlink/query-historical-prices.ts
```diff
@@ -0,0 +1,120 @@
...
+              timestamp: startedAtTimestamp.toString(),
+            });
+            const date = new Date(
+              Number(prices[prices.length - 1].timestamp) * 1000
+            );
+          }
+        });
+      });
+      if (isReverted) {
+        if (numErrors > 10) {
```
hatskier, line 84: Those numbers (like this 10) are usually created as constants in the beginning of the file and used in the code as constants

hatskier, line 84: Also, why 10?

hatskier (changes requested):

hatskier (approved):

PR #337, opened 2023-01-26
feat: added balancer fetcher

rucinpk:


packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,161 @@
...
+    if (pairedTokenPrice) {
+      let results = pool?.calcSpotPrice(
+        pool.tokens[0].address,
+        pool.tokens[1].address
+      );
+      return pairedTokenPrice.value / Number(results);
+    }
+  }
+
+  async findUnsupportedTokens(ids: string[]) {
```
hatskier, line 93: There is no need for async here

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,161 @@
...
+    return ids.filter((id: any) => id !== this.supportedToken);
+  }
+
+  async tryInit() {
+    if (this.pool === null || this.pool === undefined) {
```
hatskier, line 98: When can this pool be null or undefined?

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,161 @@
...
+      this.logger.warn(
+        `Symbols: ${unsupportedPairs} not supported in Source: ${this.name} with poolId: ${this.poolId}`
+      );
+    }
+
+    if (timestamp != undefined) {
+      return this.getHistoricalPrice(ids, opts, timestamp);
+    }
+
+    let pool = await balancer.pools.find(this.poolId);
```
hatskier, line 72: Please use `const` instead of `let`, unless it's required

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
rucinpk: Just right after the construction of an object. We can't initialize it in the constructor.

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -33,12 +37,13 @@ export abstract class BaseFetcher implements Fetcher {
...
     symbols: string[],
-    opts?: FetcherOpts
+    opts?: FetcherOpts,
+    timestamp?: number
   ): Promise<PriceDataFetched[]> {
     // Fetching data
     const fetchStartTime = Date.now();
     const ids = symbols.map((symbol) => this.convertSymbolToId(symbol));
-    let response = await this.fetchData(ids, opts);
+    let response = await this.fetchData(ids, opts, timestamp);
```
hatskier, line 46: Please remove the `timestamp` argument from the `fetchData` function. Let's specify it in constructor and save in a private variable for HistoricalFetchers

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -47,7 +52,7 @@ export abstract class BaseFetcher implements Fetcher {
       Date.now() - fetchStartTime <= MAX_RESPONSE_TIME_TO_RETRY_FETCHING_MS;
     if (shouldRetry) {
       this.logger.info("Retrying to fetch data");
-      response = await this.fetchData(ids, opts);
+      response = await this.fetchData(ids, opts, Number(timestamp!.toString()));
```
hatskier, line 55: What is the reason for this strange conversion?

packages/oracle-node/src/fetchers/balancer/historical/BalancerFetcherHistorical.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export class BalancerFetcherHistorical extends BalancerFetcher {
+  private timestamp: number;
+
+  constructor(poolId: string, timestamp: number) {
+    super(`balancer-${poolId}`);
+    this.timestamp = timestamp;
+  }
+
+  protected async calculatePrice(pairedTokenPrice: number): Promise<number> {
```
hatskier, line 18: IMO `calculatePrice` should not be `async`. It should only contain logic of price calculation based

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -15,7 +15,11 @@ export abstract class BaseFetcher implements Fetcher {
   }
```
hatskier, line 15: Please remove changes from the BaseFetcher

packages/oracle-node/src/fetchers/balancer/historical/BalancerFetcherHistorical.ts
hatskier: Also, I think you should not override the logic of price calculation in the historical fetcher. You should only override the way the required data is fetched

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,92 @@
...
+  rpcUrl: config.ethMainRpcUrl as string,
+};
+
+export class BalancerFetcher extends BaseFetcher {
+  protected poolId: string;
+  protected supportedToken?: string;
+  protected pool?: PoolWithMethods;
+  protected balancer: BalancerSDK;
+
+  constructor(poolId: string) {
```
hatskier, line 23: Please rewrite this fetcher so it supports more than one pool. You can take a look at how Uniswap or Sushiswap fetchers are implemented. It will also help to remove unnecessary code of validating supported pairs

hatskier (changes requested):

packages/oracle-node/src/fetchers/balancer/historical/BalancerFetcherHistorical.ts
rucinpk: yes, but for the historical implementation, we did not support some tokens in the past.

packages/oracle-node/src/fetchers/DexFetcher.ts
```diff
@@ -11,8 +11,8 @@ export class DexFetcher extends BaseFetcher {
 
   constructor(
     name: string,
-    private readonly subgraphUrl: string,
-    private readonly symbolToPairIdObj: SymbolToPairId
+    protected readonly subgraphUrl: string,
+    protected readonly symbolToPairIdObj: SymbolToPairId
```
cehali, line 15: What is the reason for the change from protected to private?

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,91 @@
...
+      ? pool.tokens[1].symbol!
+      : pool.tokens[0].symbol!;
+  }
+
+  protected async getPairedTokenPrice() {
+    return getLastPrice(this.baseTokenSymbol)!.value;
+  }
+
+  async extractPrices(response: any): Promise<PricesObj> {
+    const pricesObj: { [symbol: string]: number } = {};
```
cehali, line 68:
```js
const pricesObj: PricesObj = {};
```

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,91 @@
...
+    return pool.tokens[0].symbol == this.baseTokenSymbol
+      ? pool.tokens[1].symbol!
+      : pool.tokens[0].symbol!;
+  }
+
+  protected async getPairedTokenPrice() {
+    return getLastPrice(this.baseTokenSymbol)!.value;
+  }
+
+  async extractPrices(response: any): Promise<PricesObj> {
```
cehali, line 67: Here you know that response would be `spotPrices` with type `SpotPrice[]`. Please change `response` to it.

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,91 @@
...
+    pairId: string,
+    pairedTokenPrice: number
+  ): Promise<SpotPrice> {
+    const pool = await this.balancer.pools.find(pairId);
+    const spotPrice = pool?.calcSpotPrice(
+      pool!.tokens[0].address,
+      pool!.tokens[1].address
+    );
+    const price = pairedTokenPrice / Number(spotPrice);
+    return { id: this.getSymbol(pool!), price };
```
cehali, line 54: Add if statement `if (pool)`. We don't want to calculate price if pool doesn't exists. Also we don't want to push prices which could be NaN to `spotPrices` array.

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  async fetchData(ids: string[]): Promise<any> {
+    const spotPrices: SpotPrice[] = [];
+
+    const pairIds = this.getPairIdsForAssetIds(ids);
+
+    const pairedTokenPrice = await this.getPairedTokenPrice();
+    for (const pairId of pairIds) {
+      const { id, price } = await this.calculatePrice(pairId, pairedTokenPrice);
+      spotPrices.push({ id, price });
```
cehali, line 39: Here you can check if return of `calculatePrice` exists and then push it to  `spotPrices` array.

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,91 @@
...
+  rpcUrl: config.ethMainRpcUrl as string,
+};
+
+interface SpotPrice {
+  id: string;
+  price: number;
+}
+
+export class BalancerFetcher extends BaseFetcher {
+  protected balancer: BalancerSDK;
```
cehali, line 24: Why all members are protected. Most of the time it could be private which is more strict than protected. You should use protected only when subclass is going to use function or variable.

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,98 @@
...
+    return null;
+  }
+
+  protected getSymbol(pool: PoolWithMethods): string {
+    return pool.tokens[0].symbol === this.baseTokenSymbol
+      ? pool.tokens[1].symbol!
+      : pool.tokens[0].symbol!;
+  }
+
+  protected async getPairedTokenPrice() {
```
hatskier, line 70: This function doesn't need to by async

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,98 @@
...
+    const pool = await this.balancer.pools.find(pairId);
+    if (pool) {
+      const spotPrice = pool.calcSpotPrice(
+        pool.tokens[0].address,
+        pool.tokens[1].address
+      );
+      const price = pairedTokenPrice / Number(spotPrice);
+      return { id: this.getSymbol(pool), price };
+    }
+    return null;
```
hatskier, line 61: Let's just throw an error if pool is not found

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  async fetchData(ids: string[]): Promise<any> {
+    const spotPrices: SpotPrice[] = [];
+
+    const pairIds = this.getPairIdsForAssetIds(ids);
+
+    const pairedTokenPrice = await this.getPairedTokenPrice();
+    for (const pairId of pairIds) {
+      const priceResult = await this.calculatePrice(pairId, pairedTokenPrice);
```
hatskier, line 38: Let's not use async calls inside a loop. Execution in a loop causes that requests that can be sent in parallel are executed sequentially. Take a look at the [MultiRequestFetcher.ts](https://github.com/redstone-finance/redstone-oracles-monorepo/blob/069abc182f27959d84e3c33e0d57c7a93bdbbe9d/packages/oracle-node/src/fetchers/MultiRequestFetcher.ts)

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,98 @@
...
+    pairId: string,
+    pairedTokenPrice: number
+  ): Promise<SpotPrice | null> {
+    const pool = await this.balancer.pools.find(pairId);
+    if (pool) {
+      const spotPrice = pool.calcSpotPrice(
+        pool.tokens[0].address,
+        pool.tokens[1].address
+      );
+      const price = pairedTokenPrice / Number(spotPrice);
```
hatskier, line 58: IMHO this line `const price = pairedTokenPrice / Number(spotPrice);` should be moved to the `extractPrices` function. The `fetchData` function should focus only on fetching data (and should be async), while the `extractPrices` can do different manipulations, shouldn't be async and should contain the main logic of parsing response (or responses) into the result prices

hatskier (changes requested):

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
rucinpk: it needs to be async, because child classes f.e. `BalancerHistoricalFetcher`needs this function to be async

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,94 @@
...
+      spotPricesPromises.push(priceResult);
+    }
+    const spotPrices = await Promise.all(spotPricesPromises);
+    return spotPrices;
+  }
+
+  protected async calculatePrice(
+    pairId: string,
+    pairedTokenPrice: number
+  ): Promise<SpotPrice | null> {
```
hatskier, line 47: Let's remove `| null` from this type

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  async fetchData(ids: string[]): Promise<any> {
+    const spotPricesPromises: Promise<SpotPrice | null>[] = [];
+
+    const pairIds = this.getPairIdsForAssetIds(ids);
+
+    const pairedTokenPrice = await this.getPairedTokenPrice();
+    for (const pairId of pairIds) {
+      const priceResult = this.calculatePrice(pairId, pairedTokenPrice);
+      if (!priceResult) {
```
hatskier, line 35: You can remove this condition because `!priceResult` will never be true

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -0,0 +1,94 @@
...
+        throw new Error(`Could not get pool with id ${pairId}`);
+      }
+      spotPricesPromises.push(priceResult);
+    }
+    const spotPrices = await Promise.all(spotPricesPromises);
```
hatskier, line 40: What do you think about using Promise.allSettled here?

hatskier (changes requested):

cehali (approved):

PR #339, opened 2023-01-27
feat: add LWAP for DEXs

cehali:


packages/oracle-node/src/manifest/ManifestHelper.ts
```diff
@@ -57,4 +58,21 @@ export default class ManifestHelper {
   static getDeviationCheckConfigForSymbol(symbol: string, manifest: Manifest) {
     return manifest.tokens[symbol]?.deviationCheck || manifest.deviationCheck;
   }
+
+  static getDefaultAggregator(manifest: Manifest) {
+    return aggregators[manifest.priceAggregator];
+  }
+
+  static getAggregatorForToken(manifest: Manifest, symbol: string) {
```
hatskier, line 66: What do you think about removing the `getDefaultAggregator` function and simplifying this code with:
```js
const aggregatorForTokenName = this. getAggregatorName(manifest, symbol);
return aggregators[aggregatorForTokenName];
```

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -341,4 +351,27 @@ export default class PricesService {
     }
     return deviationCheckConfig;
   }
+
+  private getLiquiditiesIfNecessary(
```
hatskier, line 355: I think this approach is not really extendable. We would like to have different options for calculating liquidity, e.g. using TWAP of liquidity values

hatskier: Also please specify return types of the functions for better readability

packages/oracle-node/src/aggregators/lwap-aggregator.ts
```diff
@@ -0,0 +1,111 @@
+import {
```
hatskier, line 1: Let's please simplify this module as much as possible. IMO it should be responsible only for calculation of the LWAP values, it should not extract liquidity from supplied prices, because we will have different logics of liquidity calculations (e.g. TWAP-liquidity-value)

packages/oracle-node/src/aggregators/lwap-aggregator.ts
```diff
@@ -0,0 +1,111 @@
...
+
+interface ValueWithLiquidity {
+  value: number;
+  liquidity: number;
+}
+
+export const lwapAggregator: Aggregator = {
+  getAggregatedValue(
+    price: PriceDataBeforeAggregation,
+    liquidities?: PriceDataBeforeAggregation[]
```
hatskier, line 24: I would propose to pass liquidities here as an object `{[ sourceName: string]: number }`

packages/oracle-node/src/aggregators/lwap-aggregator.ts
```diff
@@ -0,0 +1,111 @@
...
+const getLwapValue = (
+  price: PriceDataBeforeAggregation,
+  liquidities?: PriceDataBeforeAggregation[]
+): number => {
+  const sources = Object.entries(price.source);
+  const dexsSources = sources.filter(([sourceName]) =>
+    dexs.includes(sourceName)
+  );
+  validateInputs(dexsSources, liquidities);
+  if (dexsSources.length === 1) {
```
hatskier, line 42: Would the result be different without this condition?

hatskier (changes requested):

hatskier (approved):

hatskier (approved):

PR #341, opened 2023-01-31
feat: add uniswap sushiswap historical fetchers

rucinpk:


packages/oracle-node/src/fetchers/uniswap-v3/historical/UniswapV3FetcherHistorical.ts
```diff
@@ -0,0 +1,46 @@
...
+import axios from "axios";
+
+const subgraphUrl =
+  "https://api.thegraph.com/subgraphs/name/uniswap/uniswap-v3";
+const timestampToBlockProviderUrl = "https://coins.llama.fi/block/ethereum/";
+
+export class UniswapV3FetcherHistorical extends UniswapV3Fetcher {
+  private timestamp: number;
+
+  constructor(timestamp: number) {
```
hatskier, line 13: Why did you decide to use `timestamp` in the Uniswap V3 fetcher constructor, but `blockNumber` in the Sushiswap fetcher constructor?

packages/oracle-node/tools/balancer/get-historical-balances-with-graph.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+const pools = [
+  "0xd1ec5e215e8148d76f4460e4097fd3d5ae0a35580002000000000000000003d3",
+  "0x76fcf0e8c7ff37a47a799fa2cd4c13cde0d981c90002000000000000000003d2",
+];
+const url = "https://api.thegraph.com/subgraphs/name/balancer-labs/balancer-v2";
+
+Promise.all = function promiseAllIterative(values: any): any {
+  return new Promise((resolve, reject) => {
+    let results: any[] = [];
```
hatskier, line 17: Please don't use `any` types in typescript files

hatskier: And use `const` where possible

packages/oracle-node/tools/balancer/get-historical-balances-with-graph.ts
```diff
@@ -0,0 +1,115 @@
...
+    });
+  });
+};
+
+const getHistoricalPrices = async (pool: string) => {
+  let prices: PriceWithBlockNumber[] = [];
+
+  let filePrices = readPrices("chainlink-prices-with-block.json");
+
+  let balancerPrices: any = [];
```
hatskier, line 39: Please specify types

packages/oracle-node/tools/balancer/get-historical-balances-with-graph.ts
```diff
@@ -0,0 +1,115 @@
...
+        })
+        .catch((err) => reject(err));
+    });
+  });
+};
+
+const getHistoricalPrices = async (pool: string) => {
+  let prices: PriceWithBlockNumber[] = [];
+
+  let filePrices = readPrices("chainlink-prices-with-block.json");
```
hatskier, line 37: Please use `constz

packages/oracle-node/tools/uniswap-v3/get-historical-prices.ts
```diff
@@ -0,0 +1,98 @@
+import fs from "fs";
+import { UniswapV3FetcherHistorical } from "../../src/fetchers/uniswap-v3/historical/UniswapV3FetcherHistorical";
+import { saveJSON } from "../common/fs-utils";
+
+const UNISWAP_PRICES_FILEPATH = "uniswap-prices.json";
+const TIMESTAMPS_TO_FETCH_FILEPATH = "timestamps-to-fetch.json";
+const PROMISE_BATCH_SIZE = 100;
+
+Promise.all = function promiseAllIterative(values: any): any {
```
hatskier, line 9: What is the purpose of this function?

hatskier (changes requested):

packages/oracle-node/tools/uniswap-v3/get-historical-prices.ts
rucinpk: The purpose of overriding `Promise.all` is to be able to know what timestamp was used to call the promise after resolve

packages/oracle-node/src/fetchers/uniswap-v3/historical/UniswapV3FetcherHistorical.ts
rucinpk: I already got data with `blockNumbers` for SushiSwap

cehali:
Now you can merge it to main as branch `add-uniswap-v3-fetcher` is already merged to main

packages/oracle-node/src/fetchers/DexFetcher.ts
```diff
@@ -11,8 +11,8 @@ export class DexFetcher extends BaseFetcher {
 
   constructor(
     name: string,
-    private readonly subgraphUrl: string,
-    private readonly symbolToPairIdObj: SymbolToPairId
+    protected readonly subgraphUrl: string,
+    protected readonly symbolToPairIdObj: SymbolToPairId
```
cehali, line 15: What is the reason for the change from protected to private?

packages/oracle-node/tools/uniswap-v3/get-historical-prices.ts
cehali: I think that using here `Promise.all` and overwriting it leads to code which is hard to understand. Because of the fact that this is a tool which won't be used very often and `Promise.all` doesn't meet requirements I would go with a simple for loop and fetching data for each iteration. This will increase execution time but improve readability and remove duplicates.

rucinpk: I agree that it makes code unreadable and it does not look good. But thanks to it, it is possible to query quite long historical data. I also tested iterative approach - it it is impossible to query reasonable amount of historical blocks.

packages/oracle-node/src/fetchers/DexFetcher.ts
rucinpk: Protected is needed to be able to use those properties in historical fetcher during Graph query construction

packages/oracle-node/prices.csv
```diff
@@ -0,0 +1,2822 @@
+,timestamp,chainlinkPrice,uniswapPrice,sushiswapEth,sushiswapDai,ohmDaiBased,ohmWethBased,DaiWethBased
```
hatskier, line 1: Can you move this file from the root folder?

packages/oracle-node/tools/uniswap-v3/get-historical-prices.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+  }
+
+  while (priceIterator < pricesToFetch.length - PROMISE_BATCH_SIZE + 1) {
+    let promises = [...Array(PROMISE_BATCH_SIZE).keys()].map((index: any) => {
+      let fetcher = new UniswapV3FetcherHistorical(pricesToFetch[index]);
+      return [fetcher.fetchAll(["OHM"]), pricesToFetch[index]];
+    });
+    await Promise.all(promises)
+      .then((results) => {
```
hatskier, line 69: It's not a good practice to mix async/await syntax with `.then` chaining. You can replace this code with smth like
```js
const results = await Promise.all(promises);
results.forEach((result => ...);
```

hatskier (changes requested):

hatskier (approved):

PR #344, opened 2023-02-03
chore: add storage proxy dry run

jkurdek:


packages/evm-connector/test/mock-wrapper/proxy-storage.test.ts
```diff
@@ -54,6 +54,123 @@ describe("SampleStorageProxy", function () {
     await contract.register(consumerContract.address);
   });
 
+  it("Should return correct oracle value for one asset using dry run", async () => {
```
hatskier, line 57: I think this file can be refactored by moving the following piece of code in a single place instead of duplicating in each (or almost each) test case
```js
    const mockNumericPackages = getRange({
      start: 0,
      length: NUMBER_OF_MOCK_NUMERIC_SIGNERS,
    }).map((i) =>
      getMockNumericPackage({
        dataPoints,
        mockSignerIndex: i as MockSignerIndex,
      })
    );

    const wrappedContract =
      WrapperBuilder.wrap(contract).usingMockDataPackages(mockNumericPackages);

    const dataFeedIds = dataPoints.map((dataPoint) => dataPoint.dataFeedId);
    const dataFeedIdsBytes = dataFeedIds.map(convertStringToBytes32);

```

packages/evm-connector/src/wrappers/OnDemandRequestWrapper.ts
```diff
@@ -52,4 +52,10 @@ export class OnDemandRequestWrapper extends BaseWrapper {
     const unsignedMetadata = this.getUnsignedMetadata();
     return RedstonePayload.prepare(signedDataPackages, unsignedMetadata);
   }
+
+  async requestTimestamp(contract: Contract): Promise<number> {
```
hatskier, line 56: It's not an ideal solution, because the `block.timestamp` in some blockchains may be significantly smaller than the real timestamp (Date.now()). But I'm also not sure how to mock the timestamp for the tests in hardhat + mocha.

@cehali can you please take a look?

hatskier (changes requested):

packages/evm-connector/src/wrappers/OnDemandRequestWrapper.ts
cehali: We could try something like this: https://stackoverflow.com/questions/17446064/how-can-i-simulate-the-passing-of-time-in-mocha-tests-so-that-settimeout-callbac 

But on the other hand, I don't think it is so crucial to add additional work. This solution seems to be all right for testing purposes.

cehali: I agree that could be too much integration. I pushed fixed for the failing test: https://github.com/redstone-finance/redstone-oracles-monorepo/pull/344/commits/62d43ed6ca96210003a0f3d1d605938682fdd8f2

hatskier (approved):

PR #345, opened 2023-02-03
feat: add price feed model

cehali:


packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -0,0 +1,55 @@
...
+
+  function validateTimestampFromUser() private view {
+    if (block.timestamp <= lastBlockTimestamp) {
+      revert BlockTimestampSmallerOrEqualToLastTimestamp(lastBlockTimestamp, block.timestamp);
+    }
+  }
+
+  function validateRound(uint256 proposedRound) private view {
+    if (proposedRound != lastRound + 1) {
+      return;
```
hatskier, line 31: I think we should call `return` in the `storeDataFeedsFromDataPackage`, not here ;)

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  function getChalinlikLikeContractAddresses(bytes32 dataFeedId) public view returns (address) {
```
hatskier, line 35: Let's change the name of this function to `getDataFeedContractAddress`

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.4;
+
+import "@redstone-finance/evm-connector/contracts/data-services/AvalancheDataServiceConsumerBase.sol";
+import "./PriceFeed.sol";
+
+contract PriceFeedsManager is AvalancheDataServiceConsumerBase {
+  uint256 public lastRound = 0;
+  uint256 private lastBlockTimestamp;
```
hatskier, line 10: Let's rename it to the `lastUpdateTimestamp`, because last block timestamp sounds like the timestamp of the last block

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -0,0 +1,55 @@
...
+    priceFeedsContracts[bytes32("ETH")] = new PriceFeed(address(this), bytes32("ETH"));
+  }
+
+  error BlockTimestampSmallerOrEqualToLastTimestamp(
+    uint256 lastBlockTimestamp,
+    uint256 blockTimestamp
+  );
+
+  function validateTimestampFromUser() private view {
+    if (block.timestamp <= lastBlockTimestamp) {
```
hatskier, line 24: This function should not check if the `block.timestamp > lastBlockTimestamp `. It should check if the timestamp from data packages is `> lastBlockTimestamp`.

To do this we need few updates in this contract:
1. The `storeDataFeedsFromDataPackage` function should additionally receive `timestampFromPayload` as an argument
2. We should validate if `timestampFromPayload > lastBlockTimestamp`
3. We should override the [`validateTimestamp`](https://github.com/redstone-finance/redstone-oracles-monorepo/blob/main/packages/evm-connector/contracts/core/RedstoneConsumerBase.sol#L39) function so it additionally checks if the `timestampFromPayload === receivedTimestampMilliseconds `. But we also don't want to remove the default timestamp validation logic

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
hatskier: I would change this function so it returns boolean value if the proposed round is valid

packages/evm-connector/contracts/price-feeds/PriceFeed.sol
```diff
@@ -0,0 +1,70 @@
...
+    override
+    returns (
+      uint80 roundId,
+      int256 answer,
+      uint256 startedAt,
+      uint256 updatedAt,
+      uint80 answeredInRound
+    )
+  {
+    return this.latestRoundData();
```
hatskier, line 49: What do you think about always reverting this function? It would help to avoid its incorrect usage

packages/evm-connector/contracts/price-feeds/PriceFeed.sol
```diff
@@ -0,0 +1,70 @@
...
+  function getDataFeedId() external view returns (bytes32) {
+    return dataFeedId;
+  }
+
+  function decimals() external pure override returns (uint8) {
+    return 8;
+  }
+
+  function description() external pure override returns (string memory) {
+    return "RedStone price feed";
```
hatskier, line 30: Let's return this value from storage

packages/evm-connector/contracts/price-feeds/PriceFeed.sol
```diff
@@ -0,0 +1,70 @@
...
+import "@chainlink/contracts/src/v0.8/interfaces/AggregatorV3Interface.sol";
+
+contract PriceFeed is AggregatorV3Interface {
+  address private owner;
+  bytes32 public dataFeedId;
+  uint256 public dataFeedValue;
+
+  constructor(address _owner, bytes32 _dataFeedId) {
+    dataFeedId = _dataFeedId;
+    owner = _owner;
```
hatskier, line 13: Let's add a string argument `description` in the constructor. IMO it will help with readability of the contracts in blockchain explorers

packages/contract-prices-updater/src/start-updater.ts
```diff
@@ -0,0 +1,47 @@
...
+      signer
+    );
+
+    const wrappedContract = WrapperBuilder.wrap(
+      priceFeedsManagerContract
+    ).usingDataService(
+      {
+        dataServiceId: "redstone-avalanche-prod",
+        uniqueSignersCount: 3,
+        dataFeeds: ["BTC", "ETH"],
```
hatskier, line 34: Let's remove dataFeeds param here, as it will significantly reduce gas costs for the bigger number of data feeds

packages/contract-prices-updater/src/start-updater.ts
```diff
@@ -0,0 +1,47 @@
...
+      },
+      redstoneCacheLayerUrls
+    );
+    const lastRound = await priceFeedsManagerContract.getLastRound();
+
+    try {
+      await wrappedContract.storeDataFeedsFromDataPackage(lastRound);
```
hatskier, line 41: Shouldn't we have `lastRound + 1` here?

packages/contract-prices-updater/src/start-updater.ts
```diff
@@ -0,0 +1,47 @@
...
+      priceFeedsManagerContract
+    ).usingDataService(
+      {
+        dataServiceId: "redstone-avalanche-prod",
+        uniqueSignersCount: 3,
+        dataFeeds: ["BTC", "ETH"],
+      },
+      redstoneCacheLayerUrls
+    );
+    const lastRound = await priceFeedsManagerContract.getLastRound();
```
hatskier, line 38: We still need to think about synchronising several bots, because current solution doesn't solve the problem that if we run 3 bots the number of txs will grow 3 times. Maybe we should have some off-chain mechanism for evaluating the next round. E.g. getting roundNumber based on timestamp and just skipping if someone already updated data for it.

Let's think about it and discuss on a call later

hatskier (changes requested):

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
cehali: I think `getPriceFeedContractAddress` would be more consistent with PriceFeed contract

packages/evm-connector/contracts/price-feeds/PriceFeed.sol
cehali: I am not a Solidity expert, but if we want to return a string, it needs to be a memory.
image

packages/contract-prices-updater/src/start-updater.ts
cehali: Sh*t, I was testing bad scenarios and left it there

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -0,0 +1,113 @@
...
+  function validateTimestampFromPayload(uint256 timestampFromPayload) private view {
+    if (timestampFromPayload <= lastUpdateTimestamp) {
+      revert TimestampFromPayloadSmallerOrEqualToLastTimestamp(
+        timestampFromPayload,
+        lastUpdateTimestamp
+      );
+    }
+  }
+
+  function validateRound(uint256 proposedRound) private view returns (bool) {
```
hatskier, line 90: Let's rename this function to `isProposedRoundValid`

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -0,0 +1,113 @@
...
+    return lastRound;
+  }
+
+  function updateDataFeedValues(uint256 proposedRound, uint256 timestampFromPayload) public {
+    if (!validateRound(proposedRound)) return;
+    lastRound = proposedRound;
+    validateTimestampFromPayload(timestampFromPayload);
+    lastUpdateTimestamp = timestampFromPayload;
+
+    uint256[] memory values = getOracleNumericValuesFromTxMsg(dataFeedsIds);
```
hatskier, line 108: Let's add a comment here that the `getOracleNumericValuesFromTxMsg` function will call the `validateTimestampFromPayload` function for each data package from the redstone payload

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -0,0 +1,113 @@
...
+        string(
+          abi.encodePacked("RedStone price feed for ", string(abi.encodePacked(dataFeedsIds[i])))
+        )
+      );
+    }
+  }
+
+  function validateTimestamp(uint256 receivedTimestampMilliseconds) public view override {
+    uint256 receivedTimestampSeconds = receivedTimestampMilliseconds / 1000;
+    if (block.timestamp < receivedTimestampSeconds) {
```
hatskier, line 53: Can't we just call `RedstoneDefaultsLib.validateTimestamp(receivedTimestampMilliseconds);` here?

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -0,0 +1,113 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.4;
+
+import "../data-services/AvalancheDataServiceConsumerBase.sol";
+import "./PriceFeed.sol";
+
+contract PriceFeedsManager is AvalancheDataServiceConsumerBase {
+  uint256 public lastRound = 0;
+  uint256 private lastUpdateTimestamp;
```
hatskier, line 10: Let's rename to `lastUpdateTimestampMilliseconds`

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -0,0 +1,113 @@
...
+      RedstoneDefaultsLib.DEFAULT_MAX_DATA_TIMESTAMP_DELAY_SECONDS
+    ) {
+      revert RedstoneDefaultsLib.TimestampIsTooOld(receivedTimestampSeconds, block.timestamp);
+    }
+    /* 
+      Here lastUpdateTimestamp is already updated inside updateDataFeedValues
+      after validation in valivalidateTimestampFromUser and equal to timestampFromPayload
+    */
+    if (receivedTimestampMilliseconds != lastUpdateTimestamp) {
+      revert TimestampFromPayloadAndFromDataPackageAreNotTheSame(
```
hatskier, line 74: This error naming is a bit misleading, because timestamp from payload is from data package, because data package is included in the payload. Let' change the error name to `ProposedTimestampDoesNotMatchTimestampFromPayload`

hatskier (changes requested):

packages/evm-connector/contracts/price-feeds/PriceFeed.sol
```diff
@@ -0,0 +1,76 @@
...
+import "@chainlink/contracts/src/v0.8/interfaces/AggregatorV3Interface.sol";
+
+contract PriceFeed is AggregatorV3Interface {
+  address private owner;
+  bytes32 public dataFeedId;
+  uint256 public dataFeedValue;
+  string public descriptionText;
+
+  constructor(
+    address _owner,
```
hatskier, line 13: Please add underscores at the end of naming to follow official convention.
You can read more about the style guidelines here:
- https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/GUIDELINES.md
- https://docs.soliditylang.org/en/latest/style-guide.html#avoiding-naming-collisions

packages/evm-connector/contracts/price-feeds/PriceFeed.sol
```diff
@@ -0,0 +1,76 @@
...
+
+  function description() external view override returns (string memory) {
+    return descriptionText;
+  }
+
+  function version() external pure override returns (uint256) {
+    return 1;
+  }
+
+  function getRoundData(uint80 _roundId)
```
hatskier, line 43: Please add underscore in the end
https://docs.soliditylang.org/en/latest/style-guide.html#avoiding-naming-collisions

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
cehali: You mean `validateTimestamp`?

packages/contract-prices-updater/src/start-updater.ts
```diff
@@ -0,0 +1,55 @@
...
+      priceFeedsManagerContract
+    ).usingDataService(
+      {
+        dataServiceId: "redstone-avalanche-prod",
+        uniqueSignersCount: 3,
+      },
+      redstoneCacheLayerUrls
+    );
+
+    const lastRound = await priceFeedsManagerContract.getLastRound();
```
jakub-wojciechowski, line 39: Is it possible that if we run two parallel, both will try to update with the same roundId but with slightly different timestamp?

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -0,0 +1,113 @@
...
+  function getPriceFeedContractAddress(bytes32 dataFeedId) public view returns (address) {
+    return address(priceFeedsContracts[dataFeedId]);
+  }
+
+  function getLastRound() public view returns (uint256) {
+    return lastRound;
+  }
+
+  function updateDataFeedValues(uint256 proposedRound, uint256 timestampFromPayload) public {
+    if (!validateRound(proposedRound)) return;
```
jakub-wojciechowski, line 103: Do we protect somehow against posting older price-feeds that still fit into validation window, that could trigger a higher roundId?

packages/contract-prices-updater/src/start-updater.ts
cehali: Yes, we want to mitigate by fetching `lastUpdateTimestampMilliseconds` from the contract and checking if the current timestamp is older than `lastUpdateTimestampMilliseconds` by 2 minutes.

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -0,0 +1,113 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.4;
+
+import "../data-services/AvalancheDataServiceConsumerBase.sol";
+import "./PriceFeed.sol";
+
+contract PriceFeedsManager is AvalancheDataServiceConsumerBase {
```
jakub-wojciechowski, line 8: We should also think about a policy of updating PriceManager (e.g. additional provider / change validation logic) while having the same PriceFeeds

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
hatskier: Yes, we do it using the `validateTimestampFromPayload` function

packages/evm-connector/src/WrapperBuilder.ts
```diff
@@ -47,4 +48,17 @@ export class WrapperBuilder {
       nodeUrls
     ).overwriteEthersContract(this.baseContract);
   }
+
+  usingDataPackages(
+    dataPackages: DataPackagesResponse,
+    dataPackagesRequestParams: Omit<
```
hatskier, line 54: Why do we need dataPackagesRequestParams here? I thought it would be sufficient to have only `dataPackages`

packages/evm-connector/src/wrappers/DataPackagesWrapper.ts
```diff
@@ -0,0 +1,62 @@
...
+    private dataPackagesRequestParams: Omit<
+      DataPackagesRequestParams,
+      "uniqueSignersCount" | "dataFeeds"
+    >
+  ) {
+    super();
+  }
+
+  getUnsignedMetadata(): string {
+    return `${version}#${this.dataPackagesRequestParams.dataServiceId}`;
```
hatskier, line 22: I would change it to `${version}#data-packages-wrapper`

packages/evm-connector/src/wrappers/DataPackagesWrapper.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  async getBytesDataForAppending(
+    params: ParamsForDryRunVerification
+  ): Promise<string> {
+    const redstonePayload = this.getRedstonePayload();
+    const disablePayloadsDryRun = Boolean(
+      this.dataPackagesRequestParams.disablePayloadsDryRun
+    );
+    if (!disablePayloadsDryRun) {
+      this.runDryRunVerification({ ...params, redstonePayload });
```
hatskier, line 33: IMHO we don't need dry-runs in this wrapper

hatskier (changes requested):

packages/on-chain-relayer/src/start-relayer.ts
```diff
@@ -0,0 +1,66 @@
...
+      );
+
+      const { lastRound, lastUpdateTimestamp } =
+        await getLastRoundParamsFromContract(priceFeedsManagerContract);
+
+      const currentTimestamp = Date.now();
+      const isTwoMinutesSinceLastUpdate =
+        currentTimestamp - lastUpdateTimestamp >= TWO_MINUTES_IN_MILLISECONDS;
+      if (!isTwoMinutesSinceLastUpdate) {
+        console.log("Cannot update prices yet");
```
hatskier, line 34: I would replace "cannot" with "should not", because the contract doesn't prevent updating prices even in each block

packages/on-chain-relayer/src/start-relayer.ts
```diff
@@ -0,0 +1,66 @@
...
+        contractAddress,
+        abi,
+        signer
+      );
+
+      const { lastRound, lastUpdateTimestamp } =
+        await getLastRoundParamsFromContract(priceFeedsManagerContract);
+
+      const currentTimestamp = Date.now();
+      const isTwoMinutesSinceLastUpdate =
```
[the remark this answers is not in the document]
hatskier: Also, I would propose to set it to 115 or 110 seconds to prevent potential delays

packages/evm-connector/src/WrapperBuilder.ts
cehali: If we don't want to do dry run and have name of the `dataServiceId` then we don't need `dataPackagesRequestParams` 😄

hatskier (approved):

PR #347, opened 2023-02-05
chore: fixes in token contracts after abdk audit

hatskier:


packages/eth-contracts/contracts/VestingWallet.sol
```diff
@@ -54,48 +56,54 @@ contract VestingWallet is Initializable {
     } else if (timestamp > start + cliffDuration + vestingDuration) {
       return 0;
     } else {
-      return
-        (allocation * (start + cliffDuration + vestingDuration - timestamp)) / vestingDuration;
+      uint256 elapsedTime = start + cliffDuration + vestingDuration - timestamp;
+      return SafeMath.mul(allocation, elapsedTime) / vestingDuration;
```
jakub-wojciechowski, line 60: Hmm, I think we should have an edge case for vestingDuration == 0 (only cliff and then 100% tokens are unlocked). Otherwise, we've got div by zero.

jakub-wojciechowski (approved):

PR #350, opened 2023-02-07
feat: implement poc fetchers with dex aggregators

hatskier:


hatskier:
It was a PoC, we won't merge it now

PR #353, opened 2023-02-07
feat: improved LP tokens price calculation

rucinpk:
Implemented fair LP token pricing


packages/oracle-node/src/fetchers/evm-chain/avalanche/extract-price.ts
```diff
@@ -18,6 +18,7 @@ import { yieldYakContractsDetails } from "./contracts-details/yield-yak";
 import { mooTokensContractsDetails } from "./contracts-details/moo-joe";
 import { oracleAdaptersContractsDetails } from "./contracts-details/oracle-adapters";
 import { glpManagerContractsDetails } from "./contracts-details/glp-manager";
+import bn from "bignumber.js";
```
cehali, line 21: Please use big number built in `ethers.js` -https://docs.ethers.org/v5/api/utils/bignumber/

packages/oracle-node/src/fetchers/evm-chain/avalanche/extract-price.ts
```diff
@@ -79,6 +86,10 @@ const extractPriceForLpTokens = async (
     lpTokensContractsDetails[id as LpTokensDetailsKeys];
   const reserves = multicallResult[address].getReserves.value;
 
+  const totalSupply = BigNumber.from(
+    multicallResult[address].totalSupply.value
+  );
+
```
cehali, line 92: Generally, variables should be initialized as close as possible to the place they will be used. Please move it somewhere closer, somewhere inside  `if (tokenPrices)`

packages/oracle-node/src/fetchers/evm-chain/avalanche/extract-price.ts
```diff
@@ -87,41 +98,33 @@ const extractPriceForLpTokens = async (
...
+  if (tokenPrices) {
+    const firstTokenPrice = tokenPrices[firstToken];
+    const secondTokenPrice = tokenPrices[secondToken];
+
+    const sqrtK = sqrt(
+      reservesSerialized[firstToken].mul(reservesSerialized[secondToken])
+    ).div(totalSupply);
+    const sqrtPx0Px1 = sqrt(firstTokenPrice.mul(secondTokenPrice));
+    const lpTokenPrice = sqrtK.mul(sqrtPx0Px1).mul(2);
+
```
cehali, line 114: I would improve the naming here because it gets difficult to read.

``` js
const reservesMultiplied = reservesSerialized[firstToken].mul(reservesSerialized[secondToken];
const pricesMultiplied = firstTokenPrice.mul(secondTokenPrice);
const reservesMultipliedSqrt = sqrt(reservesMultiplied);
const pricesMultipliedSqrt = sqrt(pricesMultiplied);
const reservesPricesMulitplied = sqrt(reservesMultiplied).mul( sqrt(pricesMultiplied)).mul(2);
const lpTokenPrice = reservesPricesMulitplied.div(totalSupply);
```

packages/oracle-node/src/fetchers/evm-chain/avalanche/extract-price.ts
```diff
@@ -18,6 +18,7 @@ import { yieldYakContractsDetails } from "./contracts-details/yield-yak";
 import { mooTokensContractsDetails } from "./contracts-details/moo-joe";
```
cehali, line 18: Please add it somewhere at the top of the file or just after updates from where we take the formula for calculating LP tokens price

packages/oracle-node/src/fetchers/evm-chain/avalanche/extract-price.ts
```diff
@@ -71,6 +72,12 @@ const extractPriceForYieldYakOrMoo = async (
   }
 };
 
+const sqrt = (value: BigNumber): BigNumber => {
+  return BigNumber.from(
+    new bn(value.toString()).sqrt().toFixed().split(".")[0]
+  );
+};
+
 const extractPriceForLpTokens = async (
```
cehali, line 81: I wonder if the same formula could be used in other tokens from AvalancheEVMFetcher f.g. in functions `extractPriceForYieldYakOrMoo` or `extractPriceForYieldYakOrMoo`

rucinpk: it can't be implemented, it's only relevant to LP tokens

packages/oracle-node/src/utils/math.ts
```diff
@@ -0,0 +1,15 @@
+import { ethers, BigNumber } from "ethers";
```
cehali, line 1: I forgot last time, please write few test to be sure that it works

cehali, line 1: If you import `BigNumber` you don't need `ethers`

packages/oracle-node/test/utils/math.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+  test("Should properly square integers", async () => {
+    const number = BigNumber.from(9);
+
+    const result = sqrt(number);
+
+    expect(result.toNumber()).toEqual(3);
+  });
+
+  test("Should properly square big integers", async () => {
+    const number = BigNumber.from(999999999).mul(BigNumber.from(999999999));
```
cehali, line 14: Extract `BigNumber.from(999999999)` to separate variable

packages/oracle-node/test/utils/math.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const result = sqrt(number);
+
+    expect(result.toNumber()).toEqual(999999999);
+  });
+
+  test("Should properly floor square big integers", async () => {
```
cehali, line 21: That's not good. We could miscalculate prices because of this. This could mean that your previous implementation with `bignumber.js` could be better because it seems more accurate. Please investigate it and change the implementation in `math.ts`.

cehali (changes requested):
Details in comments

packages/oracle-node/test/utils/math.spec.ts
rucinpk: No I don't think we can - BigNumber is integer so there is no way to make float squaring here. BigNumber prices are denominated in Wei so there is not reason to make floor division anyway.

cehali: Yes, you are right. I forgot that we already serialize all values before square rooting.

cehali (approved):
LGTM

hatskier (approved):

PR #363, opened 2023-02-09
feat: prepare on-chain-relayer for OHM

cehali:


hatskier (changes requested):
Please make the code more configurable so that we can configure avalanche subnet relayer with avalanche prod data feeds and OHM relayer using the same docker image but diffferent env variables

packages/on-chain-relayer/src/scripts/deploy-contracts.ts
```diff
@@ -0,0 +1,50 @@
+import { ContractFactory, Wallet, utils } from "ethers";
+import { getProvider } from "../utils";
+import { config } from "../config";
+import managerAbi from "../config/price-feeds-manager.abi.json";
+import registryAbi from "../config/price-feeds-registry.abi.json";
+import { bytesCodes } from "../config/bytes-codes";
+
+(async () => {
```
hatskier, line 8: Let's rename this file to deploy-contracts-ohm.ts

hatskier (approved):

PR #370, opened 2023-02-13
feat: added timestamp delay

rucinpk:


packages/sdk/src/index.ts
```diff
@@ -69,6 +73,20 @@ export const parseDataPackagesResponse = (
       );
     }
 
+    const maxTimestampDelay = reqParams.maxTimestampDelay;
+
+    if (
+      maxTimestampDelay &&
```
hatskier, line 79: We should run this check only on data packages after calling the `slice` method (`.slice(0, reqParams.uniqueSignersCount)`).

To give an example: Assume we have 5 nodes, but require only 3 of them. Then assume 1 of the nodes stopped working and provides outdated packages. If we check all 5 data packages from all nodes then users won't be able to use 3 valid nodes. That's why we need to check only 3 newest data packages if they contain non-outdated timestamps

packages/sdk/test/index.test.ts
```diff
@@ -6,6 +6,7 @@ import {
 } from "../src";
 import { mockSignedDataPackages } from "./mocks/mock-packages";
 import { server } from "./mocks/server";
+import axios from "axios";
```
hatskier, line 9: Why did you add this line?

hatskier (dismissed):

packages/sdk/src/index.ts
```diff
@@ -14,11 +14,14 @@ export const DEFAULT_CACHE_SERVICE_URLS = [
 
 const ALL_FEEDS_KEY = "___ALL_FEEDS___";
 
+const SECOND_IN_MILLISECONDS = 1000;
```
cehali, line 17: Unused variable?

packages/sdk/src/index.ts
```diff
@@ -69,12 +73,37 @@ export const parseDataPackagesResponse = (
...
+          currentTimestamp - maxTimestampDelay >=
+          dataFeedPackage.timestampMilliseconds;
+
+        if (isDataPackageOutdated) {
+          outdatedDataPackagesTimestamps.push(
+            dataFeedPackage.timestampMilliseconds
+          );
+        }
+        return isDataPackageOutdated;
+      })
```
cehali, line 96: I would simplify it: 
``` js
const maxTimestampDelay = reqParams?.maxTimestampDelay ?? currentTimestamp;
const outdatedDataPackages = dataFeedPackages.filter((dataFeedPackage) => currentTimestamp - maxTimestampDelay >= dataFeedPackage.timestampMilliseconds);
const isAnyPackageOutdated = outdatedDataPackages.length > 0
if (isAnyPackageOutdated) {
const outdatedDataPackagesTimestamps = outdatedDataPackages.map(({ timestampMilliseconds }) => timestampMilliseconds);
throw new Error(
        `At least one datapackage is outdated. Current timestamp: ${currentTimestamp}. Outdated datapackages timestamps: ${outdatedDataPackagesTimestamps}`
      );
}
```

packages/sdk/test/index.test.ts
```diff
@@ -91,4 +91,25 @@ describe("SDK tests", () => {
...
     );
   });
+
+  test("Should fail for outdated data package", async () => {
+    await expect(() =>
+      requestDataPackages({
+        ...reqParams,
+        maxTimestampDelay: 10000,
+      })
+    ).rejects.toThrow("At least one datapackage is outdated");
```
cehali, line 101: Please check if the error string is properly parsed. I think `outdatedDataPackagesTimestamps` could require `JSON.stringify`

cehali (approved):
LGTM

packages/sdk/test/index.test.ts
```diff
@@ -91,4 +91,27 @@ describe("SDK tests", () => {
...
   });
+
+  test("Should fail for outdated data package", async () => {
+    await expect(() =>
+      requestDataPackages({
+        ...reqParams,
+        maxTimestampDelay: 10000,
+      })
+    ).rejects.toThrow(
+      /At least one datapackage is outdated. Current timestamp: [0-9]*. Outdated datapackages timestamps: \[1654353400000,1654353400000\]/
```
hatskier, line 102: You can mock Date.now using Jest to make the tests deterministic and avoid workarounds like this "Current timestamp: [0-9]*"

packages/sdk/src/index.ts
```diff
@@ -69,8 +71,26 @@ export const parseDataPackagesResponse = (
       );
     }
 
+    const maxTimestampDelay = reqParams?.maxTimestampDelay ?? currentTimestamp;
```
hatskier, line 74: The `parseDataPackagesResponse` function became too big. I would propose to move the timestamp delay validation to a separate function.

hatskier (changes requested):

hatskier (approved):

PR #373, opened 2023-02-14
feat: prepared oracle-node for external operators

hatskier:


packages/oracle-node/manifests/data-services/kaiko.json
```diff
@@ -0,0 +1,28 @@
+{
```
cehali, line 1: We already have Kaiko manifest: https://github.com/redstone-finance/redstone-oracles-monorepo/blob/main/packages/oracle-node/manifests/single-source/kaiko.json

We currently use it main manifest for dev nodes. If we want to change the strategy for Kaiko as a source we should delete it from the main manifest.

hatskier: I would like to leave it in the data-services folder, manifests in single-source contain different update interval and should not be used by external node operators

cehali (approved):

PR #377, opened 2023-02-15
feat: on block iterations model implemented in oracle-node

hatskier:


packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -85,12 +85,7 @@ export class DataPackagesController {
     return requestParams;
   }
 
-  @Get("latest/:DATA_SERVICE_ID")
-  @Header("Cache-Control", "max-age=5")
-  async getAllLatest(
-    @Param("DATA_SERVICE_ID") dataServiceId: string
-  ): Promise<DataPackagesResponse> {
-    // Validate dataServiceId param
+  private async validateDataServiceId(dataServiceId: string) {
```
cehali, line 88: Maybe move to service or create a new file with validators?

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -52,23 +52,29 @@ export class DataPackagesService {
...
+  }
+
   async isDataServiceIdValid(dataServiceId: string): Promise<boolean> {
     const oracleRegistryState = await getOracleRegistryState();
     return !!oracleRegistryState.dataServices[dataServiceId];
   }
 
-  async getAllLatestDataPackagesFromDB(
-    dataServiceId: string
+  async getDataPackagesFromDB(
```
cehali, line 75: Just an idea for the future, maybe we could extract functions which directly interact with DB to separate file

packages/oracle-node/src/NodeRunner.ts
```diff
@@ -118,15 +122,16 @@ export default class NodeRunner {
...
     this.maybeRunDiagnosticInfoPrinting();
 
     try {
-      await this.runIteration();
-      const cronScheduleString = intervalMsToCronFormat(
-        this.currentManifest!.interval
-      );
-      schedule.scheduleJob(cronScheduleString, this.runIteration);
+      const schedulerName =
+        this.currentManifest!.useCustomScheduler ?? DEFAULT_SCHEDULER_NAME;
```
cehali, line 132: `getScheduler` already receive the manifest so maybe getting `schedulerName` could be done inside `getScheduler`

cehali: Also, I think `getScheduler` and `DEFAULT_SCHEDULER_NAME` could be moved to `ManifestHelper` in order to have all functionalities regarding manifest there

packages/oracle-node/src/signers/PriceSignerService.ts
```diff
@@ -8,6 +8,8 @@ import {
 } from "../types";
 import { trackStart, trackEnd } from "../utils/performance-tracker";
 
+// TODO: remove this service. Use Universal signer from redstone-protocol instead
+
```
cehali, line 12: I believe this one `PriceSignerService` is used in old infrastructure like API or app

packages/oracle-node/src/types.ts
```diff
@@ -2,7 +2,8 @@ import { JWKInterface } from "arweave/node/lib/wallet";
 
 export interface Manifest {
   txId?: string; // Note, you need to set this field manually (after downloading the manifest data)
-  interval: number;
+  interval: number; // It is ifnored if the `useCustomScheduler` option is specified
```
cehali, line 5: What if `useCustomScheduler` is interval?

packages/oracle-node/src/types.ts
```diff
@@ -12,6 +13,7 @@ export interface Manifest {
   httpBroadcasterURLs?: string[];
   enableStreamrBroadcaster?: boolean;
   minValidSourcesPercentage?: number;
+  signBlockNumbersInsteadOfTimestamps?: boolean;
```
cehali, line 16: Maybe remove`signBlockNumbersInsteadOfTimestamps` and sign block number or not depending on the value of `useCustomScheduler`. It seems strange that in order to switch to on block scheduler you need to add 2 env variables.

hatskier: It's not obvious that in all cases we will sign block numbers even if we schedule iterations on each block. We may also want to sign timestamps

packages/oracle-node/src/types.ts
hatskier: Updated the comment

packages/oracle-node/src/NodeRunner.ts
hatskier: Good idea. I've refactored it

packages/cache-service/src/data-packages/data-packages.service.ts
hatskier: Feel free to add a task for it :)

packages/cache-service/src/data-packages/data-packages.controller.ts
hatskier: I think this function should not be moved to the data package service, because the main logic of this function is related to managing HTTP status of the response

cehali (approved):
LGTM with one minor comment

PR #390, opened 2023-02-23
chore: benchmark single-data set signed by multiple signers

jkurdek:
Implements #319.

### DIFF
--- a/packages/evm-connector/contracts/core/CalldataExtractor.sol
+++ b/packages/evm-connector/contracts/core/CalldataExtractor.sol
@@ -66,12 +66,43 @@ contract CalldataExtractor is RedstoneConstants {
     return dataPackagesCount;
   }
 
+  function _extractDataPackageSignersCountFromCalldata(uint256 calldataNegativeOffset) internal pure returns (uint16 signerCount) {
+    uint256 calldataNegativeOffsetWithStandardSlot = calldataNegativeOffset + STANDARD_SLOT_BS;
+
+    if (calldataNegativeOffsetWithStandardSlot > msg.data.length) {
+      revert CalldataOverOrUnderFlow();
+    }
+    assembly {
+      signerCount := calldataload(
+        sub(calldatasize(), calldataNegativeOffsetWithStandardSlot)
+      )
+    }
+    return signerCount;
+  }
+
   function _extractDataPointValueAndDataFeedId(
     uint256 calldataNegativeOffsetForDataPackage,
     uint256 defaultDataPointValueByteSize,
     uint256 dataPointIndex
   ) internal pure virtual returns (bytes32 dataPointDataFeedId, uint256 dataPointValue) {
     uint256 negativeOffsetToDataPoints = calldataNegativeOffsetForDataPackage + DATA_PACKAGE_WITHOUT_DATA_POINTS_BS;
+    uint256 dataPointNegativeOffset = negativeOffsetToDataPoints.add(
+      (1 + dataPointIndex).mul((defaultDataPointValueByteSize + DATA_POINT_SYMBOL_BS))
+    );
+    uint256 dataPointCalldataOffset = msg.data.length.sub(dataPointNegativeOffset);
+    assembly {
+      dataPointDataFeedId := calldataload(dataPointCalldataOffset)
+      dataPointValue := calldataload(add(dataPointCalldataOffset, DATA_POINT_SYMBOL_BS))
+    }
+  }
+
+  function _extractDataPointValueAndDataFeedIdMultiSign(
+    uint256 calldataNegativeOffsetForDataPackage,
+    uint256 defaultDataPointValueByteSize,
+    uint256 dataPointIndex,
+    uint256 signatureCount
+  ) internal pure virtual returns (bytes32 dataPointDataFeedId, uint256 dataPointValue) {
+    uint256 negativeOffsetToDataPoints = calldataNegativeOffsetForDataPackage + signatureCount * SIG_BS + SIGNERS_COUNT_BS + DATA_PACKAGE_WITHOUT_DATA_POINTS_BS - SIG_BS;
     uint256 dataPointNegativeOffset = negativeOffsetToDataPoints.add(
       (1 + dataPointIndex).mul((defaultDataPointValueByteSize + DATA_POINT_SYMBOL_BS))
     );
@@ -107,6 +138,28 @@ contract CalldataExtractor is RedstoneConstants {
     }
 
     // Prepare returned values
+    dataPointsCount = dataPointsCount_;
+    eachDataPointValueByteSize = eachDataPointValueByteSize_;
+  }
+
+  function _extractDataPointsDetailsForDataPackageMultiSign (uint256 calldataNegativeOffsetForDataPackage, uint256 signatureCount) internal pure returns (uint256 dataPointsCount, uint256 eachDataPointValueByteSize)
+  {
+    uint24 dataPointsCount_;
+
+    uint32 eachDataPointValueByteSize_;
+
+    uint256 negativeCalldataOffset = calldataNegativeOffsetForDataPackage + signatureCount * SIG_BS + SIGNERS_COUNT_BS;
+
+    uint256 calldataOffset = msg.data.length.sub(negativeCalldataOffset + STANDARD_SLOT_BS);
+    assembly {
+      dataPointsCount_ := calldataload(calldataOffset)
+    }
+
+    calldataOffset = calldataOffset.sub(DATA_POINTS_COUNT_BS);
+    assembly {
+      eachDataPointValueByteSize_ := calldataload(calldataOffset)
+    }
+
     dataPointsCount = dataPointsCount_;
     eachDataPointValueByteSize = eachDataPointValueByteSize_;
   }
--- a/packages/evm-connector/contracts/core/RedstoneConstants.sol
+++ b/packages/evm-connector/contracts/core/RedstoneConstants.sol
@@ -26,6 +26,7 @@ contract RedstoneConstants {
   uint256 internal constant SIG_BS = 65;
   uint256 internal constant TIMESTAMP_BS = 6;
   uint256 internal constant DATA_PACKAGES_COUNT_BS = 2;
+  uint256 internal constant SIGNERS_COUNT_BS = 2;
   uint256 internal constant DATA_POINTS_COUNT_BS = 3;
   uint256 internal constant DATA_POINT_VALUE_BYTE_SIZE_BS = 4;
   uint256 internal constant DATA_POINT_SYMBOL_BS = 32;
--- a/packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
+++ b/packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
@@ -3,6 +3,7 @@
 pragma solidity ^0.8.4;
 
 import "@openzeppelin/contracts/utils/math/SafeMath.sol";
+import "hardhat/console.sol";
 
 import "./RedstoneConstants.sol";
 import "./RedstoneDefaultsLib.sol";
@@ -61,6 +62,115 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
 
   /* ========== FUNCTIONS WITH IMPLEMENTATION (CAN NOT BE OVERRIDDEN) ========== */
 
+  function _securelyExtractOracleValuesFromTxMsgMultiSign(bytes32[] memory dataFeedIds)
+    internal
+    view
+    returns (uint256[] memory)
+  {
+    uint256 calldataNegativeOffset = _extractByteSizeOfUnsignedMetadata();
+
+    uint256 signersCount = _extractDataPackageSignersCountFromCalldata(calldataNegativeOffset);
+    bytes32 signedHash;
+
+    // Extracting the number of data points
+    (
+      uint256 dataPointsCount,
+      uint256 eachDataPointValueByteSize
+    ) = _extractDataPointsDetailsForDataPackageMultiSign(calldataNegativeOffset, signersCount);
+
+    console.log("dataPointsCount: %s", dataPointsCount);
+    console.log("eachDataPointValueByteSize: %s", eachDataPointValueByteSize);
+
+    {
+      bytes memory signedMessage;
+      uint48 extractedTimestamp;
+
+      uint256 signaturesBytesCount = signersCount.mul(SIG_BS);
+      uint256 signedMessageBytesCount = dataPointsCount.mul(
+        eachDataPointValueByteSize + DATA_POINT_SYMBOL_BS
+      ) + DATA_PACKAGE_WITHOUT_DATA_POINTS_AND_SIG_BS; //DATA_POINT_VALUE_BYTE_SIZE_BS + TIMESTAMP_BS + DATA_POINTS_COUNT_BS
+
+      uint256 timestampCalldataOffset = msg.data.length.sub(
+        calldataNegativeOffset +
+          SIGNERS_COUNT_BS +
+          signaturesBytesCount +
+          DATA_POINT_VALUE_BYTE_SIZE_BS +
+          DATA_POINTS_COUNT_BS
+      );
+
+      uint256 signedMessageCalldataOffset = msg.data.length.sub(
+        calldataNegativeOffset + SIGNERS_COUNT_BS + signaturesBytesCount + signedMessageBytesCount
+      );
+
+      assembly {
+        // Extracting the signed message
+        signedMessage := extractBytesFromCalldata(
+          signedMessageCalldataOffset,
+          signedMessageBytesCount
+        )
+
+        // Hashing the signed message
+        signedHash := keccak256(add(signedMessage, BYTES_ARR_LEN_VAR_BS), signedMessageBytesCount)
+
+        extractedTimestamp := calldataload(timestampCalldataOffset)
+
+        function initByteArray(bytesCount) -> ptr {
+          ptr := mload(FREE_MEMORY_PTR)
+          mstore(ptr, bytesCount)
+          ptr := add(ptr, BYTES_ARR_LEN_VAR_BS)
+          mstore(FREE_MEMORY_PTR, add(ptr, bytesCount))
+        }
+
+        function extractBytesFromCalldata(offset, bytesCount) -> extractedBytes {
+          let extractedBytesStartPtr := initByteArray(bytesCount)
+          calldatacopy(extractedBytesStartPtr, offset, bytesCount)
+          extractedBytes := sub(extractedBytesStartPtr, BYTES_ARR_LEN_VAR_BS)
+        }
+      }
+
+      validateTimestamp(extractedTimestamp);
+    }
+
+    {
+      // Validating signatures
+      for (uint256 i = 0; i < signersCount; i++) {
+        uint256 signatureCalldataOffset = calldataNegativeOffset +
+          SIGNERS_COUNT_BS +
+          (i + 1).mul(SIG_BS);
+
+        address signerAddress = SignatureLib.recoverSignerAddress(
+          signedHash,
+          signatureCalldataOffset
+        );
+        getAuthorisedSignerIndex(signerAddress);
+      }
+    }
+
+    {
+      // Extracting data points values
+      bytes32 dataPointDataFeedId;
+      uint256 dataPointValue;
+      uint256[] memory dataPointsValues = new uint256[](dataFeedIds.length);
+
+      for (uint256 dataPointIndex = 0; dataPointIndex < dataPointsCount; dataPointIndex++) {
+        (dataPointDataFeedId, dataPointValue) = _extractDataPointValueAndDataFeedIdMultiSign(
+          calldataNegativeOffset,
+          eachDataPointValueByteSize,
+          dataPointIndex,
+          signersCount
+        );
+
+        for (uint256 i = 0; i < dataFeedIds.length; i++) {
+          if (dataFeedIds[i] == dataPointDataFeedId) {
+            dataPointsValues[i] = dataPointValue;
+          }
+        }
+      }
+
+      return dataPointsValues;
+    }
+  }
+
   /**
    * @dev This is an internal helpful function for secure extraction oracle values
    * from the tx calldata. Security is achieved by signatures verification, timestamp
@@ -81,7 +191,7 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
     returns (uint256[] memory)
   {
     // Initializing helpful variables and allocating memory
-    uint256[] memory uniqueSignerCountForDataFeedIds = new uint256[](dataFeedIds.length);
+    uint256[] memory uniqueSignerCountForDataFeedIds = new uint256[](dataFeedIds.length); // trivial
     uint256[] memory signersBitmapForDataFeedIds = new uint256[](dataFeedIds.length);
     uint256[][] memory valuesForDataFeeds = new uint256[][](dataFeedIds.length);
     for (uint256 i = 0; i < dataFeedIds.length; i++) {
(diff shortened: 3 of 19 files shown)


packages/protocol/src/data-package/MultiSignDataPackage.ts
```diff
@@ -0,0 +1,104 @@
...
+{
+  public readonly signatures: Signature[];
+
+  constructor(
+    public readonly dataPackage: DataPackage,
+    signatures: Signature[] | string[]
+  ) {
+    super();
+    // TODO: improve this
+    if (typeof signatures[0] === "string") {
```
hatskier, line 36: Why can't we move the type checking into the `map` function?

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -61,6 +62,115 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
...
+      for (uint256 i = 0; i < signersCount; i++) {
+        uint256 signatureCalldataOffset = calldataNegativeOffset +
+          SIGNERS_COUNT_BS +
+          (i + 1).mul(SIG_BS);
+
+        address signerAddress = SignatureLib.recoverSignerAddress(
+          signedHash,
+          signatureCalldataOffset
+        );
+        getAuthorisedSignerIndex(signerAddress);
```
hatskier, line 145: Do we check signer addresses for duplicates?

jkurdek: Added this functionality, now only unique addresses are counted

packages/protocol/src/redstone-payload/MultiSignRedstonePayload.ts
```diff
@@ -0,0 +1,61 @@
...
+} from "../common/redstone-constants";
+import { Serializable } from "../common/Serializable";
+import { convertIntegerNumberToBytes } from "../common/utils";
+import {
+  RedstonePayloadParser,
+  RedstonePayloadParsingResult,
+} from "./RedstonePayloadParser";
+import { MultiSignDataPackage } from "../data-package/MultiSignDataPackage";
+
+export class MultiSignRedstonePayload extends Serializable {
```
hatskier, line 14: IMO RedstonePayload should stay the same and should contain unsigned information (in unsigned metadata) of what version to select (MultiSign or Old)

packages/protocol/src/data-package/MultiSignDataPackage.ts
```diff
@@ -0,0 +1,92 @@
...
+    return recoverSignerPublicKeys(this);
+  }
+
+  recoverSignersAddresses(): string[] {
+    return recoverSignerAddresses(this);
+  }
+
+  toBytes(): Uint8Array {
+    return concat([
+      this.dataPackage.toBytes(),
```
hatskier, line 67: I would add `this.serializeSignaturesCount()` here instead of the `serializeSignaturesToBytes` function, so it's easuer to understand the new structure by reading this function

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -61,6 +61,136 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
...
+    uint256 uniqueSignersThreshold = getUniqueSignersThreshold();
+
+    {
+      // Validating signatures
+      uint256 uniqueSignersCount = 0;
+      uint256 singersBitmap = 0;
+      for (uint256 i = 0; i < signersCount; i++) {
+        uint256 signatureCalldataOffset = calldataNegativeOffset +
+          SIGNERS_COUNT_BS +
+          (i + 1).mul(SIG_BS);
```
hatskier, line 139: let's accumulate the offset instead of multiplying `i + 1`

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -61,6 +61,136 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
...
+      // Extracting data points values
+      bytes32 dataPointDataFeedId;
+      uint256 dataPointValue;
+      uint256[] memory dataPointsValues = new uint256[](dataFeedIds.length);
+
+      for (uint256 dataPointIndex = 0; dataPointIndex < dataPointsCount; dataPointIndex++) {
+        (dataPointDataFeedId, dataPointValue) = _extractDataPointValueAndDataFeedIdMultiSign(
+          calldataNegativeOffset,
+          eachDataPointValueByteSize,
+          dataPointIndex,
```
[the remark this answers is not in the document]
hatskier: Is signersCount required here?

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -61,6 +61,136 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
...
+          signersCount
+        );
+
+        for (uint256 i = 0; i < dataFeedIds.length; i++) {
+          if (dataFeedIds[i] == dataPointDataFeedId) {
+            if (!BitmapLib.getBitFromBitmap(dataFeedIdsBitmap, i)) {
+              dataFeedIdsBitmap = BitmapLib.setBitInBitmap(dataFeedIdsBitmap, i);
+              uniqueDataFeedIdsCount++;
```
hatskier, line 178: I suppose this implementation of calculating `uniqueDataFeedIdsCount` is not correct, as it counts indexes (not data feed ids)

hatskier (changes requested):

packages/evm-connector/benchmarks/multi-sign-packages-benchmark.json
```diff
@@ -0,0 +1,236 @@
...
+      "forAttachingDataToCalldataMultiSign": 18444,
+      "forDataExtractionAndVerificationMultiSign": 152897
+    },
+    "10 signers, 20 symbols, 20 points": {
+      "forAttachingDataToCalldata": 88320,
+      "forDataExtractionAndVerification": 892127,
+      "forAttachingDataToCalldataMultiSign": 18444,
+      "forDataExtractionAndVerificationMultiSign": 152897
+    }
+  }
```
hatskier, line 236: Please add a newline here

packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -66,6 +66,20 @@ contract CalldataExtractor is RedstoneConstants {
...
+
+    if (calldataNegativeOffsetWithStandardSlot > msg.data.length) {
+      revert CalldataOverOrUnderFlow();
+    }
+    assembly {
+      signerCount := calldataload(
+        sub(calldatasize(), calldataNegativeOffsetWithStandardSlot)
+      )
+    }
+    return signerCount;
```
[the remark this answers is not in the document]
jkurdek: It was unnecessary. Removed.

packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -110,4 +141,26 @@ contract CalldataExtractor is RedstoneConstants {
...
+  {
+    uint24 dataPointsCount_;
+
+    uint32 eachDataPointValueByteSize_;
+
+    uint256 negativeCalldataOffset = calldataNegativeOffsetForDataPackage + signatureCount * SIG_BS + SIGNERS_COUNT_BS;
+
+    uint256 calldataOffset = msg.data.length.sub(negativeCalldataOffset + STANDARD_SLOT_BS);
+    assembly {
+      dataPointsCount_ := calldataload(calldataOffset)
```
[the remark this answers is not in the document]
jkurdek: This implementation mirrors the old implementation of extracting data point details function. dataPointCount is 3 bytes, eachDataPointValueByteSize is 4 bytes, so this is necessary to get the correct values inside uint256. 

We could maybe return both of them as respectively uint24 and uint32, which would simplify the new and old implementations. What do you think?

packages/protocol/src/data-package/MultiSignDataPackage.ts
```diff
@@ -0,0 +1,92 @@
...
+  ) {
+    super();
+    this.signatures = signatures.map((signature) =>
+      typeof signature === "string" ? splitSignature(signature) : signature
+    );
+  }
+
+  serializeSignaturesToBytes(): Uint8Array {
+    const signaturesBytes = concat(
+      this.serializeSignaturesToHex().map((signature) => arrayify(signature))
```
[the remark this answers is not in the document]
jkurdek: This raises a type issue, probably because arrayify expects two arguments (one nullable).

packages/cache-service/test/e2e/data-packages.spec.ts
```diff
@@ -259,8 +260,9 @@ describe("Data packages (e2e)", () => {
     const payload = new RedstonePayloadParser(payloadBytes).parse();
     expect(payload.signedDataPackages.length).toBe(expectedDataPackagesLength);
 
-    const signedDataPackage = payload.signedDataPackages[0];
+    const signedDataPackage: SignedDataPackage = payload.signedDataPackages[0] as SignedDataPackage;
```
hatskier, line 263: Why have you added it?

packages/protocol/test/plain-obj-serialization.test.ts
```diff
@@ -102,6 +115,29 @@ describe("Fixed size data package", () => {
...
+    const serializedPlainObj = signedDataPackage.toObj();
+    expect(serializedPlainObj).toEqual({
+      dataPoints,
+      timestampMilliseconds: TIMESTAMP_FOR_TESTS,
+      signatures: [
+        "NX5yd/Cs8HzVdNchrM59uOoSst7n9KK5Ou9pA6S5GTM0RwghGlFjA0S+SVfb85ipg4HzUTKATBZSqPXlWldEEhw=",
+        "c4JkY5+tYDdxbtn42Ky7afAp/60TWihf8AqWEhIwqDZW5WALQ4cNB+RIJOFoHL7fS1me18LrVqI0yKJzsx/z/Bs=",
+      ],
+    });
+    const deserializedSignedDataPackage = MultiSignDataPackage.fromObj(
```
hatskier, line 135: `deserializedSignedDataPackage` -> `signedDataPackageHex`?

packages/protocol/test/plain-obj-serialization.test.ts
```diff
@@ -149,4 +185,54 @@ describe("Fixed size data package", () => {
...
+        },
+      ],
+      timestampMilliseconds: TIMESTAMP_FOR_TESTS,
+      signatures: [
+        "WJ+EFIe6pSwzCRcjKWIYMWyhmtKJP+tN2aI55+Ip5w4osIGH0ngUEjTO4b7sAPBGd5MIv11qbPaFxWReppbGDRs=",
+        "iVfU2iRNiNSblSqjCzJ7KaACnSpTccHbWsa+BGTTXcIwYvP6Ae90apjRU/FCA1s7oRjTnLN0ZAKN5QvQKbpfLxw=",
+      ],
+    });
+    const deserializedSignedDataPackage = MultiSignDataPackage.fromObj(
+      JSON.parse(JSON.stringify(serializedPlainObj))
```
hatskier, line 234: Why do you do this trick?

packages/protocol/test/redstone-payload.test.ts
```diff
@@ -63,6 +71,25 @@ describe("Fixed size data package", () => {
...
+      multiSignedDataPackages,
+      UNSIGNED_METADATA,
+    );
+
+    expect(serializedHex).toBe(
+      EXPECTED_SERIALIZED_DATA_PACKAGE +
+        EXPECTED_SIGNATURES[0] +
+        EXPECTED_SIGNATURES[1] +
+        "0002" + // signatures count
+        "0001" + // data packages count
```
hatskier, line 85: Why do we need this? Isn't is always equal 1?

packages/protocol/test/redstone-payload.test.ts
```diff
@@ -63,6 +71,25 @@ describe("Fixed size data package", () => {
...
+        "0002" + // payload version
```
hatskier, line 86: Can you also update the logic of the function for the standard redstone payload (a sinle-signature data packages). We would also like to test if we set the correct payload version there

packages/protocol/src/redstone-payload/RedstonePayload.ts
```diff
@@ -14,22 +18,74 @@ import {
...
   ): string {
     return new RedstonePayload(
       signedDataPackages,
-      unsignedMetadata
+      unsignedMetadata,
+      SINGLESIGN_REDSTONE_PAYLOAD_VERSION
     ).toBytesHexWithout0xPrefix();
   }
 
+  private static getVersion(signedDataPackages: SignedDataPackage[] | MultiSignDataPackage[])
```
hatskier, line 71: Maybe `getVersion`  -> `getRedstonePayloadVersion`?

packages/protocol/src/redstone-payload/RedstonePayload.ts
```diff
@@ -14,22 +18,74 @@ import {
 
 export class RedstonePayload extends Serializable {
   constructor(
-    public readonly signedDataPackages: SignedDataPackage[],
-    public readonly unsignedMetadata: string
+    public readonly signedDataPackages:
+      | SignedDataPackage[]
+      | MultiSignDataPackage[],
```
hatskier, line 23: Do we really want to have an array of multi=sign data packages?

packages/protocol/src/redstone-payload/RedstonePayload.ts
```diff
@@ -14,22 +18,74 @@ import {
...
+    public readonly unsignedMetadata: string,
+    public readonly version: number
```
hatskier, line 25: I don't think this constructor needs to have the version argument, because it can be determined based on the type of the `signedDataPackages` argument

hatskier: It will make impossible the case with inconsistent version and `signedDataPackages` type

packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -15,6 +15,24 @@ import "./RedstoneConstants.sol";
...
+    uint256 unsignedMetadataByteSize = calldataNegativeOffset - UNSIGNED_METADATA_BYTE_SIZE_BS - REDSTONE_MARKER_BS;
+
+    // Version 1 does not have a version field in the unsigned metadata
+    if (unsignedMetadataByteSize < VERSION_BS) {
+      return 1;
+    }
+
+    uint16 version;
+    assembly {
+      let versionOffset := sub(calldatasize(), add(sub(calldataNegativeOffset, VERSION_BS), STANDARD_SLOT_BS))
```
hatskier, line 29: I think this operation can be optimised `add(sub(calldataNegativeOffset, VERSION_BS), STANDARD_SLOT_BS))`. Also I would propose to do it on Solidity level (not assembly level) to benefit from the built-in SafeMath checks for over and underflows

packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -15,6 +15,24 @@ import "./RedstoneConstants.sol";
 contract CalldataExtractor is RedstoneConstants {
   using SafeMath for uint256;
 
+  // Extracts Redstone payload version from calldata. Used to differentiate between single and multi sign approach
+  function _extractVersionFromUnsignedMetadata(uint256 calldataNegativeOffset) internal pure returns (uint256) {
```
hatskier, line 19: I would call this argument not `calldataNegativeOffset` but `unsignedMetadataByteSize`

packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -15,6 +15,24 @@ import "./RedstoneConstants.sol";
...
+  // Extracts Redstone payload version from calldata. Used to differentiate between single and multi sign approach
+  function _extractVersionFromUnsignedMetadata(uint256 calldataNegativeOffset) internal pure returns (uint256) { 
+    uint256 unsignedMetadataByteSize = calldataNegativeOffset - UNSIGNED_METADATA_BYTE_SIZE_BS - REDSTONE_MARKER_BS;
+
+    // Version 1 does not have a version field in the unsigned metadata
+    if (unsignedMetadataByteSize < VERSION_BS) {
+      return 1;
+    }
+
+    uint16 version;
```
hatskier, line 27: IMO, it would be better to use `VERSION_BS` and bitwise operators to get value from 2 bytes, because it can cause inconsistencies if e.g. we update VERSION_BS without updating `uint16`

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -287,7 +516,8 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
       if (uniqueSignerCountForDataFeedIds[dataFeedIndex] < uniqueSignersThreshold) {
         revert InsufficientNumberOfUniqueSigners(
           uniqueSignerCountForDataFeedIds[dataFeedIndex],
-          uniqueSignersThreshold);
+          uniqueSignersThreshold
+        );
```
hatskier, line 520: Why do you add formatting changes in this PR?

packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -15,6 +15,24 @@ import "./RedstoneConstants.sol";
 contract CalldataExtractor is RedstoneConstants {
   using SafeMath for uint256;
 
+  // Extracts Redstone payload version from calldata. Used to differentiate between single and multi sign approach
+  function _extractVersionFromUnsignedMetadata(uint256 calldataNegativeOffset) internal pure returns (uint256) { 
+    uint256 unsignedMetadataByteSize = calldataNegativeOffset - UNSIGNED_METADATA_BYTE_SIZE_BS - REDSTONE_MARKER_BS;
+
+    // Version 1 does not have a version field in the unsigned metadata
+    if (unsignedMetadataByteSize < VERSION_BS) {
+      return 1;
```
hatskier, line 24: Here we can revert

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -61,6 +61,238 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
...
+   *
+   * Note that this function should not be called from a consumer contract. Instead, use `getOracleNumericValuesFromTxMsg`
+   * or `getOracleNumericValueFromTxMsg` as appropriate.
+   *
+   * @param dataFeedIds An array of unique data feed identifiers.
+   * @return An array of the extracted and verified oracle values in the same order as the requested data feed identifiers.
+   */
+  function _securelyExtractOracleValuesFromTxMsg(
+    bytes32[] memory dataFeedIds
+  ) internal view returns (uint256[] memory) {
```
hatskier, line 77: Just noticed, that we may probably change visiblity of this function from internal to private

packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -66,6 +84,20 @@ contract CalldataExtractor is RedstoneConstants {
     return dataPackagesCount;
   }
 
+  // Used for multi sign approach
+  function _extractDataPackageSignersCountFromCalldata(uint256 calldataNegativeOffset) internal pure returns (uint16 signerCount) {
```
hatskier, line 88: Same, maybe use bitwise operations to get SIGNERS_COUNT_BS bytes?

packages/evm-connector/contracts/core/RedstoneConstants.sol
```diff
@@ -26,17 +26,22 @@ contract RedstoneConstants {
   uint256 internal constant SIG_BS = 65;
   uint256 internal constant TIMESTAMP_BS = 6;
   uint256 internal constant DATA_PACKAGES_COUNT_BS = 2;
+  uint256 internal constant SIGNERS_COUNT_BS = 2;
```
hatskier, line 29: Maybe rename SIGNERS_COUNT_BS -> MULTI_ SIGNERS_COUNT_BS?

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -61,6 +61,238 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
...
+    uint256 signersCount = _extractDataPackageSignersCountFromCalldata(calldataNegativeOffset);
+
+    calldataNegativeOffset += SIGNERS_COUNT_BS;
+    uint256 signaturesByteSize = signersCount.mul(SIG_BS);
+
+    (
+      uint256 dataPointsCount,
+      uint256 eachDataPointValueByteSize
+    ) = _extractDataPointsDetailsForDataPackageMultiSign(
+        calldataNegativeOffset + signaturesByteSize
```
hatskier, line 116: We use the same expression (`calldataNegativeOffset + signaturesByteSize`) many times. Maybe just increase `calldataNegativeOffset` instead?

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -61,6 +61,238 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
...
+   * keccak256, and returns the resulting hash.
+   *
+   * Note that this function should not be called from a consumer contract. It is intended for internal use only.
+   *
+   * @param calldataNegativeOffset The negative offset of the transaction calldata.
+   * @param dataPointsCount The number of data points in the data package.
+   * @param eachDataPointValueByteSize The byte size of each data point value in the data package.
+   * @return signedHash The keccak256 hash of the signed message extracted from the data package.
+   */
+  function _extractSignedHash(
```
hatskier, line 189: Can you please check how much simpler implementation is more expensive in terms of gas. I mean smth like `return msg.data[calldataNegativeOffset:calldataNegativeOffset + signedMessageBytesCount]`

packages/evm-connector/contracts/core/CalldataExtractor.sol
```diff
@@ -82,6 +114,18 @@ contract CalldataExtractor is RedstoneConstants {
     }
   }
 
+  // Used for multi sign approach
+  function _extractDataPointValueAndDataFeedIdMultiSign(
```
hatskier, line 118: Do we really need a separate function for it? For me it looks like the logic of extracting dataFeed value and id should be similar for multi-sig and single-sig approaches

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -61,6 +61,238 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
...
+        dataPointNegativeOffset =
+          dataPointNegativeOffset +
+          eachDataPointValueByteSize +
+          DATA_POINT_SYMBOL_BS;
+
+        (dataPointDataFeedId, dataPointValue) = _extractDataPointValueAndDataFeedIdMultiSign(
+          dataPointNegativeOffset
+        );
+
+        for (uint256 i = 0; i < dataFeedIds.length; i++) {
```
hatskier, line 161: I would propose to rename `i` to the `requestedDataFeedId`

hatskier (changes requested):

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
jkurdek: I do not think it is possible, as it would become unaccessible to contracts iheriting from this one

packages/evm-connector/contracts/core/CalldataExtractor.sol
jkurdek: Reverted to single function for both approaches. This is slightly (10%) less optimal in terms of gas usage.

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
jkurdek: We need original calldataNegativeOffset when calling _validateSignatures

PR #394, opened 2023-02-24
feat: added an on chain relayer for eth on starknet

Lukasz2891:


packages/cairo-connector/data/test.hex
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+0x455448000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000025869aed8601867a4e578000000020000001883135efe31173469737a3d603f053c7555f7eea52646a2d3ef3e01b112234c0427a570e9af786d5fecd1f231714c11f627fa70f38c81db0730a286f77c1c5ed1c0001000000000002ed57011e0000
```
cehali, line 1: Please add new line

packages/cairo-connector/data/test.json
```diff
@@ -1 +1 @@
...
+"statusCode":500,"message":"Internal server error"
```
cehali, line 1: Why does `test.json` look like an error? Maybe rename a file?

packages/cairo-connector/src/contracts/price_manager.cairo
```diff
@@ -0,0 +1,121 @@
...
+}(
+    rndNumber: felt,
+    feed_ids_len: felt,
+    feed_ids: felt*,
+    payload_data_len: felt,
+    payload_data: felt*,
+) {
+    alloc_locals;
+
+    assert_owner();
```
cehali, line 95: I am wondering if it is the same as in the solidity contract. We want to PriceFeed be updated only by PriceManager and PriceManager function to update prices can be called by anyone

packages/cairo-connector/src/contracts/price_feed.cairo
```diff
@@ -0,0 +1,84 @@
...
+@view
+func latest_round_data{
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, bitwise_ptr: BitwiseBuiltin*, range_check_ptr
+}() -> (round: Round) {
+    alloc_locals;
+
+    let (contract_address) = manager_contract_address.read();
+    let (feed_id) = feed_identifier.read();
+
+    let (package_ts, round_id) = IPriceManager.read_timestamp_and_round(
```
cehali, line 36: Good idea to get those values and add them to return. We should do the same in solidity contract

packages/cairo-connector/data/test.json
Lukasz2891: sample data files are not needed anymore

packages/cairo-connector/src/contracts/price_manager.cairo
Lukasz2891: I've asked Alex about that (calling manager from price feed) - so we can change it, but Idk if it's important for now ;)

packages/starknet-relayer/src/starknet/StarknetDataProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+    return this.requestParams.dataFeeds.map(
+      // @ts-ignore
+      (dataFeedId) => StarknetDataProvider.dataFeedIdNumber[dataFeedId]
+    );
+  }
+
+  private static preparePayloadData(payloadHex: string) {
+    payloadHex = payloadHex.startsWith("0x") ? payloadHex : `0x${payloadHex}`;
+
+    let payloadData = [];
```
hatskier, line 47: `let` can be changed to `const` here since you don't change the variable directly and only set values for indexes

packages/cairo-connector/src/contracts/price_feed.cairo
```diff
@@ -0,0 +1,84 @@
...
+    );
+
+    return (round=round);
+}
+
+@view
+func round_data{
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, bitwise_ptr: BitwiseBuiltin*, range_check_ptr
+}(round_id: felt) -> (round: Round) {
+    with_attr error_message("Method not allowed. Use latest_round_data instead.") {
```
hatskier, line 60: Hahaha, a very interesting way to always throw an error

packages/cairo-connector/src/contracts/price_feed.cairo
```diff
@@ -0,0 +1,84 @@
...
+}() -> (round: Round) {
+    alloc_locals;
+
+    let (contract_address) = manager_contract_address.read();
+    let (feed_id) = feed_identifier.read();
+
+    let (package_ts, round_id) = IPriceManager.read_timestamp_and_round(
+        contract_address=contract_address
+    );
+    let (block_ts) = IPriceManager.get_block_timestamp(contract_address=contract_address);
```
hatskier, line 39: Shouldn't we use `read_timestamp_and_round` here instead of `get_block_timestamp `  and `get_block_number`?

packages/cairo-connector/src/contracts/price_manager.cairo
```diff
@@ -0,0 +1,121 @@
...
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, bitwise_ptr: BitwiseBuiltin*, range_check_ptr
+}() -> (timestamp: felt, round: felt) {
+    let (ts) = timestamp.read();
+    let (rnd) = round_number.read();
+
+    return (timestamp=ts, round=rnd);
+}
+
+@view
+func get_block_timestamp{
```
hatskier, line 68: Can this function be removed?

packages/cairo-connector/src/contracts/price_manager.cairo
```diff
@@ -0,0 +1,121 @@
...
+    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, bitwise_ptr: BitwiseBuiltin*, range_check_ptr
+}() -> (res: felt) {
+    let (res) = syscalls_get_block_timestamp();
+    return (res=res);
+}
+
+@view
+func get_block_number{
```
hatskier, line 76: Can this function be removed?

packages/cairo-connector/src/contracts/price_manager.cairo
hatskier: As we discussed on the call, current model is better. PriceManager can store values and PriceFeed contracts will query from its PriceManager's storage. This model is even more secure, because of its permissionless nature

hatskier, line 95: I would allow everyone to update the prices (not only the owner)

packages/cairo-connector/src/contracts/price_manager.cairo
```diff
@@ -0,0 +1,121 @@
...
+
+    let (savedround) = round_number.read();  // Cannot evaluate ap-based or complex references: ['saved_round']
+    with_attr error_message(
+            "The increased by 1 saved round#: #{savedround} doesn't convene with the passed round#: {rndNumber}") {
+        assert rndNumber = savedround + 1;
+    }
+
+    let requested_feed_ids = Array(ptr=feed_ids, len=feed_ids_len);
+
+    let (values, ts) = get_aggregated_values(
```
hatskier, line 105: How is this `ts` value is calculated? Is it the minimum timestamp each package's timestamp?

packages/data-generator/Makefile
```diff
@@ -0,0 +1,21 @@
+PAYLOAD_URL=http://localhost:3000/data-packages/payload?unique-signers-count=2&data-service-id=redstone-avalanche-prod&format=
+#PAYLOAD_URL=https://d33trozg86ya9x.cloudfront.net/data-packages/payload?data-packages/payload?unique-signers-count=1&data-service-id=redstone-rapid-demo&data-feed-id=BTC,ETH,BNB,AR,AVAX,CELO&format=
```
hatskier, line 2: Should we remove the commented version?

packages/data-generator/.gitignore
```diff
@@ -0,0 +1 @@
+data/
```
hatskier, line 1: Why should we create a separate package for the data generation? Also, the name `data-generator` is too generic. I would call it payload generator and would probably move it to the redstone-protocol package (e.g. to the scripts folder)

hatskier (changes requested):

packages/starknet-relayer/src/start-relayer.ts
```diff
@@ -0,0 +1,66 @@
+import { config } from "./config";
+import { Starknet } from "./starknet/Starknet";
+import { FEE_MULTIPLIER } from "./starknet/StarknetCommand";
+
+(() => {
```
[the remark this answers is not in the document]
Lukasz2891: generally - there are 2 logics here ;)
1) on-chain relayers
2) chains 

The second one should be extracted to have it reused also in the showroom purposes.

packages/cairo-connector/src/contracts/price_manager.cairo
Lukasz2891: it's just ```payload.data_packages.ptr[0].timestamp```

should we use `min` here?

packages/cairo-connector/src/contracts/price_feed.cairo
hatskier: Oh right. Then I think we should have a function for extracting all details about the latest update (including round id, data package timestamp, block timestamp, block number, and round number)

packages/cairo-connector/src/contracts/price_manager.cairo
hatskier: Yes, for sure. Otherwise it's possible to provide many outdated packages but put the newest one into the 0th index

packages/cairo-connector/src/contracts/price_manager.cairo
Lukasz2891: it can be removed ;) so it should be moved to the feed.

packages/starknet-relayer/src/starknet/StarknetDataProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+    for (let i = 2; i < Array.from(payloadHex).length - 1; i++) {
```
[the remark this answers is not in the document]
Lukasz2891: It's not needed ;)

packages/cairo-connector/src/contracts/price_feed.cairo
```diff
@@ -0,0 +1,86 @@
...
+}() -> (round: Round) {
+    alloc_locals;
+
+    let (contract_address) = manager_contract_address.read();
+    let (feed_id) = feed_identifier.read();
+
+    let (package_ts, round_id) = IPriceManager.read_timestamp_and_round(
+        contract_address=contract_address
+    );
+    let (block_ts) = get_block_timestamp();
```
hatskier, line 41: Shouldn't we return the block number and the blok timestamp of the latest price update?

hatskier (changes requested):

packages/cairo-connector/src/contracts/price_manager.cairo
Lukasz2891: still we verify the packages, so outdated means here "not earlier than 15 minutes before the block-timestamp" :) but ok ;p

packages/cairo-connector/src/contracts/price_feed.cairo
Lukasz2891: hm, that means we should move the `get_block_timestamp` back to the manager ;)

hatskier (approved):
LGTM!

PR #400, opened 2023-03-02
feat: add on chain fetchers dexes with uniswap v3

rucinpk:


hatskier (changes requested):
- Can pool address be changed? If not, can we hardcode its address?
- We should fetch liquidity data from Uniswap V3 in exactly the same moment when we fetch price. We can utilise multi-call contracts for it. Don't use Promise.all
- What identifies the pool (token1, token2 and fee)? Can the fee value be other than 3000, 5000 and 10000? We should create fetchers for each existing pool
- Liquidity should be calculated in USD. Also, we should deeply understand how the liquidity is calculated? We should also think how to calculate liquidity calculated in a similar way to Uniswap V2
- Let's remove jsbi
- Let's remove awaits where they are not necessary
- Let's add *.abi.json instead of *.json for abis
- Let's check ig ethes.formatUnits or ethers.parseUnits can replace `fromReadableAmount` and JSBI
- Let's name constants so their names tell about their puropse (not just TEN for 10 ;))

packages/oracle-node/src/fetchers/uniswap-v3-on-chain/types.ts
```diff
@@ -0,0 +1,24 @@
+import { BigNumber } from "ethers";
```
hatskier, line 1: I would move the uniswap-v3-on-chain folder to evm-chains/ethereum/

packages/oracle-node/src/fetchers/uniswap-v3-on-chain/uniswap-v3-fetchers-config.json
```diff
@@ -0,0 +1,232 @@
+{
+  "uniswap-v3-on-chain-OHM": {
```
hatskier, line 2: Shouldn't we also include fee in the fetcher name and create fetchers for each available fee?

packages/oracle-node/test/fetchers/mocks/UniswapV3FactoryMock.json
```diff
@@ -0,0 +1,15 @@
+{
```
hatskier, line 1: Should we add `abi.json` to the filename?

packages/oracle-node/test/fetchers/mocks/UniswapV3PoolMock.json
```diff
@@ -0,0 +1,11 @@
+{
+  "abi": [
```
hatskier, line 2: Should we add `abi.json` to the filename?

packages/oracle-node/test/fetchers/mocks/UniswapV3QuoterMock.json
```diff
@@ -0,0 +1,49 @@
+{
```
hatskier, line 1: Should we add `abi.json` to the filename?

packages/oracle-node/tools/uniswap-v3-on-chain/get-pool-configs.ts
```diff
@@ -0,0 +1,156 @@
...
+  "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
+  "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599",
+  "0x6B175474E89094C44Da98b954EedeAC495271d0F",
+  "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+];
+
+const QUOTER_ADDRESS = "0x61fFE014bA17989E743c5F6cB21bF9697530B21e";
+const FACTORY_ADDRESS = "0x1F98431c8aD98523631AE4a59f267346ea31F984";
+
+const FEE_AMOUNTS = [10000, 3000, 500, 100];
```
hatskier, line 22: How do you know if other fees are not possible? Can you add a comment with explanation to this place in the code, ideally with the link to the source

yarn.lock
```diff
@@ -13849,6 +13849,11 @@ js-yaml@^3.13.1:
     argparse "^1.0.7"
     esprima "^4.0.0"
 
+jsbi@^4.3.0:
```
hatskier, line 13852: Let's remove jsbi dependency

packages/oracle-node/tools/uniswap-v3-on-chain/get-pool-configs.ts
```diff
@@ -0,0 +1,156 @@
+import { Contract, ethers } from "ethers";
+import { config } from "../../src/config";
+import { saveJSON } from "../common/fs-utils";
+import UniswapV3Factory from "./UniswapV3Factory.abi.json";
+import ERC20Token from "./ERC20Token.abi.json";
+
+const POOL_PoolConfigs_FILE_PATH = "uniswap-v3-fetchers-config.json";
```
hatskier, line 7: Let's have a consistent name for this file, e.g. POOL_CONFIGS_FILE_PATH

packages/oracle-node/tools/uniswap-v3-on-chain/get-pool-configs.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const UNISWAP_NAME_PREFIX = "uniswap-v3-on-chain-";
+
+const TOKEN_ADDRESSES = [
```
hatskier, line 11: Let's add comments with symbols to the addresses

packages/oracle-node/tools/uniswap-v3-on-chain/get-pool-configs.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const getPoolConfigs = async () => {
+  const poolConfigs: PoolConfigs = {};
+  const uniswapV3FactoryContract = new ethers.Contract(
+    FACTORY_ADDRESS,
+    UniswapV3Factory.abi,
+    provider
+  );
+
+  for (let tokenIndex = 0; tokenIndex < TOKEN_ADDRESSES.length; tokenIndex++) {
```
hatskier, line 60: Can't we iterate over addresses and compare them instead of indexes?
e.g.
```js
for (const tokenAddress1 of TOKEN_ADDRESSES) {
     for (const tokenAddress2 of TOKEN_ADDRESSES) {
         if (tokenAddress1 !== tokenAddress2) {...}
     }
}

```

packages/oracle-node/tools/uniswap-v3-on-chain/get-pool-configs.ts
```diff
@@ -0,0 +1,156 @@
...
+    );
+    const tokenPoolConfigs: TokenPoolConfigs = {};
+
+    for (
+      let pairedTokenIndex = 0;
+      pairedTokenIndex < TOKEN_ADDRESSES.length;
+      pairedTokenIndex++
+    ) {
+      if (tokenIndex !== pairedTokenIndex) {
+        const { poolAddress, foundFee } = await findPool(
```
hatskier, line 72: Do we have a guarantee that each given pair of symbols always will have only one pool (with one fee)?

packages/oracle-node/tools/uniswap-v3-on-chain/get-pool-configs.ts
```diff
@@ -0,0 +1,156 @@
...
+  const decimals = await getDecimals(tokenContract);
+  return { symbol, address, decimals };
+};
+
+const findPool = async (
+  address: string,
+  pairedAddress: string,
+  factory: Contract
+) => {
+  for (const fee of FEE_AMOUNTS) {
```
hatskier, line 142: Shouldn't we return all found pools? Not only the first one that we found?

packages/oracle-node/test/fetchers/uniswap-v3-on-chain.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+      "uniswap-v3-mock",
+      mockTokenConfig,
+      provider
+    );
+    await saveMockPriceInLocalDb(1570.82, "MockToken2");
+    const result = await fetcher.fetchAll(["MockToken"]);
+    expect(result).toEqual([{ symbol: "MockToken", value: 9.3863 }]);
+  });
+
+  test("should properly fetch liquidity", async () => {
```
hatskier, line 66: I am not sure if we should allow to fetch liquidity since we don't deeply understand the meaning of this value in Uni V3

packages/oracle-node/src/fetchers/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,116 @@
...
+import UniswapV3Quoter from "./UniswapV3Quoter.abi.json";
+import UniswapV3Pool from "./UniswapV3Pool.abi.json";
+import { ethers } from "ethers";
+import { getLastPrice } from "../../db/local-db";
+import { PoolsConfig, QuoterOutAmount } from "./types";
+
+import { isLiquidity, parseLiquidityDataFeedId } from "../liquidity/utils";
+
+const DOLLAR_IN_CENTS = 100;
+const DOLLAR_IN_CENTS_EXPONENT = 2;
```
hatskier, line 12: Why 2?

packages/oracle-node/src/fetchers/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,116 @@
...
+    name: string,
+    private readonly poolsConfig: PoolsConfig,
+    private readonly provider: providers.JsonRpcProvider
+  ) {
+    super(name);
+  }
+
+  async makeRequest(id: string): Promise<QuoterOutAmount> {
+    let assetId = id;
+    if (isLiquidity(id)) {
```
hatskier, line 26: This condition will never be met. `DexOnChainFetcher` is implemented in the way that the `makeRequest` function will always receive `spotAssetId`. Liquidity ids will be filtered out

packages/oracle-node/src/fetchers/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,116 @@
...
+    return {
+      amountOut: amountOut as BigNumber,
+      assetId: id,
+      outAssetId,
+      outAssetDecimals,
+      inAssetDecimals,
+      liquidity: liquidity as BigNumber,
+    };
+  }
+  calculateLiquidity(assetId: string, response: QuoterOutAmount): number {
```
hatskier, line 55: Please add the `override` keyword to improve readability

packages/oracle-node/src/fetchers/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,116 @@
...
+    return response.liquidity.toNumber();
+  }
+
+  calculateSpotPrice(assetId: string, quoterResponse: QuoterOutAmount) {
```
hatskier, line 59: Please add the `override` keyword to improve readability

packages/oracle-node/src/fetchers/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,116 @@
...
+      this.buildParamsForQuoter(
+        this.poolsConfig[assetId].token0Address,
+        this.poolsConfig[assetId].token1Address,
+        assetId
+      );
+
+    const quoterContract = this.getQuoter(assetId);
+    const poolContract = this.getPool(assetId);
+
+    const [amountOutResult, liquidity] = await Promise.all([
```
hatskier, line 40: It doesn't guarantee that the blockchain calls will be made for the same blockchain state. To be sure, we should use multicall. But it's for future.
For now, we can simply remove the liquidity fetching in Uni v3 and override the `calculateLiquidity` to always throw an error

packages/oracle-node/src/fetchers/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,116 @@
...
+    };
+  }
+  calculateLiquidity(assetId: string, response: QuoterOutAmount): number {
+    return response.liquidity.toNumber();
+  }
+
+  calculateSpotPrice(assetId: string, quoterResponse: QuoterOutAmount) {
+    const outAssetLastPrice = getLastPrice(quoterResponse.outAssetId)!.value;
+
+    const outAssetDecimalsExp = BigNumber.from(10).pow(
```
hatskier, line 62: @cehali can you please review the spot price calculation algorithm when you have a free minute?

hatskier (changes requested):

packages/oracle-node/test/fetchers/mocks/UniswapV3FactoryMock.json
rucinpk: In other mock files in the project do not have `abi.json` in their filename and are named with just `.json`

packages/oracle-node/src/fetchers/evm-chain/uniswap-v3-on-chain/all-uniswap-v3-fetchers.ts
```diff
@@ -0,0 +1,19 @@
+import uniswapV3FetchersConfig from "./uniswap-v3-fetchers-config.json";
+import { UniswapV3FetcherOnChain } from "./UniswapV3FetcherOnChain";
+import { ethers } from "ethers";
+import { config } from "../../../config";
+
+const uniswapV3Fetchers: Record<string, UniswapV3FetcherOnChain> = {};
+const provider = new ethers.providers.JsonRpcProvider(config.ethMainRpcUrl);
```
cehali, line 7: You could consider using `StaticJsonRpcProvider` here https://docs.ethers.org/v5/api/providers/jsonrpc-provider/#StaticJsonRpcProvider

packages/oracle-node/test/fetchers/mocks/UniswapV3FactoryMock.json
cehali, line 1: I am wondering why we need Mocked abis. They are only interfaces and they should be the same as the one deployed. I used to mock bytecodes but now I know it was bad idea and we should use https://ethereum-waffle.readthedocs.io/en/latest/mock-contract.html?highlight=mock

packages/oracle-node/test/fetchers/uniswap-v3-on-chain.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+  beforeAll(async () => {
+    setupLocalDb();
+    provider = new MockProvider();
+    const [wallet] = provider.getWallets();
+    quoterContract = await deployMockContract(wallet, UniswapV3QuoterMock.abi);
+    poolContract = await deployMockContract(wallet, UniswapV3Pool);
+
+    await poolContract.mock.liquidity.returns(123456789);
+
+    await quoterContract.mock.quoteExactInputSingle.returns(
```
cehali, line 30: Good idea, we will switch other tests to mock functions that return 0 right after deployment.

packages/oracle-node/test/fetchers/uniswap-v3-on-chain.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+  let mockTokenConfig: PoolsConfig;
+
+  beforeAll(async () => {
+    setupLocalDb();
+    provider = new MockProvider();
+    const [wallet] = provider.getWallets();
+    quoterContract = await deployMockContract(wallet, UniswapV3QuoterMock.abi);
+    poolContract = await deployMockContract(wallet, UniswapV3Pool);
+
+    await poolContract.mock.liquidity.returns(123456789);
```
cehali, line 28: If we don't know how to calculate liquidity I don't think this is neccessary

packages/oracle-node/src/fetchers/evm-chain/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,118 @@
...
+      inAssetDecimals,
+      liquidity: liquidity as BigNumber,
+    };
+  }
+
+  override calculateSpotPrice(
+    assetId: string,
+    quoterResponse: QuoterOutAmount
+  ) {
+    const outAssetLastPrice = getLastPrice(quoterResponse.outAssetId)!.value;
```
cehali, line 59: `outAssetLastPrice` could be undefined here. We should throw an error if so.

packages/oracle-node/src/fetchers/evm-chain/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,118 @@
...
+  private buildParamsForQuoter(
+    token0Address: string,
+    token1Address: string,
+    assetId: string
+  ) {
+    return {
+      quoterParams: {
+        tokenIn: token0Address,
+        tokenOut: token1Address,
+        amountIn: ethers.utils
```
cehali, line 107: Does this parameter take into account decimals? Or is it like `I want to change 1 OHM for ETH`? If the second option is correct I believe we could use here a fixed number and not `token0Decimals`.

rucinpk: Yes, the calculation must take into account the number of decimals for both token0 and token1.

packages/oracle-node/manifests/single-source/uniswap-v3-on-chain-WETH-10000.json
```diff
@@ -0,0 +1,19 @@
+{
+  "interval": 15000,
+  "priceAggregator": "median",
+  "defaultSource": ["uniswap-v3-on-chain-WETH-10000"],
```
hatskier, line 4: Can we call it uniswap-v3-on-chain-weth-10000 to be consistent with other fetchers' names?

packages/oracle-node/manifests/single-source/uniswap-v3-on-chain-WETH-10000.json
```diff
@@ -0,0 +1,19 @@
...
+    "deviationWithRecentValues": {
+      "maxPercent": 25,
+      "maxDelayMilliseconds": 300000
+    }
+  },
+  "tokens": {
+    "DAI": {},
+    "USDC": {},
+    "WETH": {
+      "source": ["uniswap"]
```
hatskier, line 16: Why only uniswap? Also, I think w should use ETH price for WETH.
Also, I think we should add `skipSigning` option here for WETH (or ETH) token

packages/oracle-node/src/fetchers/evm-chain/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,131 @@
...
+      secondsAgoStart,
+      secondsAgoEnd,
+    });
+    const multicallResult = await this.createMulticallInstance().call(
+      multicallContext
+    );
+    const poolConfig = this.poolsConfig[assetId];
+
+    // TODO how to convert this to something meaningful?
+    const liquidity = BigNumber.from(
```
hatskier, line 31: We can use he heura with slippage -> liquidity conversion

packages/oracle-node/src/fetchers/evm-chain/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,131 @@
...
+      multicallResult.results.poolContract.callsReturnContext[1]
+        .returnValues[0];
+    // twapTick is defined as log(1.0001, token0_sub_units/token1_sub_units)
+    const twapTick = BigNumber.from(tickCumulatives[1])
+      .sub(tickCumulatives[0])
+      .div(secondsAgoStart - secondsAgoEnd);
+    const decimalsDifferenceMultiplier = new Decimal(10).toPower(
+      poolConfig.token0Decimals - poolConfig.token1Decimals
+    );
+    const priceRatio = new Decimal(1.0001)
```
hatskier, line 45: What's this value 1.0001? Can we move it to a constant?

packages/oracle-node/src/fetchers/evm-chain/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,131 @@
...
+    );
+
+    const tickCumulatives: BigNumber[] =
+      multicallResult.results.poolContract.callsReturnContext[1]
+        .returnValues[0];
+    // twapTick is defined as log(1.0001, token0_sub_units/token1_sub_units)
+    const twapTick = BigNumber.from(tickCumulatives[1])
+      .sub(tickCumulatives[0])
+      .div(secondsAgoStart - secondsAgoEnd);
+    const decimalsDifferenceMultiplier = new Decimal(10).toPower(
```
hatskier, line 42: Can we use a constant (TEN_AS_BASE or smth similar), which we already have?

packages/oracle-node/src/fetchers/evm-chain/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
```diff
@@ -0,0 +1,131 @@
...
+      poolConfig.token0Decimals - poolConfig.token1Decimals
+    );
+    const priceRatio = new Decimal(1.0001)
+      .toPower(twapTick.toNumber())
+      .times(decimalsDifferenceMultiplier)
+      .toNumber();
+
+    const isSymbol1Current = poolConfig.token1Symbol === assetId;
+
+    this.logger.error(
```
hatskier, line 52: Why do we log with the `error` level here?

packages/oracle-node/manifests/single-source/uniswap-v3-on-chain-WETH-10000.json
rpiszczatowski: This was only to prove that the feature works. I'll add more realistic example with swETH. Athough I still don't know how e.g. uniswap-v2-on-chain-weth.json works without ETH price being pulled from somewhere.

packages/oracle-node/src/fetchers/evm-chain/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
rpiszczatowski: this is uniswap protocol-specific constant. I don't belive it has a name. I'll check it though.

packages/oracle-node/src/fetchers/evm-chain/uniswap-v3-on-chain/UniswapV3FetcherOnChain.ts
rpiszczatowski: This is for testing purposes. I'll get rid of log messages before merge.

hatskier: Can you get rid now, so we don't need to re-approve?

packages/oracle-node/manifests/single-source/uniswap-v3-on-chain-WETH-10000.json
hatskier: I would say we can add ETH price with few free sources here so that the oracle-node can be properly launched with this manifest

packages/oracle-node/manifests/single-source/uniswap-v3-on-chain-WETH-10000.json
rpiszczatowski: Sure. The name was added by original author of this PR.

PR #408, opened 2023-03-06
feat: improve price feeds contracts

cehali:


packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -2,15 +2,13 @@
...
-import "../data-services/MainDemoConsumerBase.sol";
 import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
-import "./PriceFeedsRegistry.sol";
-import "./PriceFeed.sol";
+import "../data-services/MainDemoConsumerBase.sol";
 
 contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
   uint256 public lastRound = 0;
   uint256 public lastUpdateTimestampMilliseconds = 0;
-  PriceFeedsRegistry public priceFeedRegistry;
```
hatskier, line 13: Can we remove the pricefeed registry contract now?

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -65,7 +59,11 @@ contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
     return (lastRound, lastUpdateTimestampMilliseconds);
   }
 
-  function updateDataFeedValues(uint256 proposedRound, uint256 proposedTimestamp) public {
+  function updateDataFeedValues(
+    uint256 proposedRound,
+    uint256 proposedTimestamp,
+    bytes32[] calldata dataFeedsIds
```
hatskier, line 65: If we now provide the `dataFeedsIds` argument, it looks like we easily can update only part of data feeds. So, a potential attacker can e.g. update only one data feed (updating lastUpdateTime), which will result in outdated values for all other data feeds.

I think we need to require in this function updating all the data feeds managed by the PriceFeedsManager.

hatskier (changes requested):

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -65,22 +90,57 @@ contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
...
 
-  function updateDataFeedValues(uint256 proposedRound, uint256 proposedTimestamp) public {
+  function getDataFeedsIds() public view returns (bytes32[] memory) {
+    return dataFeedsIds;
+  }
+
+  function updateDataFeedValues(
+    uint256 proposedRound,
+    uint256 proposedTimestamp,
+    bytes32[] calldata dataFeedsIdsToUpdate
```
hatskier, line 100: Why do we still need to pass this argument?

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -49,6 +52,28 @@ contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
     }
   }
 
+  function validateDataFeedsToUpdate(bytes32[] calldata dataFeedsIdsToUpdate) private view {
```
hatskier, line 55: I would remove this function

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -22,8 +21,12 @@ contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
     uint256 receivedTimestampMilliseconds
   );
 
-  function initialize(address priceFeedsRegistryAddress) public initializer {
-    priceFeedRegistry = PriceFeedsRegistry(priceFeedsRegistryAddress);
+  error InvalidNumberOfDataFeedsToUpdate(uint256 dataFeedsIds, uint256 dataFeedsIdsToUpdate);
+
+  error InvalidDataFeedsIdsToUpdate(bytes32[] dataFeedsIdsToUpdate);
```
hatskier, line 26: Since we use custom errors here, can we also use them in PriceFeed.sol instead of `revert("Use latestRoundData to get data feed price");`

hatskier (changes requested):

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -49,6 +52,10 @@ contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
     }
   }
 
+  function addDataFeedId(bytes32 newDataFeedId) public onlyOwner {
```
hatskier, line 55: Let's add a check for duplicates here

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -75,12 +86,37 @@ contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
...
     for (uint256 i = 0; i < dataFeedsIds.length; i++) {
-      PriceFeed(priceFeedRegistry.getPriceFeedContractAddress(dataFeedsIds[i])).storeDataFeedValue(
-          values[i]
-        );
+      dataFeedsValues[dataFeedsIds[i]] = values[i];
+    }
+  }
+
+  function getValueForDataFeed(bytes32 dataFeedId) public view returns (uint256) {
+    return dataFeedsValues[dataFeedId];
```
hatskier, line 96: Let's add a requirement that the value can't be 0

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -75,12 +86,37 @@ contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
...
+  function getValueForDataFeedAndLastRoundParams(bytes32 dataFeedId)
+    public
+    view
+    returns (
+      uint256,
+      uint256,
+      uint256
+    )
+  {
+    return (dataFeedsValues[dataFeedId], lastRound, lastUpdateTimestampMilliseconds);
```
hatskier, line 108: Let's add a requirement that the value can't be 0

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -75,12 +86,37 @@ contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
...
+  }
+
+  function getValuesForDataFeeds(bytes32[] memory dataFeedsIds_)
+    public
+    view
+    returns (bytes32[] memory, uint256[] memory)
+  {
+    uint256[] memory values = new uint256[](dataFeedsIds_.length);
+    for (uint256 i = 0; i < dataFeedsIds.length; i++) {
+      values[i] = dataFeedsValues[dataFeedsIds[i]];
```
hatskier, line 118: Let's add a requirement that the value can't be 0

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -75,12 +86,37 @@ contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
...
+    return (dataFeedsValues[dataFeedId], lastRound, lastUpdateTimestampMilliseconds);
+  }
+
+  function getValuesForDataFeeds(bytes32[] memory dataFeedsIds_)
+    public
+    view
+    returns (bytes32[] memory, uint256[] memory)
+  {
+    uint256[] memory values = new uint256[](dataFeedsIds_.length);
+    for (uint256 i = 0; i < dataFeedsIds.length; i++) {
```
hatskier, line 117: Looks like there is a bug in this function ;)

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -75,12 +86,37 @@ contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
...
+      uint256
+    )
+  {
+    return (dataFeedsValues[dataFeedId], lastRound, lastUpdateTimestampMilliseconds);
+  }
+
+  function getValuesForDataFeeds(bytes32[] memory dataFeedsIds_)
+    public
+    view
+    returns (bytes32[] memory, uint256[] memory)
```
hatskier, line 114: Why do we return `bytes32[] memory`?

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
```diff
@@ -75,12 +86,37 @@ contract PriceFeedsManager is MainDemoConsumerBase, Initializable {
...
+
+  function getValueForDataFeed(bytes32 dataFeedId) public view returns (uint256) {
+    return dataFeedsValues[dataFeedId];
+  }
+
+  function getValueForDataFeedAndLastRoundParams(bytes32 dataFeedId)
+    public
+    view
+    returns (
+      uint256,
```
hatskier, line 103: Let's add names to these returned values please

packages/evm-connector/contracts/price-feeds/PriceFeedsManager.sol
hatskier: I have another concern about this function.
Because right after its execution the contract can end up in a weird state. E.g.:
- We deployed the contract with the following feeds `[bytes32("BTC"), bytes32("ETH")]`
- We called the `updateDataFeedValues` function few times
- Then we call `addDataFeedId(bytes32("AVAX"))`
- And then any user can call `getValueForDataFeedAndLastRoundParams(bytes32("AVAX"))` and receive a response with 0 value and valid `lastUpdateTimestampMilliseconds` and `lastRound`

hatskier: But if we require non-zero values everywhere this issue will not occur

packages/evm-connector/contracts/price-feeds/PriceFeed.sol
```diff
@@ -52,7 +49,7 @@ contract PriceFeed is AggregatorV3Interface {
       uint80 answeredInRound
     )
   {
-    revert("Use latestRoundData to get data feed price");
+    revert UseLatestRoundToGetDataFeedPrice();
```
jakub-wojciechowski, line 52: Hmmm, to be compatible with the interface we can check if the roundId == latest Round, and revert only if they differ

hatskier: IMHO, it would create unclear behaviour. If this function can not be implemented properly, partial implementation may be misleading for devs. E.g. someone can rely on it, because it worked with the latest roundId, but then it stops working for any other value

packages/evm-connector/contracts/price-feeds/DefaultsLib.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+library DefaultsLib {
```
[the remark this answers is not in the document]
cehali: Changed name to `CustomErrors`

PR #418, opened 2023-03-07
feat: overwrite populate transaction function

hatskier:


packages/evm-connector/src/wrappers/BaseWrapper.ts
```diff
@@ -7,62 +7,100 @@ export interface ParamsForDryRunVerification {
...
-            const result = await contract[
-              shouldUseSigner ? "signer" : "provider"
-            ].call(tx);
+  private overwriteFunction({
+    wrappedContract,
+    contract,
+    functionName,
+  }: OverwriteFunctionArgs) {
+    const isCall = contract.interface.getFunction(functionName).constant;
+    const isDryRun = functionName.endsWith("DryRun");
```
cehali, line 77: Do we still use it somewhere or is it some `ether.js` native functionality?

cehali (approved):
Minor changes proposals in comments

packages/evm-connector/src/wrappers/BaseWrapper.ts
hatskier: It is required for forcing static calls of functions that need to use storage (temporarily) but are created for fetching data (e.g. cross-contracts calls with storage)

PR #420, opened 2023-03-08
refactor: multi request fetcher

hatskier:
I've changed the logic of MultiRequestFetcher, so it can has custom requestIds (in previous version only data feed ids were used as request ids). I also heavily refactored UniswapV2LikeFetcher, by removing unnecessary operations on BigNumbers. And added few other tweaks here and there.


packages/oracle-node/src/fetchers/MultiRequestFetcher.ts
```diff
@@ -1,45 +1,89 @@
...
+        success: false,
+        response: undefined,
+      };
+    }
   }
 
-  async fetchData(ids: string[]): Promise<any> {
-    const context = await this.getRequestContext(ids);
+  override fetchData(dataFeedIds: string[]): Promise<ExtendedPromiseResult[]> {
     const promises: Promise<any>[] = [];
```
cehali, line 47: What do you think about adding generic? I think it worked pretty well in `UniswapV2LikeFetcher`

hatskier: I would like to not use generic here as sometimes we will have different return types for different requests

packages/oracle-node/src/fetchers/MultiRequestFetcher.ts
```diff
@@ -1,45 +1,89 @@
...
+      if (promiseResult.success) {
+        validResponses[promiseResult.requestId] = promiseResult.response;
       }
+    }
 
-      result = this.processData(response.value, result, context);
+    // Extracting price values for each symbol
+    for (const dataFeedId of dataFeedIds) {
+      try {
+        const extractedPrice = this.extractPrice(dataFeedId, validResponses);
```
cehali, line 74: In all fetchers we will need to get a response per dataFeedId `validResponses[dataFeedId]`, maybe we should pass `validResponses[dataFeedId]` and not all responses. 

Also, I wondering why we cannot iterate through responses rather than `dataFeedsIds`. I know we have both solutions in fetchers but I cannot come up with an idea of what could be better.

hatskier: We need to iterate over data feed ids, because sometimes (e.g. uniswap-v2-like fetcher) we have fewer responses than data feed ids. Also, sometimes data feed id needs to have access to some other response (example in Verto fetcher with ar price result)

PR #421, opened 2023-03-08
fix: price manager tests

cehali:


packages/evm-connector/test/price-feeds/helpers.ts
```diff
@@ -12,7 +13,7 @@ interface DataPoint {
 
 export const dataFeedsIds = [ethDataFeed, btcDataFeed];
 
-export const getWrappedContract = (
+export const getWrappedContract = async (
```
hatskier, line 16: Can we change the name of this function so it indicates that we also reset block.time here?

hatskier (changes requested):

cehali:
It will be included in https://github.com/redstone-finance/redstone-oracles-monorepo/pull/422

PR #442, opened 2023-03-14
feat: add adapter for mento oracles

hatskier:


packages/on-chain-relayer/.env.mento-alfajores
```diff
@@ -0,0 +1,14 @@
+RELAYER_ITERATION_INTERVAL=10000
```
cehali, line 1: We already have folder with configs of deployed relayers: https://github.com/redstone-finance/redstone-oracles-monorepo/tree/main/packages/on-chain-relayer/deployed-config

packages/on-chain-relayer/test/relayer/update-prices.test.ts
```diff
@@ -10,44 +10,98 @@ import {
...
   });
 
-  beforeEach(async () => {
-    const MangerContractFactory = await ethers.getContractFactory(
-      "PriceFeedsAdapterMock"
-    );
-    managerContract = await MangerContractFactory.deploy(dataFeedsIds);
-    await managerContract.deployed();
-  });
+  beforeEach(async () => {});
```
cehali, line 27: remove unnecessary

packages/on-chain-relayer/contracts/mocks/MentoAdapterMock.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.4;
+
+import "@redstone-finance/evm-connector/contracts/mocks/AuthorisedMockSignersBase.sol";
+import "../custom-integrations/mento/MentoAdapter.sol";
+
+contract MentoAdapterMock is MentoAdapter, AuthorisedMockSignersBase {
```
cehali, line 8: I wondering if we need this mock. If we want to mock only function maybe we could go with https://ethereum-waffle.readthedocs.io/en/latest/mock-contract.html?highlight=mock

packages/on-chain-relayer/src/core/contract-interactions/update-prices.ts
```diff
@@ -1,38 +1,99 @@
...
-) => {
+): Promise<void> => {
   const dataPackagesTimestamps = Object.values(dataPackages).flatMap(
     (dataPackages) =>
       dataPackages.map(
         (dataPackage) => dataPackage.dataPackage.timestampMilliseconds
       )
   );
   const minimalTimestamp = Math.min(...dataPackagesTimestamps);
+  const wrappedAdapterContract =
```
cehali, line 35: Why did you move the wrapping before the condition?

packages/on-chain-relayer/contracts/core/PermissionlessPriceUpdater.sol
```diff
@@ -0,0 +1,115 @@
...
+    if (proposedTimestamp <= getLastUpdateTimestamp()) {
+      revert ProposedTimestampMustBeNewerThanLastTimestamp(
+        proposedTimestamp,
+        getLastUpdateTimestamp()
+      );
+    }
+  }
+
+  function isProposedRoundValid(uint256 proposedRound) private view returns (bool) {
+    return proposedRound == getLastRound() + 1;
```
jakub-wojciechowski, line 80: Hmm, I see some benefits by iterating the round, but does the additional cost justify it. 
The efficiency of relayer is really crucial. 
Shouldn't the timestamps be enough to avoid race conditions (updating with the same timestamp)?

packages/on-chain-relayer/contracts/core/PermissionlessPriceUpdater.sol
```diff
@@ -0,0 +1,115 @@
...
+    uint256 proposedRound,
+    uint256 proposedTimestamp
+  ) internal {
+    validateProposedRound(proposedRound);
+    validateProposedTimestamp(proposedTimestamp);
+    setLastRound(proposedRound);
+    setLastUpdateTimestamp(proposedTimestamp);
+  }
+
+  function validateDataPackageTimestampAgainstProposedTimestamp(
```
jakub-wojciechowski, line 39: Do we allow receivedTimestampMilliseconds == lastUpdateTimestampMilliseconds?

packages/on-chain-relayer/contracts/custom-integrations/mento/MentoAdapter.sol
```diff
@@ -0,0 +1,117 @@
...
+   * @param locationsInSortedLinkedLists The array of locations in linked list for reported values
+   */
+  function updatePriceValues(
+    uint256 proposedRound,
+    uint256 proposedTimestamp,
+    LocationInSortedLinkedList[] calldata locationsInSortedLinkedLists
+  ) public {
+    validateAndUpdateProposedRoundAndTimestamp(proposedRound, proposedTimestamp);
+
+    uint256 dataFeedsCount = getDataFeedsCount();
```
jakub-wojciechowski, line 104: Do we really need to access storage and get tdataFeeds and count?
(It's expensive) + there might be inconsistency between dataFeeds from contract and the locationsInSortedLinkedList. Maybe it'll be cheaper/easier to pass both from relayer?

packages/on-chain-relayer/contracts/core/PermissionlessPriceUpdater.sol
hatskier: Yes

packages/on-chain-relayer/contracts/core/PermissionlessPriceUpdater.sol
hatskier: We will remove rounds in a separate PR

packages/on-chain-relayer/contracts/mocks/MentoAdapterMock.sol
hatskier: I did it in a consistent way with the price feed adapter ([PriceFeedsAdapterMock.sol](https://github.com/redstone-finance/redstone-oracles-monorepo/blob/8a84b787f82781740b1b4e51cabc7c32cadbcaab/packages/on-chain-relayer/contracts/mocks/PriceFeedsAdapterMock.sol)). It's also consistent with approach that we use for testing smart contracts in evm-connector. I've added a new issue to consider switching to waffle mock instead of mock contract files: https://github.com/redstone-finance/redstone-oracles-monorepo/issues/482

packages/on-chain-relayer/.env.mento-alfajores
hatskier: Removed

packages/on-chain-relayer/test/relayer/update-prices.test.ts
hatskier: Removed

packages/on-chain-relayer/src/core/contract-interactions/update-prices.ts
hatskier: Moved after the condition

cehali (approved):
LGTM

PR #443, opened 2023-03-14
feat: add value deviation check to on-chain relayer

cehali:


packages/on-chain-relayer/src/core/contract-interactions/get-values-for-data-feeds.ts
```diff
@@ -0,0 +1,13 @@
+import { BigNumber, Contract, utils } from "ethers";
+
+export const getValuesForDataFeeds = async (
+  priceFeedsAdapterContract: Contract,
+  dataFeeds: string[]
```
hatskier, line 5: Can we return number[] instead of string[]?

packages/on-chain-relayer/src/core/update-conditions/should-update.ts
```diff
@@ -1,11 +1,45 @@
...
+    shouldUpdatePrices,
+    warningMessage: JSON.stringify(warningMessages),
+  };
+};
 
-export const shouldUpdate = (lastUpdateTimestamp: number) => {
-  const { updateCondition } = config;
-  switch (updateCondition) {
+const checkConditionByName = (conditionName: string, context: Context) => {
+  switch (conditionName) {
```
hatskier, line 35: Just an idea: what do you think about using an object instead of the `switch .. case`, e.g. `{"time": timeUpdateCondition,  "value-deviation": valueDeviationCondition}`?

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
```diff
@@ -0,0 +1,41 @@
...
+import { config } from "../../config";
+
+export const valueDeviationCondition = (
+  dataPackages: DataPackagesResponse,
+  pricesFromContract: number[]
+) => {
+  const dataFeedsIds = Object.keys(dataPackages);
+  const { minDeviationPercentage } = config;
+  const shouldUpdatePrices = dataFeedsIds.every((dataFeedId, index) => {
+    const minimalValue = getMinValueFromDataPackages(dataPackages, dataFeedId);
```
hatskier, line 12: I don't think we should get the min value, as it's possible that with the max value deviation can be higher. I would do like this:
```js
let maxDevitation = 0;
for (const dataFeedId of dataFeedIds) {
  for (const dataPackage of dataPackages[dataFeedId]) {
    const curDevitaion = calculateDeviation(dataPackage.value, pricesFromContract[dataFeedId].value);
    maxDeviation = Math.max(curDevitaion, maxDevitation);
  }
}
if (maxDevitation >= minDeviationPercentageToTriggerUpdate) {
  // ...
}
```

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
```diff
@@ -0,0 +1,41 @@
+import { INumericDataPoint } from "redstone-protocol";
+import { DataPackagesResponse } from "redstone-sdk";
+import { config } from "../../config";
+
+export const valueDeviationCondition = (
+  dataPackages: DataPackagesResponse,
+  pricesFromContract: number[]
```
hatskier, line 7: I would propose to return `pricesFromContract` as an object {[dataFeedId: string]: number}, as it minifies the risk of bugs related to different orders of data feeds in `dataPackages` and in pricesFromContract

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
```diff
@@ -0,0 +1,41 @@
...
+    const minimalValue = getMinValueFromDataPackages(dataPackages, dataFeedId);
+    const priceFromContract = pricesFromContract[index];
+    const pricesDiff = Math.abs(priceFromContract - minimalValue);
+    const percentageDiff = (pricesDiff * 100) / priceFromContract;
+    return percentageDiff >= minDeviationPercentage;
+  });
+
+  return {
+    shouldUpdatePrices,
+    warningMessage: shouldUpdatePrices
```
hatskier, line 21: Why do we pass the message? Can't we just log it right away here?

hatskier (changes requested):

packages/on-chain-relayer/src/core/contract-interactions/get-values-for-data-feeds.ts
hatskier: I would even return an object with mapping from dataFeedId to value

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
cehali: It could be misleading. When we use two different checks we want to log a warning when both conditions are not met. If we log right away we would have a situation where you have a warning and later a message about successfully updated prices.

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
cehali: I am not sure if we should get the max deviation. I can imagine a situation where someone wants to change the value in the contract when the price changes by 10%. We get data packages from multiple cache services and get different values. Max deviation passes the check but later in the contract during on-chain aggregation, we get a value which is only 8-9% different than the previous one.

hatskier: It's much smaller priblem if we update more often than needed, but it would be abig problem if we don't update when the price actually changed

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
hatskier: Ok, makes sense

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
hatskier: We can also calculate the median of data packages off-chain, but it will cause more moving parts and potential bugs (I can elaborate on this). I would still vote for the max deviation

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
```diff
@@ -0,0 +1,46 @@
...
+      : "Value has not deviated enough to be updated",
+  };
+};
+
+const calculateDeviation = (
+  valueFromFetchedDataPackage: number,
+  valueFromContract: number
+) => {
+  const pricesDiff = Math.abs(valueFromContract - valueFromFetchedDataPackage);
+  return (pricesDiff * 100) / valueFromContract;
```
hatskier, line 45: I would change `(pricesDiff * 100) / valueFromContract` to `(pricesDiff * 100) / Math.min(valueFromContract, valueFromFetchedDataPackage )`

hatskier (approved):
LGTM!

PR #446, opened 2023-03-14
chore: added twelve-data to manifests

rucinpk:


packages/oracle-node/src/fetchers/twelve-data/TwelveDataFetcher.ts
```diff
@@ -17,6 +17,9 @@ export class TwelveDataFetcher extends BaseFetcher {
   }
 
   override convertSymbolToId(symbol: string): string {
+    if (symbol.includes("USD=X")) {
```
hatskier, line 20: Let's use mapping instead (in the same way as we do in coingecko fetcher)

hatskier (changes requested):

packages/oracle-node/src/fetchers/twelve-data/twelve-data-symbol-to-id.json
```diff
@@ -0,0 +1,12 @@
+{
```
hatskier, line 1: IMO it would be cleaner to have real ids here, not the ones that should be merged with `/USD` later.
I mean smth like `{"EUR": "EUR/USD", "BRLUSD=X": "BRL/USD", ...}`

hatskier (changes requested):

hatskier (approved):

PR #453, opened 2023-03-16
feat: update coingecko fetcher to pro version

cehali:


packages/oracle-node/src/types.ts
```diff
@@ -161,6 +161,8 @@ export interface NodeConfig {
   stlouisfedApiKey?: string;
   minDataFeedsPercentageForBigPackage: number;
   providerIdForPriceBroadcasting?: string;
+  coingeckoApiUrl: string;
```
hatskier, line 164: Just curious, why coingeckoApiUrl is required but coingeckoApiKey can be undefined?

PR #456, opened 2023-03-16
feat: add manifest validator

cehali:


packages/oracle-node/src/manifest/ManifestHelper.ts
```diff
@@ -42,19 +43,15 @@ export default class ManifestHelper {
     return result;
   }
 
-  static getTimeoutForSource(
-    source: string,
-    manifest: Manifest
-  ): number | null {
+  static validateManifest(manifest: Manifest) {
```
hatskier, line 46: Looks a bit weird, why do we need to do it?

packages/oracle-node/src/manifest/validate-manifest.ts
```diff
@@ -0,0 +1,89 @@
+import { Manifest } from "../types";
+import ManifestConfigError from "./ManifestConfigError";
+
+interface FieldToValidate {
+  field: keyof Manifest;
+  validator: (manifest: Manifest, field?: keyof Manifest) => boolean;
+}
+
+const numberValidator = (manifest: Manifest, field?: keyof Manifest) => {
+  if (!field) {
```
hatskier, line 10: Why do we have this condition?

packages/oracle-node/src/manifest/validate-manifest.ts
```diff
@@ -0,0 +1,89 @@
+import { Manifest } from "../types";
```
hatskier, line 1: What do you think about using ready solutions for runtime type validations? e.g. https://dev.to/busypeoples/notes-on-advanced-typescript-runtime-validations-5b

packages/oracle-node/src/manifest/ManifestHelper.ts
cehali: I like the idea that everything regarding manifest we have in one `ManifestHelper` but functionality with validation manifest should be in a separate file

packages/oracle-node/src/manifest/validate-manifest.ts
cehali: Let's try zod: https://github.com/colinhacks/zod

packages/oracle-node/src/manifest/ManifestHelper.ts
```diff
@@ -46,19 +47,15 @@ export default class ManifestHelper {
...
+    validateManifest(manifest);
+  }
+
+  static getTimeoutForSource(source: string, manifest: Manifest) {
     if (!source.length) {
-      throw "Source for timeout not defined";
+      throw new Error("Source for timeout not defined");
```
hatskier, line 56: Hmm, maybe timeout for source, not `Source for timeout`?

packages/oracle-node/src/manifest/validate-manifest.ts
```diff
@@ -0,0 +1,60 @@
...
+});
+
+export const validateManifest = (manifest: Manifest) => {
+  const validationResult = ManifestSchema.safeParse(manifest);
+  if (!validationResult.success) {
+    const errors = validationResult.error.errors;
+    const errorsPaths = errors.map(
+      ({ path, message }) => `${path.join("/")}: ${message}`
+    );
+    const uniqueErrorsPaths = new Set(errorsPaths);
```
hatskier, line 54: Why do we need to create a set here? When the duplication is possible here?

packages/oracle-node/test/manifest/validate-manifest.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      { ...correctManifest, tokens: [] },
+    ] as Manifest[];
+    manifests.forEach((manifest, index) =>
+      expect(() => validateManifest(manifest)).toThrowError(
+        `Invalid manifest configuration - tokens: ${tokensErrors[index]}`
+      )
+    );
+  });
+
+  test("should not throw error if correct manifest", () => {
```
hatskier, line 79: Maybe `if manifest is correct` or `in case of correct manifest`

packages/oracle-node/src/manifest/ManifestHelper.ts
hatskier: Ok

hatskier (dismissed):
Approved with few minor comments

hatskier (commented):
One more idea, maybe we should get rid of the manifest type since we can use smth like `type Manifest = z.infer<typeof ManifestSchema>;`

packages/oracle-node/src/manifest/validate-manifest.ts
cehali: Good catch, it could be leftover when restrictions for manifest values were less strict

hatskier (approved):

PR #460, opened 2023-03-19
test: implement integration tests in monorepo

hatskier:


.github/workflows/monorepo-integration-tests.yml
```diff
@@ -0,0 +1,32 @@
...
+    branches: "**"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: packages/integration
+    steps:
+      - uses: actions/checkout@v2
```
cehali, line 14: You should use `actions/checkout@v3` (https://github.com/actions/checkout) in order to use node 16 and do not receive warnings

packages/evm-connector/benchmarks/local-mock-test.ts
```diff
@@ -0,0 +1,50 @@
+import { expect } from "chai";
+import { BigNumber } from "ethers";
+import { formatBytes32String } from "ethers/lib/utils";
+import { ethers } from "hardhat";
+import { WrapperBuilder } from "../src/index";
+import { Benchmark } from "../typechain-types";
+
+// This test is used in monorepo intergration tests
+describe("Local mock test", function () {
+  let contract: Benchmark;
```
cehali, line 10: Maybe a separate contract?

packages/integration/integration-test.sh
```diff
@@ -0,0 +1,110 @@
+set -e # to exit when any command fails
```
cehali, line 1: When I was thinking about how to write an integration test for our infrastructure I thought about multiple dockers instances. But maybe bash script is enough 😄. I am only wondering if this could be reused somehow f.g. in my task in which I have to to write a test which will check fetching and aggregation of crucial tokens.

packages/sdk/src/index.ts
```diff
@@ -28,7 +29,10 @@ export interface DataPackagesResponse {
 
 export const getOracleRegistryState =
   async (): Promise<RedstoneOraclesState> => {
-    return redstoneOraclesInitialState;
+    // Mock data service is used for monorepo integration tests
+    const shouldUseMockState =
+      process && process.env.USE_MOCK_ORACLE_STATE === "true";
+    return shouldUseMockState ? mockOraclesState : redstoneOraclesInitialState;
```
cehali, line 35: It is a little bit tricky because `sdk` itself doesn't have a `.env` file. Maybe we could override the [check](https://github.com/redstone-finance/redstone-oracles-monorepo/blob/main/packages/cache-service/src/data-packages/data-packages.controller.ts#L97) in cache-service depending on this env variable.

packages/integration/integration-test.sh
hatskier: Bash script is way faster. I don't think it should be reused in your dry-run test as the dry-run for crucial tokens can be implemented in the oracle-node context only

packages/oracle-node/manifests/dev/one-iteration-mock.json
```diff
@@ -0,0 +1,18 @@
+{
+  "interval": 15000,
+  "useCustomScheduler": "one-iteration",
```
jakub-wojciechowski, line 3: One-iteration is already implemented but have you considered using timeout (https://linuxize.com/post/timeout-command-in-linux/), because: 
1) We can have flexibility on how long to we want to keep node for (maybe test a few iterations)
2) Test another scheduler (arbitrum)
3) Don't produce too much code that is only used in testing env / test as close as possible to prod env (one-iteration scheduler)

hatskier: 1. The only thing I don't like about this approach is that tests then become non-deterministic, because we don't know exactly how much time it takes to prepare a node (compile code, connect to local db etc.) before running the first iteration (locally it's usually few seconds, but it can significantly vary depending on computing resources)
2. I agree that for testing another scheduler we need to thinks about something like this, but I would prefer to do it in a separate PR
3. I believe that one-iteration scheduler can actually be used sometimes (e.g. if we decide to deploy oracle-node with lambda + custom scheduler, or as ECS scheduled task, these ways have their advantages (e.g. lack of dependence between iteration))

cehali: On the one hand, the `one-iteration` solution is really clean and easy to understand. But on the other hand, I agree with Kuba that this test should be as close to a production solution as possible. We could miss something without using `node-scheduler` or after development, the production code will start to differ even more. I suggested waiting for some logs from the node the same way we wait for files or URLs. But we could also run for some time during which we are sure that few iterations are done.

packages/integration/integration-test-core.sh
```diff
@@ -0,0 +1,132 @@
...
+
+  # Launching one iteration of oracle-node
+  cd ../oracle-node
+  installAndBuild
+  cp .env.example .env
+  updateDotEnvFile "OVERRIDE_DIRECT_CACHE_SERVICE_URLS" '["http://localhost:3000"]'
+  updateDotEnvFile "OVERRIDE_MANIFEST_USING_FILE" "./manifests/single-source/mock.json"
+  updateDotEnvFile "ECDSA_PRIVATE_KEY" $HARDHAT_MOCK_PRIVATE_KEY
+  runWithLogPrefix "yarn start:prod" "oracle-node" &
+  oracleNodePid=$!
```
cehali, line 105: I am not sure if `oracleNodePid` is the one you are trying to get there. This would be the id of the yarn command, not node one.

packages/cache-service/scripts/wait-for-data-packages.ts
```diff
@@ -0,0 +1,38 @@
+import { DataPackage } from "../src/data-packages/data-packages.model";
+import { ALL_FEEDS_KEY } from "../src/data-packages/data-packages.service";
+import mongoose from "mongoose";
+import config from "../src/config";
+
+// USAGE: yarn run-ts scripts/wait-for-data-packages.ts <expected-count> <data-feed-id>
+
+// Note! This script is used only in monorepo integration tests
+
+const INTERVAL_MILLISECONDS = 5000; // 5 seconds
```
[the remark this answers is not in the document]
hatskier: It won't be 15 seconds

cehali (approved):
LGTM

PR #467, opened 2023-03-20
feat: add ping uptime-kuma for on-chain relayer

cehali:


packages/on-chain-relayer/src/core/monitoring/ping-uptime-kuma.ts
```diff
@@ -0,0 +1,7 @@
+import axios from "axios";
+import { config } from "../../config";
+
+export const pingUptimeKuma = async () => {
```
hatskier, line 4: pingUptimeKuma -> sendHealthcheckPing

hatskier (changes requested):

packages/on-chain-relayer/src/core/monitoring/ping-uptime-kuma.ts
```diff
@@ -0,0 +1,7 @@
...
+  const uptimeKumaUrl = config.uptimeKumaUrl;
```
hatskier, line 5: Let's check if the variable is not empty here. If empty we can skip the healthcheck. It will simplify to run the relayer locally for development and test

hatskier (approved):
LGTM!

PR #469, opened 2023-03-21
feat: added contracts sdk

Lukasz2891:


packages/sdk/src/contracts/prices/PricesContractAdapter.ts
```diff
@@ -0,0 +1,31 @@
+import { ContractDataProvider } from "../ContractDataProvider";
+import { PricesCommandFactory } from "./PricesCommandFactory";
+import { ContractAdapter } from "../ContractAdapter";
+
+export class PricesContractAdapter implements ContractAdapter {
+  constructor(private commandFactory: PricesCommandFactory) {}
+
+  async getPrices(dataProvider: ContractDataProvider): Promise<number[]> {
```
hatskier, line 8: Let's think about removing that layer as it doesn't have any custom logic

packages/sdk/src/contracts/ContractConnector.ts
```diff
@@ -0,0 +1,6 @@
+import { ContractAdapter } from "./ContractAdapter";
+
+export interface ContractConnector<A extends ContractAdapter> {
+  getAdapter(): Promise<A>;
+  getBlockNumber(rpcUrl: string): Promise<number>;
```
hatskier, line 5: Maybe we can remove the rpcUrl

packages/starknet-connector/src/StarknetContractConnector.ts
```diff
@@ -0,0 +1,44 @@
...
+      ["ACCEPTED_ON_L2", "REJECTED"]
+    );
+  }
+
+  getContract(): Contract {
+    return new Contract(this.getAbi(), this.contractAddress, this.provider);
+  }
+
+  async getBlockNumber(rpcUrl: string): Promise<number> {
+    const response = await axios.post(rpcUrl, {
```
hatskier, line 35: Let's check if you can use this.provider to get block number

packages/sdk/src/contracts/prices/PricesCommandFactory.ts
```diff
@@ -0,0 +1,18 @@
+import { ContractDataProvider } from "../ContractDataProvider";
+import { ContractCommand } from "../ContractCommand";
+
+export interface PricesCommandFactory {
+  makeGetPricesCommand(
+    dataProvider: ContractDataProvider
+  ): ContractCommand<number[]>;
+
+  makeWritePricesCommand(
+    dataProvider: ContractDataProvider
```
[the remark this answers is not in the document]
Lukasz2891: Enhancement :)

packages/sdk/src/contracts/ContractConnector.ts
Lukasz2891: The provider doesn't provide the rpcUrl :(

packages/starknet-connector/src/StarknetContractConnector.ts
Lukasz2891: It's possible to check data of the block passed as the argument, but there is not possible to check the last block data.

packages/sdk/src/contracts/ContractParamsProvider.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  async getPayloadData(): Promise<number[]> {
+    this.getDataFeedIds();
+
+    const payloadHex = await this.requestPayload(this.requestParams);
+
+    return Array.from(arrayify(payloadHex));
+  }
+
+  getFeedIds(): BigNumberish[] {
```
hatskier, line 26: I would change the name to smth like `getHexlifiedFeedIds` or `getFeedIdsBytes`

hatskier (approved):

PR #476, opened 2023-03-22
fix: isolate extract price for diffrent symbols

### LINKED ISSUE #474: [oracle-node] Fix isolation of processing price in fetchers
- [x] check list for creating new `Fetchers`

koonopek:
In some cases it might seem that nothing can fail in the try/catch block, however, I think it is better to leave it there now, then count that when someone will be updated given fetcher (and introduce change that can fail), he will remember to add try/catch block.
Closes #474

### DIFF
--- a/packages/oracle-node/src/fetchers/BaseFetcher.ts
+++ b/packages/oracle-node/src/fetchers/BaseFetcher.ts
@@ -1,5 +1,6 @@
 import { Consola } from "consola";
 import { Fetcher, PriceDataFetched, FetcherOpts, PricesObj } from "../types";
+import { stringifyError } from "../utils/error-stringifier";
 import createLogger from "../utils/logger";
 
 const MAX_RESPONSE_TIME_TO_RETRY_FETCHING_MS = 3000;
@@ -94,5 +95,35 @@ export abstract class BaseFetcher implements Fetcher {
       }
     }
     return prices;
+  }
+
+  protected extractPricesSafely<T>(
+    items: T[],
+    extractPriceFn: (item: T, pricesObj: PricesObj) => number | undefined,
+    extractIdFn: (item: T) => string
+  ): PricesObj {
+    const pricesObj: PricesObj = {};
+    for (const item of items) {
+      let id;
+      try {
+        id = extractIdFn(item);
+        if (!id) {
+          throw Error("Could not extract id from response");
+        }
+        const price = extractPriceFn(item, pricesObj);
+
+        if (price) {
+          pricesObj[id] = price;
+        } else {
+          throw Error("Could not extract price from response");
+        }
+      } catch (e: unknown) {
+        this.logger.error(
+          `Extracting price failed for id: ${id} error: ${stringifyError(e)}`
+        );
+      }
+    }
+
+    return pricesObj;
   }
 }
--- a/packages/oracle-node/src/fetchers/BaseFetcher.ts
+++ b/packages/oracle-node/src/fetchers/BaseFetcher.ts
@@ -2,8 +2,10 @@ import { Consola } from "consola";
 import { Fetcher, PriceDataFetched, FetcherOpts, PricesObj } from "../types";
 import { stringifyError } from "../utils/error-stringifier";
 import createLogger from "../utils/logger";
+import { isDefined } from "../utils/objects";
 
 const MAX_RESPONSE_TIME_TO_RETRY_FETCHING_MS = 3000;
+export type ExtractPricePairFn<T> = (item: T) => { id: string; value: number };
 
 export abstract class BaseFetcher implements Fetcher {
   protected name: string;
@@ -97,26 +99,43 @@ export abstract class BaseFetcher implements Fetcher {
     return prices;
   }
 
+  /**
+   * Function used to map items to pricesObj. Errors between extracts are isolated.
+   * @param items list of objects that price and id will be extracted from
+   * @param extractPricePairFn this function should extract price pair from item.
+   *  If any of return values will be undefined or function wil throw, price pair
+   * will not be added to result.
+   * @returns pricesObj
+   */
   protected extractPricesSafely<T>(
     items: T[],
-    extractPriceFn: (item: T, pricesObj: PricesObj) => number | undefined,
-    extractIdFn: (item: T) => string
+    extractPricePairFn: ExtractPricePairFn<T>
   ): PricesObj {
     const pricesObj: PricesObj = {};
     for (const item of items) {
       let id;
       try {
-        id = extractIdFn(item);
-        if (!id) {
-          throw Error("Could not extract id from response");
-        }
-        const price = extractPriceFn(item, pricesObj);
+        const pricePair = extractPricePairFn(item);
 
-        if (price) {
-          pricesObj[id] = price;
-        } else {
-          throw Error("Could not extract price from response");
+        if (!isDefined(pricePair)) {
+          throw new Error("Could not extract price pair from response");
         }
+
+        if (!isDefined(pricePair.id)) {
+          throw new Error("Could not extract id from response");
+        }
+        id = pricePair.id;
+
+        if (!isDefined(pricePair.value)) {
+          throw new Error("Could not extract price from response");
+        }
+
+        if (isDefined(pricesObj[pricePair.id])) {
+          // case were same id is extracted more than once
+          continue;
+        }
+
+        pricesObj[pricePair.id] = pricePair.value as number;
       } catch (e: unknown) {
         this.logger.error(
           `Extracting price failed for id: ${id} error: ${stringifyError(e)}`
--- a/packages/oracle-node/src/fetchers/DexFetcher.ts
+++ b/packages/oracle-node/src/fetchers/DexFetcher.ts
@@ -1,7 +1,6 @@
-import graphProxy from "../utils/graph-proxy";
 import { PricesObj } from "../types";
+import graphProxy from "../utils/graph-proxy";
 import { BaseFetcher } from "./BaseFetcher";
-import { stringifyError } from "../utils/error-stringifier";
 
 export interface DexFetcherResponse {
   data: {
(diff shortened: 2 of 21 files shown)


packages/oracle-node/src/fetchers/DexFetcher.ts
```diff
@@ -72,25 +73,31 @@ export class DexFetcher extends BaseFetcher {
...
-        const reserve0 = parseFloat(pair.reserve0);
-        const reserve1 = parseFloat(pair.reserve1);
-        const reserveUSD = parseFloat(pair.reserveUSD);
-
-        if (symbol0 === currentAssetId) {
-          pricesObj[currentAssetId] = reserveUSD / (2 * reserve0);
-        } else if (symbol1 === currentAssetId) {
-          pricesObj[currentAssetId] = reserveUSD / (2 * reserve1);
+      try {
+        const pairId = this.symbolToPairIdObj[currentAssetId];
```
hatskier, line 77: What do you think about moving the code inside the `try` block into a separate function?

hatskier (commented):
Refactoring idea to discuss: we can create a function (similar to forEach) e.g. `extractSafelyForEach` to reduce amount of duplicated code (catch blocks with error logging)

packages/oracle-node/src/fetchers/DexFetcher.ts
koonopek: Good idea

koonopek:
> Refactoring idea to discuss: we can create a function (similar to forEach) e.g. extractSafelyForEach to reduce amount of duplicated code (catch blocks with error logging). This function can work exactly as forEach, but also can wrap each iteration into try .. catch block

Yes, we can do it, however, if we want to log error with a proper message, we will also have to pass a function that will extract symbol from "thing" that we are iterating because the "thing" is not always pure id it is sometimes object.  But let's try it!

@hatskier

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -95,4 +96,24 @@ export abstract class BaseFetcher implements Fetcher {
     }
     return prices;
   }
+
+  protected extractSafelyForEach<T>(
```
koonopek, line 100: Here is implementation

packages/oracle-node/src/fetchers/DexFetcher.ts
```diff
@@ -69,32 +71,38 @@ export class DexFetcher extends BaseFetcher {
   }
 
   extractPrices(response: DexFetcherResponse, assetIds: string[]): PricesObj {
-    const pricesObj: PricesObj = {};
-
-    for (const currentAssetId of assetIds) {
-      const pairId = this.symbolToPairIdObj[currentAssetId];
-      const pair = response.data.pairs.find((pair) => pair.id === pairId);
+    return this.extractSafelyForEach(
```
koonopek, line 74: Example usage,  I haven't implemented it for all fetchers yet, WDYT?

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -95,4 +96,24 @@ export abstract class BaseFetcher implements Fetcher {
...
+    items: T[],
+    setPriceFn: (item: T, pricesObj: PricesObj) => void,
```
hatskier, line 102: Does this function really need the `pricesObj: PricesObj` argument?

koonopek: Yeah that was my first implementation however it doesn't cover all cases, for example CCXTFetcher
```js
    let pairSymbol = ticker.symbol;
    pairSymbol = this.serializePairSymbol(pairSymbol);
    const lastPrice = ticker.last as number;
    if (pairSymbol.endsWith("/USD")) {
      pricesObj[pairSymbol] = lastPrice;
    } else {
      if (!pricesObj[pairSymbol]) {
        const lastUsdInStablePrice = this.getStableCoinPrice(pairSymbol);
        if (lastUsdInStablePrice) {
          pricesObj[pairSymbol] = lastPrice * lastUsdInStablePrice;
        }
      }
    }
```
Pass function sometimes need access to PricesObject

koonopek: However, I think we can merge these two idea, and still pass pricesObj for cases like the one above where it is needed, and still assign to pricesObj in `extractSinglePriceFn`

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -95,4 +96,34 @@ export abstract class BaseFetcher implements Fetcher {
...
+    extractId: (item: T) => string
+  ): PricesObj {
+    const pricesObj: PricesObj = {};
+    for (const item of items) {
+      let id: string = "unknown";
+      try {
+        id = extractId(item);
+        if (!id) {
+          throw Error("Could not extract id from response");
+        }
```
cehali, line 112: Maybe not give the `id` initial value because it could lead to the situation when the `if (!id)` check will return false for the `unknown` value when it should throw an error.

koonopek:
Ohh great catche

Pozdrawiam,
Michał Konopka

pon., 27 mar 2023, 12:17 użytkownik Cezary Haliniarz <
***@***.***> napisał:

packages/oracle-node/src/fetchers/BaseFetcher.ts
koonopek: Ohh great cache, fixed!

packages/oracle-node/src/fetchers/ccxt/CcxtFetcher.ts
```diff
@@ -51,24 +52,27 @@ export class CcxtFetcher extends BaseFetcher {
   }
 
   extractPrices(response: any): PricesObj {
-    const pricesObj: PricesObj = {};
+    return this.extractPricesSafely(
+      Object.values(response) as Ticker[],
+      (ticker, pricesObj) => this.extractPrice(ticker, pricesObj),
```
hatskier, line 57: Can we just pass `this.extractPrice.bind(this)` instead?

packages/oracle-node/src/fetchers/chainlink/ChainlinkFetcher.ts
```diff
@@ -15,14 +15,16 @@ export class ChainlinkFetcher extends BaseFetcher {
   }
 
   extractPrices(response: any): PricesObj {
-    const pricesObj: PricesObj = {};
+    return this.extractPricesSafely(
+      Object.keys(response),
```
hatskier, line 19: Shouldn't we pass ids here?

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -95,4 +96,34 @@ export abstract class BaseFetcher implements Fetcher {
     }
     return prices;
   }
+
+  protected extractPricesSafely<T>(
+    items: T[],
+    extractPriceFn: (item: T, pricesObj: PricesObj) => number | undefined,
+    extractId: (item: T) => string
```
hatskier, line 103: Let's call it `extractIdFn` to be consistent with `extractPriceFn`. Also, what do you think about passing here a default implementation (`(id) => id`)?

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -95,4 +96,34 @@ export abstract class BaseFetcher implements Fetcher {
...
+  ): PricesObj {
+    const pricesObj: PricesObj = {};
+    for (const item of items) {
+      let id;
+      try {
+        id = extractId(item);
+        if (!id) {
+          throw Error("Could not extract id from response");
+        }
+        const price = extractPriceFn(item, pricesObj);
```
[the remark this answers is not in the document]
koonopek: Yes, I answered it here: https://github.com/redstone-finance/redstone-oracles-monorepo/pull/476#discussion_r1147176135, maybe if we could try to change the underlying implementation  of cctx fetcher, however i don't got what is the cons of passing pricesObj

packages/oracle-node/src/fetchers/ccxt/CcxtFetcher.ts
koonopek: yep

packages/oracle-node/src/fetchers/chainlink/ChainlinkFetcher.ts
koonopek: Don't get it, seems okey for me, same as with previous implemnetation

packages/oracle-node/src/fetchers/BaseFetcher.ts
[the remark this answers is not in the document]
hatskier: The only con I can see is that it's unclear what is happening with pricesObj after passing it to the function. AFAIU it is used only to read data, but technically it can be modified by the `extractPriceFn` function and it will be different to debug it

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -95,4 +96,34 @@ export abstract class BaseFetcher implements Fetcher {
...
+    for (const item of items) {
+      let id;
+      try {
+        id = extractIdFn(item);
+        if (!id) {
+          throw Error("Could not extract id from response");
+        }
+        const price = extractPriceFn(item, pricesObj);
+
+        if (price) {
```
hatskier, line 115: What about 0 values? Shouldn't we also save them to pricesObj?

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -95,4 +96,34 @@ export abstract class BaseFetcher implements Fetcher {
...
+    extractPriceFn: (item: T, pricesObj: PricesObj) => number | undefined,
+    extractIdFn: (item: T) => string
+  ): PricesObj {
+    const pricesObj: PricesObj = {};
+    for (const item of items) {
+      let id;
+      try {
+        id = extractIdFn(item);
+        if (!id) {
+          throw Error("Could not extract id from response");
```
hatskier, line 111: Shouldn't we add the `new` keyword? (`throw new Error(...`)

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -95,4 +96,34 @@ export abstract class BaseFetcher implements Fetcher {
...
+        }
+        const price = extractPriceFn(item, pricesObj);
+
+        if (price) {
+          pricesObj[id] = price;
+        } else {
+          throw Error("Could not extract price from response");
```
hatskier, line 118: Shouldn't we add the `new` keyword? (`throw new Error(...`)

packages/oracle-node/src/fetchers/BaseFetcher.ts
koonopek: Uhh perfect catch, however it is copied code, so I will change underlying business logic of extracting prices

packages/oracle-node/src/fetchers/BaseFetcher.ts
koonopek: @hatskier WDYT?

packages/oracle-node/src/fetchers/BaseFetcher.ts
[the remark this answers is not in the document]
koonopek: Same as `forEach` the third parameter :) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#parameters However I get the argument, it is okay for me to go with such an implementation.

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -95,4 +100,51 @@ export abstract class BaseFetcher implements Fetcher {
...
+   * Function used to map items to pricesObj. Errors between extracts are isolated.
+   * @param items list of objects that price and id will be extracted from
+   * @param extractPricePairFn this function should extract price pair from item.
+   *  If any of return values will be undefined or function wil throw, price pair
+   * will not be added to result.
+   * @returns pricesObj
+   */
+  protected extractPricesSafely<T>(
+    items: T[],
+    extractPricePairFn: ExtractPricePairFn<T>
```
koonopek, line 114: Change implementation to return {id, price} obj, thus no additional function is needed and the code is easier to read.

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -1,8 +1,13 @@
 import { Consola } from "consola";
 import { Fetcher, PriceDataFetched, FetcherOpts, PricesObj } from "../types";
+import { stringifyError } from "../utils/error-stringifier";
 import createLogger from "../utils/logger";
+import { isDefined } from "../utils/objects";
 
 const MAX_RESPONSE_TIME_TO_RETRY_FETCHING_MS = 3000;
+export type ExtractPricePairFn<T> = (
+  item: T
```
koonopek, line 9: Remove passing pricesObj

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -1,8 +1,13 @@
...
+) => { id?: string; price?: number } | undefined;
```
hatskier, line 10: Let's remove `?s` from id and price and `| undefined`, but still have validation (in runtime) in BaseFetcher

hatskier (approved):
Great job!

PR #478, opened 2023-03-23
feat: make saves to bundlr and db independent

### LINKED ISSUE #477: CacheService - if we fail to save data to Mongo we should not fail to save data to Bundlr (and vice versa)

koonopek:
Close #477


packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -36,6 +37,48 @@ export interface StatsRequestParams {
 
 @Injectable()
 export class DataPackagesService {
+  private readonly logger = new Logger(DataPackagesService.name);
+
+  constructor(private readonly bundlrService: BundlrService) {}
+
+  /**  Save dataPackages to DB and bundlr if enabled */
+  async saveMany(
```
koonopek, line 45: This class sound like something that should abstract storage layer, so that is why I added this function. It is used twice by bulk endpoint and streamer listener.

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -36,6 +37,48 @@ export interface StatsRequestParams {
...
+    dataPackagesToSave: CachedDataPackage[],
+    nodeEvmAddress: string
+  ): Promise<void> {
+    const savePromises: Promise<any>[] = [];
+    const saveToDbPromise = this.saveManyDataPackagesInDB(dataPackagesToSave)
+      .catch(() =>
```
koonopek, line 51: Quite verbose code, however I didn't see added value in extracting it to new function to many questions started to come up

koonopek: However I'm open to changes

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -36,6 +37,48 @@ export interface StatsRequestParams {
...
+        this.logger.log(
+          `Save ${dataPackagesToSave.length} data packages for node ${nodeEvmAddress} to Database`
+        )
+      );
+    savePromises.push(saveToDbPromise);
+
+    if (config.enableArchivingOnArweave) {
+      const saveToBundlrPromise = this.bundlrService
+        .saveDataPackages(dataPackagesToSave)
+        .catch(() =>
```
hatskier, line 66: Let's log error details here `.catch((err) => ...`

packages/cache-service/src/data-packages/data-packages.service.ts
hatskier, line 51: Let's log error details here please `.catch((err) => ...`

hatskier: What do you think about creating a helpful function to wrap promises with log messages, e.g. smth like this
```js
// pseudo-code
const addLogs = (promise: Promise<void>, promiseName: string): Promise<void> => {
   return promise.then(() => console.log(`${promiseName} completed}).catch(err => console.err(`${promiseName} failed. ${stringifyErr(err)}`);
}
```

hatskier (changes requested):

packages/cache-service/src/data-packages/data-packages.service.ts
koonopek: I agree good idea, however, I think that addLogs function should rethrow the error after logging it, and handling of the error should happen where business logic happens. Also, the name doesn't signalize that function will consume error.

packages/cache-service/src/utils/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+  } else if (e.response) {
+    return JSON.stringify(e.response.data) + " | " + e.stack;
+  } else if (e.toJSON) {
+    return JSON.stringify(e.toJSON());
+  } else {
+    return e.stack || String(e);
+  }
+}
+
+export function withLogging<T>(
```
koonopek, line 16: Implementation after refactor

hatskier: Can we call it `runPromiseWithLogging`?

hatskier (approved):

PR #495, opened 2023-03-30
[oracle-node] make redstone Numbers great again

koonopek:
* Current implementation is using Js Native numbers under the hood, to minimize the "change step" and because they are the fastest solution
* If we want to implement decimals, we should pass numbers in JSON as a string to avoid losing precision! (this requires changes across multiple components)
* This PR doesn't incorporate "RedstoneNumber" into fetchers logic 
* Some flaws of this solution are pointed as the comments


packages/oracle-node/.jest/set-redstone-number-config.js
```diff
@@ -0,0 +1,14 @@
+import {
+    setJsNativeRedstoneNumberConfig,
+} from "../src/numbers/JsNativeRedstoneNumber";
+
+const throwErr = msg => { throw new Error(msg) };
+
+setJsNativeRedstoneNumberConfig({
+    ON_NUMBER_VALIDATION_ERROR: {
+        1: throwErr,
```
koonopek, line 9: Js files, so Typescript Enum doesn't work here :(

packages/oracle-node/jest.config.ts
```diff
@@ -7,7 +7,10 @@ const config: Config.InitialOptions = {
     "^.+\\.(ts|js)$": "ts-jest",
   },
   testEnvironment: "node",
-  setupFiles: ["<rootDir>/.jest/set-env-vars.js"],
+  setupFiles: [
+    "<rootDir>/.jest/set-env-vars.js",
+    "<rootDir>/.jest/set-redstone-number-config.js",
```
koonopek, line 12: Set to throw error on every invalid number

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -12,7 +11,7 @@ const DEFAULT_LEVEL_OPTS = {
 
 export interface PriceValueInLocalDB {
   timestamp: number;
-  value: number;
+  value: string;
```
koonopek, line 14: I saw that this will be stored as "text" in sqlite anyway

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -115,12 +114,26 @@ const lastPrices: LastPrices = {};
...
-export const getLastPrice = (symbol: string): PriceValueInLocalDB | undefined =>
-  lastPrices[symbol];
+type PriceValueInLocalDBAsNumber =
+  | {
+      timestamp: number;
+      value: number;
+    }
+  | undefined;
+
+export const getLastPrice = (
```
koonopek, line 128: Own review: change it to always return string (no lost precision) let fetchers decide how he converts string to number or maybe some decimal

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
```diff
@@ -0,0 +1,237 @@
...
+    } else {
+      throw new Error(
+        `Invalid number format: Tried to create JsNativeRedstoneNumber from ${numberLike}`
+      );
+    }
+  }
+
+  private _value: number;
+
+  private constructor(value: number) {
```
koonopek, line 72: private thus, we have to use from

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
```diff
@@ -0,0 +1,237 @@
...
+  gte(numberArg: NumberArg): boolean {
+    return (
+      this._value >=
+      JsNativeRedstoneNumber.from(numberArg)._value -
+        JsNativeRedstoneNumberConfig.EPSILON
+    );
+  }
+
+  toJSON(): number {
+    return this.unsafeToNumber();
```
koonopek, line 158: In case of decimals we would loose here precision. However in this solution it is okey and have to be like this

packages/oracle-node/src/numbers/RedstoneNumber.ts
```diff
@@ -0,0 +1,21 @@
+import { NumberLike } from "redstone-protocol/src/common/utils";
+
+export interface RedstoneNumber {
```
koonopek, line 3: Any suggestion to naming earlier was `PreciseNumber` however it is not so precise

packages/oracle-node/src/numbers/RedstoneNumberFactory.ts
```diff
@@ -0,0 +1,6 @@
+import { JsNativeRedstoneNumber } from "./JsNativeRedstoneNumber";
+import { NumberArg } from "./RedstoneNumber";
+
+/** Factory for RedstoneNumber */
+export const N = (numberLike: NumberArg) =>
```
koonopek, line 5: If we wan't to change impl just change JsNativeRedstoneNumber to anything implementing RedstonNumber

packages/oracle-node/src/types.ts
```diff
@@ -52,10 +54,13 @@ export interface Fetcher {
   ) => Promise<PriceDataFetched[]>;
 }
 
+export type SanitizedPriceDataBeforeAggregation =
```
koonopek, line 57: Sanitized values are using RedstonNumber

packages/oracle-node/test/numbers/JsNativeRedstoneNumber.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+        }
+      }
+    );
+  });
+
+  describe("arithmetic operations", () => {
+    describe("add", () => {
+      it("should add floats (tricky 0.3)", () => {
+        const result = JsNativeRedstoneNumber.from("100.131").add("100.123");
+        expect(result.toString()).toBe("200.25400000000002"); // with this implementation we we
```
koonopek, line 88: Still things like this will happen, only using  decimals can prevent this

packages/oracle-node/.jest/set-redstone-number-config.js
cehali: But why it needs to be .js file?

packages/oracle-node/src/aggregators/lwap-aggregator/get-liquidities.ts
```diff
@@ -1,16 +1,20 @@
 import { buildLiquidityDataFeedId } from "../../fetchers/liquidity/utils";
 import { PricesWithLiquidity } from "./lwap-aggregator";
-import { PriceDataBeforeAggregation, Source } from "../../types";
+import { PriceDataBeforeAggregation, PriceSource } from "../../types";
+import { RedstoneNumber } from "../../numbers/RedstoneNumber";
+import { N } from "../../numbers/RedstoneNumberFactory";
```
cehali, line 5: I am not a big fan of such shortcuts, it could be hard to understand what is going on without the context. Maybe `SafeNumber`?

packages/oracle-node/src/db/local-db/index.ts
cehali: Not yet decided 😄

packages/oracle-node/src/db/local-db/index.ts
cehali: Values here are the same as already broadcasted. If the precision was lost it is already lost. I think we should have here number type for values as default

packages/oracle-node/src/numbers/RedstoneNumber.ts
cehali: `SafeNumber`?

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
```diff
@@ -0,0 +1,237 @@
...
+  return mergedConfig;
+};
+
+export class JsNativeRedstoneNumber implements RedstoneNumber {
+  /** This method should only be called by factory {N} */
+  static from(numberLike: NumberArg): JsNativeRedstoneNumber {
+    if (numberLike instanceof JsNativeRedstoneNumber) {
+      return new JsNativeRedstoneNumber(
+        parseToSafeNumber(numberLike.toString())
+      );
```
cehali, line 57: Do we need this?

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
cehali: Could you elaborate more?

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
```diff
@@ -0,0 +1,237 @@
...
+  /** In the case of this implementation it is actually safe. */
+  unsafeToNumber(): number {
+    return this._value;
+  }
+
+  eq(numberArg: NumberArg): boolean {
+    const number = JsNativeRedstoneNumber.from(numberArg);
+
+    return (
+      Math.abs(number.unsafeToNumber() - this.unsafeToNumber()) <
```
cehali, line 128: Why we don't use the `sub` function?

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
```diff
@@ -0,0 +1,237 @@
...
+      JsNativeRedstoneNumber.from(numberArg)._value -
+        JsNativeRedstoneNumberConfig.EPSILON
+    );
+  }
+
+  toJSON(): number {
+    return this.unsafeToNumber();
+  }
+
+  private produceNewNumber(result: number) {
```
cehali, line 161: I feel like `produceNewNumber` and `from` could be merged into one function as it similar functionality

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  gte(numberArg: NumberArg): boolean {
+    return (
+      this._value >=
+      JsNativeRedstoneNumber.from(numberArg)._value -
+        JsNativeRedstoneNumberConfig.EPSILON
+    );
+  }
+
+  toJSON(): number {
```
cehali, line 157: Do we need this function?

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
```diff
@@ -0,0 +1,237 @@
...
+    return this.produceNewNumber(result);
+  }
+
+  assertNonNegative(): JsNativeRedstoneNumber {
+    assert(this._value >= 0, `${this.toString} >= 0`);
+    return this;
+  }
+
+  /** In the case of this implementation it is actually safe. */
+  unsafeToNumber(): number {
```
cehali, line 120: We can assume that the value is always valid as it is checked in the constructor. Maybe remove unsafe from the name?

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
```diff
@@ -0,0 +1,237 @@
+import { Consola } from "consola";
+import { assert, NumberLike } from "redstone-protocol/src/common/utils";
+import { NumberArg, RedstoneNumber } from "./RedstoneNumber";
+
+const logger = require("../utils/logger")(
+  "numbers/JsNativeRedstoneNumber"
+) as Consola;
+
+enum NumberValidationResult {
+  isOk = 0,
```
cehali, line 10: `= 0` is not necessary?

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
```diff
@@ -0,0 +1,237 @@
...
+    Math.abs(number) < JsNativeRedstoneNumberConfig.MIN_NUMBER &&
+    number !== 0
+  ) {
+    return [
+      NumberValidationResult.isUnderflow,
+      `Invalid number format: Number is smaller then min number acceptable by REDSTONE ${number}`,
+    ];
+  }
+
+  return [NumberValidationResult.isOk, ""];
```
cehali, line 211: Maybe return object {result, message}

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
cehali: I think we can use only the `from` function. In all cases, we want to run `parseToSafeNumber` which the `JsNativeRedstoneNumber` constructor doesn't do. Also in `produceNewNumber` we ran `assertValidAndRound` twice, once in the constructor and then in the function itself.

packages/oracle-node/test/numbers/JsNativeRedstoneNumber.spec.ts
cehali: The only solution is to multiply by 10 to the power of the number of decimals in order to operate on integers or is there any other solution to address this issue?

packages/oracle-node/.jest/set-redstone-number-config.js
koonopek: Heh, good point (i thought that it has to be cause set-env-vars.js). But it works when writtern in TS.

packages/oracle-node/src/db/local-db/index.ts
koonopek: Exactly!
That is why I postulated that we should broadcast prices as string, string can't lose precision :)
In case of DB it is easy fix and it is future proof solution (if we decide to use decimals implementation)

packages/oracle-node/src/numbers/RedstoneNumber.ts
koonopek: So no change?

packages/oracle-node/test/numbers/JsNativeRedstoneNumber.spec.ts
koonopek: Yes that is one of the solution then we can use BigInt to represent an integer. Another solution is to use lib like decimal.js, however then we should also: `If we want to implement decimals, we should pass numbers in JSON as a string to avoid losing precision! (this requires changes across multiple components)`. There is also pretty big cost in performance, both in memory (20x) and cpu (200x)

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
koonopek: Yes, cause we want to clone. Thus

```js
const a = new JsNativeRedstoneNumber(5);
const b = new JsNativeRedstoneNumber(a);
a===b => false
```

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
koonopek: In case of this implementation that is totally true, however in other implementations that won't be true, and this name was introduced from the perspective of interface - `RedstoneNumber` - not concrete implementation

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
koonopek: Yes to handle positive result

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
koonopek: Why not

packages/oracle-node/src/aggregators/lwap-aggregator/get-liquidities.ts
koonopek: Work for me, when we establish the name I will rename

packages/oracle-node/src/db/local-db/index.ts
hatskier: I agree, I also think that we should move to keeping strings in cache service DB instead of values. Let's do it as a separate task

packages/oracle-node/src/numbers/RedstoneNumber.ts
hatskier: Maybe SmartNumber?

hatskier: or SafeNumber

cehali: xD 
I wanted to write `RedstoneSafeNumber`

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
cehali: I know why `isOk` is there but `= 0` seems unnecessary 😄

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
cehali: I think I know why it was like that before. When adding two numbers the first one is already parsed and checked. For the second one, we are parsing and then through `assertValidAndRound` in the constructor we were checking if the second value is correct.

packages/oracle-node/src/numbers/JsNativeRedstoneNumber.ts
koonopek: Valid point, thanks we can use a private  constructor here

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -212,13 +209,18 @@ export default class PricesService {
...
         );
 
-        // Throwing an error if price is invalid, too deviated or out of hard limits
-        this.assertValidPrice(
-          priceAfterAggregation,
-          pricesInLocalDBForSymbol,
-          deviationCheckConfig,
-          pricesLimits
+        // Throwing an error if price < 0 is invalid or too deviated
+        priceAfterAggregation.value.assertNonNegative();
```
hatskier, line 213: Can you add a task to make this assertion optional (by setting a pram in manifest)?

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -232,92 +234,66 @@ export default class PricesService {
...
     for (const [sourceName, valueFromSource] of Object.entries(price.source)) {
-      const valueFromSourceNum = safelyConvertAnyValueToNumber(valueFromSource);
-      const { isValid, reason } = this.validatePrice({
-        value: valueFromSourceNum,
-        timestamp: price.timestamp,
-        deviationConfig: deviationCheckConfig,
-        recentPrices: recentPricesInLocalDBForSymbol,
-      });
+      try {
+        const valueFromSourceNum = SafeNumber(valueFromSource);
```
hatskier, line 250: Shouldn't we add a word `new` each time we create a new instance of `SafeNumber`?

packages/oracle-node/src/numbers/SafeNumberFactory.ts
```diff
@@ -0,0 +1,6 @@
+import { JsNativeSafeNumber } from "./JsNativeSafeNumber";
+import { NumberArg } from "./ISafeNumber";
+
+/** Factory for SafeNumber */
+export const SafeNumber = (numberLike: NumberArg) =>
```
hatskier, line 5: IMHO, If this function has a capital first letter it's easy to confuse it with a class. I would propose to rename it to smth like `createSafeNumber`

hatskier (changes requested):

packages/oracle-node/src/fetchers/PricesService.ts
koonopek: Nope this is fabric method, I tried to follow JS convention as with `Number => Number(10)` or `BigInt(1000) for BigInt`

packages/oracle-node/src/numbers/SafeNumberFactory.ts
koonopek: I tried to follow JS convention as with Number => Number(10) or BigInt(1000) for BigInt.
However, I can rename if you don't like it @cehali  what is your opinion

packages/oracle-node/src/fetchers/PricesService.ts
koonopek: Done

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -117,13 +117,31 @@ const lastPrices: LastPrices = {};
...
 export const getLastPrice = (
   symbol: string
-): PriceValueInLocalDB | undefined => {
+): PriceValueInLocalDBAsNumber | undefined => {
+  const price = getRawPrice(symbol);
+  if (!price) {
+    return undefined;
+  }
+  return {
+    value: Number(lastPrices[symbol].value),
```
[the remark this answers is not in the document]
koonopek: Of course, great catch

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,243 @@
+import { Consola } from "consola";
+import { utils } from "redstone-protocol";
+import { NumberArg, ISafeNumber } from "./ISafeNumber";
+
+const logger = require("../utils/logger")(
+  "numbers/JsNativeSafeNumber"
+) as Consola;
+
+export enum NumberValidationResult {
+  isOk = 0,
```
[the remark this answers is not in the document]
koonopek: We don't have to

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,243 @@
...
+  isOverflow,
+  isUnderflow,
+}
+
+export type NumberValidationError = Exclude<
+  NumberValidationResult,
+  NumberValidationResult.isOk
+>;
+
+export let JsNativeSafeNumberConfig = {
```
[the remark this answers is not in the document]
koonopek: It can be modified in `setjsNatvieSafeNumberConfig`

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,243 @@
...
+    return {
+      result: NumberValidationResult.isUnderflow,
+      message: `Invalid number format: Number is smaller then min number acceptable by REDSTONE ${number}`,
+    };
+  }
+
+  return { result: NumberValidationResult.isOk, message: "" };
+};
+
+const parseToSafeNumber = (value: utils.NumberLike) => {
```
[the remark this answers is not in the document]
koonopek: I would leave it as is, just for the seek for readability - `from` function would grow at least twice

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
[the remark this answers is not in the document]
koonopek: Removed

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,243 @@
...
+
+  mul(numberLike: NumberArg): JsNativeSafeNumber {
+    const result = this._value * JsNativeSafeNumber.from(numberLike)._value;
+
+    return this.produceNewNumber(result);
+  }
+
+  assertNonNegative(): JsNativeSafeNumber {
+    utils.assert(this._value >= 0, `${this.toString} >= 0`);
+    return this;
```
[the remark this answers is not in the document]
koonopek: agree

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,243 @@
...
+      JsNativeSafeNumber.from(numberArg)._value -
+        JsNativeSafeNumberConfig.EPSILON
+    );
+  }
+
+  toJSON(): number {
+    return this.unsafeToNumber();
+  }
+
+  private produceNewNumber(result: number) {
```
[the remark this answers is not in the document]
koonopek: done

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,243 @@
...
+    return this._value.toString();
+  }
+
+  abs(): JsNativeSafeNumber {
+    const result = Math.abs(this._value);
+    return this.produceNewNumber(result);
+  }
+
+  add(numberLike: NumberArg): JsNativeSafeNumber {
+    const result = this._value + JsNativeSafeNumber.from(numberLike)._value;
```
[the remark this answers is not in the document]
koonopek: Why you think it's not a good pratice? It is always one function call less, and in my opinion it is more readable

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,243 @@
...
+    [NumberValidationResult.isNotFinite]: (msg) => {
+      throw new Error(msg);
+    },
+    [NumberValidationResult.isOverflow]: logger.error,
+    [NumberValidationResult.isUnderflow]: logger.error,
+  } as Record<NumberValidationError, (msg: string) => any>,
+  EPSILON: 1e-14,
+};
+
+export const setJsNativeSafeNumberConfig = (
```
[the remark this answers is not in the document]
koonopek: Good point, will do

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,243 @@
...
+    );
+  }
+
+  lt(numberArg: NumberArg): boolean {
+    return this._value < JsNativeSafeNumber.from(numberArg)._value;
+  }
+
+  lte(numberArg: NumberArg): boolean {
+    return (
+      this._value <=
```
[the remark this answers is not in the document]
koonopek: I like this `return this.lt(numberArg) || this.eq(numberArg)` (my first impl)

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,243 @@
...
+    );
+  }
+
+  gt(numberArg: NumberArg): boolean {
+    return this._value > JsNativeSafeNumber.from(numberArg)._value;
+  }
+
+  gte(numberArg: NumberArg): boolean {
+    return (
+      this._value >=
```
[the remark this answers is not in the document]
koonopek: same as in lite :)

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,243 @@
...
+  return mergedConfig;
+};
+
+export class JsNativeSafeNumber implements ISafeNumber {
+  /** This method should only be called by factory {N} */
+  static from(numberLike: NumberArg): JsNativeSafeNumber {
+    if (numberLike instanceof JsNativeSafeNumber) {
+      return new JsNativeSafeNumber(numberLike._value);
+    } else if (
+      typeof numberLike === "number" ||
```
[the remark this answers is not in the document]
koonopek: I don't have an opinion on style, but if use `["number", "string"].includes(typeof numberLike)` then typescript compiler doesn't recognize that function indeed has to be string or number and we get type error

packages/oracle-node/src/numbers/SafeNumberFactory.ts
koonopek: Done, because @cehali  is on vacation :)

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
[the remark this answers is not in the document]
hatskier: Gotcha

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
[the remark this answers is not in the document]
hatskier: I think it's OK for the objects, on which the mthod is called, but you are referring to it in another (external) object here `JsNativeSafeNumber.from(numberLike)._value`

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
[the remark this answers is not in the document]
hatskier: Ok

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,229 @@
...
+    [NumberValidationResult.isUnderflow]: logger.error,
+  } as Record<NumberValidationError, (msg: string) => any>,
+  EPSILON: 1e-14,
+};
+
+export class JsNativeSafeNumber implements ISafeNumber {
+  /** This method should only be called by factory {N} */
+  static from(numberLike: NumberArg): JsNativeSafeNumber {
+    if (numberLike instanceof JsNativeSafeNumber) {
+      return new JsNativeSafeNumber(numberLike._value);
```
hatskier, line 44: IMO it's not very nice to use a private property of an external object

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,229 @@
...
+    );
+  }
+
+  lt(numberArg: NumberArg): boolean {
+    return this._value < JsNativeSafeNumber.from(numberArg)._value;
+  }
+
+  lte(numberArg: NumberArg): boolean {
+    return (
+      this._value < JsNativeSafeNumber.from(numberArg)._value ||
```
hatskier, line 123: Why not lt?

packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -0,0 +1,229 @@
...
+    );
+  }
+
+  gt(numberArg: NumberArg): boolean {
+    return this._value > JsNativeSafeNumber.from(numberArg)._value;
+  }
+
+  gte(numberArg: NumberArg): boolean {
+    return (
+      this._value > JsNativeSafeNumber.from(numberArg)._value ||
```
hatskier, line 134: Why not gt?

packages/oracle-node/src/utils/numbers.ts
```diff
@@ -13,24 +15,30 @@ export const safelyConvertAnyValueToNumber = (value: any): number => {
...
-  measuredValue: number;
-  trueValue: number;
+  measuredValue: ISafeNumber;
+  trueValue: ISafeNumber;
 }) => {
   const { measuredValue, trueValue } = args;
 
-  const secureTrueValue = trueValue === 0 ? Number.MIN_VALUE : trueValue;
+  if (trueValue.eq(0)) {
+    // TODO: make it more generic
```
hatskier, line 37: What do you mean here? Can you do it within this PR?

PR #496, opened 2023-03-30
feat: implement getrounddata support in redstone classic model

hatskier:


packages/on-chain-relayer/contracts/price-feeds/PriceFeed.sol
```diff
@@ -34,20 +36,28 @@ contract PriceFeed is AggregatorV3Interface {
...
-    revert UseLatestRoundToGetDataFeedPrice();
+    (uint256 dataFeedValue, uint256 roundTimestampInMilliseconds) = priceFeedsAdapter.getRoundData(
+      dataFeedId,
+      requestedRoundId
+    );
+    roundId = requestedRoundId;
+    answer = int256(dataFeedValue);
+    startedAt = roundTimestampInMilliseconds / 1000;
+    updatedAt = startedAt;
+    answeredInRound = requestedRoundId;
```
cehali, line 60: Is it ok from the perspective of the compatibility that `startedAt` and  `updatedAt` or `roundId` and `answeredInRound` are the same values?

packages/on-chain-relayer/contracts/price-feeds/PriceFeedsAdapter.sol
```diff
@@ -11,9 +11,17 @@ contract PriceFeedsAdapter is MainDemoConsumerBase, Ownable, PermissionlessPrice
...
+  struct RoundData {
+    uint256 value;
+    uint256 timestamp;
+  }
 
   EnumerableSet.Bytes32Set private dataFeedsIds;
-  mapping(bytes32 => uint256) dataFeedsValues;
+
+  mapping(bytes32 => mapping(uint256 => uint256)) public roundValues; // dataFeedId => (roundId => value)
+  mapping(uint256 => uint256) public roundTimestamps; // roundId => timestamp
```
cehali, line 24: Is it better to use two mappings or one mapping with struct above?

packages/on-chain-relayer/contracts/price-feeds/PriceFeed.sol
hatskier: answeredInRound is a deprrecated param, it should be equal to roundId
And we don't separate starting time and updating time, so IMO it's okay

packages/on-chain-relayer/contracts/price-feeds/PriceFeedsAdapter.sol
hatskier: Yes, it's better, because it uses much less memory thanks t not duplicating timestamps for each data feed

packages/on-chain-relayer/contracts/price-feeds/PriceFeedsAdapter.sol
```diff
@@ -11,9 +11,17 @@ contract PriceFeedsAdapter is MainDemoConsumerBase, Ownable, PermissionlessPrice
   using EnumerableSet for EnumerableSet.Bytes32Set;
 
   error DataFeedValueCannotBeZero(bytes32 dataFeedId);
+  error RoundNotFound(uint256 roundId);
+
+  struct RoundData {
+    uint256 value;
+    uint256 timestamp;
+  }
```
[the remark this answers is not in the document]
hatskier: Ups, yep, will remove 😄

cehali (approved):
LGTM

PR #499, opened 2023-03-30
feat: remove hardhat from relayer code

cehali:


koonopek (approved):
Overall looks good! 
I'm only wondering if we could tune typescript to work with hardhat, but you have probably tried :)

packages/on-chain-relayer/test/relayer/update-prices.test.ts
```diff
@@ -58,6 +59,8 @@ describe("#updatePrices", () => {
   });
 
   it("should update prices in mento adapter", async () => {
+    (getProviderOrSigner as any).getProvider = () => ethers.provider;
```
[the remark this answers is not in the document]
koonopek: I see I will have to add in my PR anyway

hatskier (commented):
Please wait before merging this

packages/on-chain-relayer/scripts/mento/deploy-mock-sorted-oracles.ts
```diff
@@ -0,0 +1,28 @@
+import { Signer } from "ethers";
+import { ethers } from "hardhat";
+
+export const deployMockSortedOracles = async (signer?: Signer) => {
```
hatskier, line 4: IMO this is not a script and should not be located in the scripts folder. Script should be executable, but this file just exports a function

packages/on-chain-relayer/src/custom-integrations/mento/mento-utils.ts
```diff
@@ -95,29 +94,3 @@ export const prepareLinkedListLocationsForMentoAdapterReport = async ({
 
   return locationsInSortedLinkedLists;
 };
-
-export const deployMockSortedOracles = async (signer?: Signer) => {
```
hatskier, line 99: IMO this is a good location for this function

hatskier (changes requested):

packages/on-chain-relayer/src/custom-integrations/mento/mento-utils.ts
cehali: But it is not used anywhere in files from the src folder. It is only used in one script and tests.

PR #506, opened 2023-04-03
feat: add fallback mechanism for relayer-rpc-node

koonopek:


packages/on-chain-relayer/src/core/contract-interactions/get-contract.ts
```diff
@@ -3,14 +3,23 @@ import { abi as priceFeedsAdapterABI } from "../../../artifacts/contracts/price-
...
-  const signer = new Wallet(privateKey, provider);
+  const { adapterContractAddress } = config;
   const abi = getAbiForAdapter();
-  return new Contract(adapterContractAddress, abi, signer);
+  const contract = new Contract(adapterContractAddress, abi);
+  const signers = config.rpcUrls.map((_rpcUrl: string, index: number) =>
+    getSigner(getProvider(index))
+  );
+
+  return contractDecorator.decorate(contract, signers);
```
koonopek, line 22: We are Wrapping (decorating) contract with "fallback logic"

packages/on-chain-relayer/src/core/contract-interactions/get-contract.ts
```diff
@@ -3,14 +3,23 @@ import { abi as priceFeedsAdapterABI } from "../../../artifacts/contracts/price-
...
 
 export const getAdapterContract = () => {
-  const { privateKey, adapterContractAddress } = config;
-  const provider = getProvider();
-  const signer = new Wallet(privateKey, provider);
+  const { adapterContractAddress } = config;
   const abi = getAbiForAdapter();
-  return new Contract(adapterContractAddress, abi, signer);
+  const contract = new Contract(adapterContractAddress, abi);
+  const signers = config.rpcUrls.map((_rpcUrl: string, index: number) =>
```
koonopek, line 18: Providers will be used in given sequencer if [0] fails then [1] etc.

packages/on-chain-relayer/src/fallback-contract.ts
```diff
@@ -0,0 +1,131 @@
+import { JsonRpcProvider, Provider } from "@ethersproject/providers";
+import { Contract, PopulatedTransaction, Signer } from "ethers";
+import { ErrorCode } from "@ethersproject/logger";
+import { config } from "./config";
+
+export class FallbackContractDecorator {
```
koonopek, line 6: Same idea as with RedstoneWrapper for contracts (most of the code is coppied).
I could have used Proxy, however I thought that there is no need for reinventing the wheel

koonopek:
Will deployed these nodes for testing purposes

packages/on-chain-relayer/deployed-config/arbitrum-goerli-3-nodes.json
```diff
@@ -0,0 +1,16 @@
+{
```
cehali, line 1: I think we can update `arbitrum-goerli.json`. We will not keep 4th relayer alive for a long time.

packages/on-chain-relayer/src/fallback-contract.ts
cehali, line 6: I generally like the idea. Also, we want something similar in oracle-nodes and I think it could be an excellent opportunity to extract this to the small package that can be reused in any project and published on npm.

cehali: I want to use it in oracle-node already 😄

packages/rpc-providers/README.md
```diff
@@ -0,0 +1,11 @@
+# Magic providers, compatible with ethers.Provider interface
+
+**Tested only with JsonRpcProviders**
+
+## ProviderWithFallback
```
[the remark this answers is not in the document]
koonopek: For sure need battle testing

packages/rpc-providers/test/ProviderWithFallback.test.ts
```diff
@@ -0,0 +1,132 @@
...
+import { ProviderWithFallback } from "../src/ProviderWithFallback";
+import { Counter } from "../typechain-types";
+import { ethers } from "hardhat";
+
+chai.use(chaiAsPromised);
+
+const TEST_PRIV_KEY =
+  "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
+
+describe("ProviderWithFallback", () => {
```
koonopek, line 15: Pretty hard to test because ethers.provider is proxy, thus it can't be mocked ..

.github/workflows/on-chain-relayer-tests.yml
```diff
@@ -14,6 +14,8 @@ jobs:
         working-directory: packages/on-chain-relayer
     steps:
       - uses: actions/checkout@v3
+      - name: Build dependent packages
+        run: cd ../rpc-providers && yarn && yarn build
```
koonopek, line 18: Is there a better way?

packages/on-chain-relayer/src/core/contract-interactions/get-provider-or-signer.ts
```diff
@@ -1,16 +1,28 @@
...
+    (rpcUrl: string) =>
+      new Providers.StaticJsonRpcProvider(rpcUrl, {
+        name: chainName,
+        chainId: Number(chainId),
+      })
+  );
+
+  if (providers.length === 0) {
+    throw new Error("Please provide at least one RPC_URL in RPC_URLS");
+  } else if (providers.length === 1) {
```
koonopek, line 18: Running fallback provider with one provider doesnt have sense

packages/rpc-providers/contracts/Counter.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.4;
+
+contract Counter {
```
koonopek, line 5: for testing purposes

koonopek:
@cehali @hatskier Okey so I have finally finished this big guy. I think I implemented it 3 times and after every implementation, I have changed my mind, I hope this version is okey.

.github/workflows/on-chain-relayer-tests.yml
```diff
@@ -14,6 +14,8 @@ jobs:
         working-directory: packages/on-chain-relayer
     steps:
       - uses: actions/checkout@v3
+      - name: Build dependent packages
+        run: cd ../rpc-providers && yarn && yarn build && cd ../on-chain-relayer
```
cehali, line 18: I think that can be handled by reference in tsconfig file, take a look at `tsconfig.json` in `on-chain-relayer`

packages/on-chain-relayer/package.json
```diff
@@ -15,7 +15,8 @@
     "@redstone-finance/evm-connector": "^0.0.19",
     "dotenv": "^16.0.3",
     "ethers": "^5.7.2",
-    "toad-scheduler": "^2.2.0"
+    "toad-scheduler": "^2.2.0",
+    "redstone-rpc-providers": "*"
```
cehali, line 19: This dependency also can be handled by reference in tsconfig file, we use `redstone-sdk` and `redstone-protocol` but we don't have it specified as a dependency.

packages/on-chain-relayer/package.json
```diff
@@ -34,6 +35,7 @@
     "hardhat": "^2.13.0",
     "hardhat-gas-reporter": "^1.0.8",
     "msw": "^1.1.0",
+    "sinon": "^15.0.3",
```
cehali, line 38: Change from other PR? 😄

packages/on-chain-relayer/src/core/contract-interactions/get-last-round-params.ts
```diff
@@ -1,11 +1,11 @@
...
+import { BigNumber, Contract } from "ethers";
 
 const parseBigNumberParam = (valueInBigNumber: BigNumber) =>
-  Number(utils.formatUnits(valueInBigNumber, 0));
+  valueInBigNumber.toNumber();
 
 export const getLastRoundParamsFromContract = async (
   managerContract: Contract
-) => {
+): Promise<{ lastRound: number; lastUpdateTimestamp: number }> => {
```
cehali, line 8: Please extract the type to separate the variable with the proper name

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,255 @@
...
+
+export type ProviderWithFallbackConfig = {
+  unrecoverableErrors: ErrorCode[];
+  logger: { info: (message: string) => void; warn: (message: string) => void };
+  providerNames?: string[];
+};
+
+export const FALLBACK_DEFAULT_CONFIG: ProviderWithFallbackConfig = {
+  unrecoverableErrors: [
+    ErrorCode.NONCE_EXPIRED,
```
cehali, line 25: I think we could retry nonce error, in most cases it is probably race condition problem

packages/rpc-providers/test/ProviderWithFallback.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    const ContractFactory = await hardhat.ethers.getContractFactory("Counter");
+    const contract = await ContractFactory.deploy();
+    await contract.deployed();
+
+    const alwaysFailingProvider = new providers.JsonRpcProvider(
+      "http://blabla.xd"
+    );
+    const fallbackProvider = new ProviderWithFallback([
+      alwaysFailingProvider,
+      // hardhat.ethers.provider,
```
cehali, line 114: Unnecessary comment?

.github/workflows/on-chain-relayer-tests.yml
koonopek: Tried but still doesn't work without pre-building deps

packages/on-chain-relayer/package.json
koonopek: Hmm, so what are the pros of using "reference"? I would stay with explicit adding is a dep, I think that is "yarn workspaces way of doing this"

packages/on-chain-relayer/package.json
koonopek: No actually using this in tests to mock Provider

packages/on-chain-relayer/src/core/contract-interactions/get-last-round-params.ts
koonopek: Done

packages/rpc-providers/src/ProviderWithFallback.ts
koonopek: That is the case for relayer node, where we actually expect nonce errors.
But that his valid point I will extract it from Default Config, and include in Relayer Code

packages/rpc-providers/test/ProviderWithFallback.test.ts
koonopek: yep

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,255 @@
...
+    const nextProviderName = this.extractProviderName(nextProviderIndex);
+
+    this.config.logger.info(
+      `Fallback in to next provider ${nextProviderName}.`
+    );
+
+    return nextProviderIndex;
+  }
+
+  sendTransaction(...args: any[]): Promise<TransactionResponse> {
```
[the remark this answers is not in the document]
koonopek: I tried, but if we try to do it dynamically we don't get TypeScript support. For example
```ts
interface Hi { hi(): void };
class A implements Hi {
 constructor() {
   this['hi'] = console.log('hi')
 }
}
```
It will fail to compile.

Thus, I have extracted the abstract class which repeats all boilerplate. Maybe it is more readable now WDYT?

.github/workflows/on-chain-relayer-tests.yml
cehali: But in the current implementation, the relayer depends on `redstone-sdk` and we don't need to build it in github action.

koonopek: Ohhh it was failing cause in package json `tsc -build` instead of `tsc --build`.
But why do we have @redstone-finance/evm-connector as dependency then and not reference?

koonopek: Also had to add tsconfig.build.json to exclude tests :)
Thanks @cehali for validating this!

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    if (retryNumber === this.providers.length - 1) {
+      this.config.logger.warn(
+        `All providers failed to execute action, rethrowing error`
+      );
+      throw error;
+    }
+
+    // if we haven't tried provider for this request
+    const nextProviderIndex =
```
Lukasz2891, line 132: Won't `%` (modulus) work here?

```js
nextProviderIndex = (this.providerIndex + 1) % this.providers.length
```

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    if (error?.code && this.config.unrecoverableErrors.includes(error.code)) {
+      this.config.logger.warn(
+        `Unrecoverable error ${error.code}, rethrowing error`
+      );
+      throw error;
+    }
+
+    this.config.logger.warn(
+      `Provider ${providerName} failed with error: ${error?.code} ${error.message}`
```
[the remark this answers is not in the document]
koonopek: Good catch!

packages/rpc-providers/src/ProviderWithFallback.ts
koonopek: Yep, cleaner!

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,203 @@
...
+    this.currentProvider.removeAllListeners(eventName);
+    this.updateGlobalListenerAfterRemoval();
+
+    return this.currentProvider;
+  }
+
+  override removeListener(eventName: EventType, listener: Listener): Provider {
+    this.currentProvider.removeListener(eventName, listener);
+    this.updateGlobalListenerAfterRemoval();
+    return this.currentProvider;
```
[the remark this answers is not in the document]
koonopek: I have added new lines by hand for functions with multiple lines + return. Changes to eslint should be done in different PR I think because It will be very hard to review this one (we have global eslint setup)

koonopek:
Benchmarks on read arbitrum 3 providers:
FALLBACK READ 0: 707.616ms
FALLBACK READ 1: 561.783ms
FALLBACK READ 2: 1.051s
FALLBACK READ 3: 568.889ms
FALLBACK READ 4: 552.365ms
FALLBACK READ 5: 584.257ms
FALLBACK READ 6: 594.999ms
FALLBACK READ 7: 650.666ms
FALLBACK READ 8: 597.99ms
FALLBACK READ 9: 591.665ms
SOLO  READ 0: 296.884ms
SOLO  READ 1: 152.708ms
SOLO  READ 2: 146.817ms
SOLO  READ 3: 147.758ms
SOLO  READ 4: 163.572ms
SOLO  READ 5: 158.537ms
SOLO  READ 6: 149.834ms
SOLO  READ 7: 148.664ms
SOLO  READ 8: 164.003ms
SOLO  READ 9: 182.735ms

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,192 @@
...
+function cmpCallResult(a: CallResult, b: CallResult): boolean {
+  if (a.error || b.error) {
+    return false;
+  }
+
+  return a.data === b.data;
+}
+
+/** Only difference from Promise.race is that it replace resolved promises if undefined */
+function smartPromiseRace<T>(promises: Promise<T>[]): {
```
koonopek, line 159: .

koonopek: This one makes us fast, because we don't have to wait for all promises to resolve

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,192 @@
...
+    const agreedResults: CallResult[] = [];
+    let resolvedCalls = 0;
+
+    while (true) {
+      const { theFastestPromise, notResolvedPromises } = await smartPromiseRace(
+        promises
+      );
+
+      const result = await theFastestPromise
+        .then((data) => ({
```
hatskier, line 95: In my opinion, it's not really readable to mix async/await syntax with promises chaining (using `then` and `catch` functions). But it's a matter of taste

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    return blockNumbers.sort((a, b) => a - b).at(1) as number;
+  }
+
+  private async resolveCallPromises(promises: Promise<any>[]) {
+    const agreedResults: CallResult[] = [];
+    let resolvedCalls = 0;
+
+    while (true) {
+      const { theFastestPromise, notResolvedPromises } = await smartPromiseRace(
```
hatskier, line 90: I think this algorithm can be improved, as currently it can send too many unnecessary requests to the RPC URLs. We can call all the promises one time and just handle responses on demand, and stop handling them when received enough valid and equal responses. We can save responses in a multiset, or a map like this: hash(response.tx.data) => response.tx.data

hatskier (changes requested):

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,219 @@
...
+  protected override async executeWithFallback(
+    retryNumber = 0,
+    fnName: string,
+    ...args: any[]
+  ): Promise<any> {
+    try {
+      return await (this.currentProvider as any)[fnName](...[...args]);
+    } catch (error: any) {
+      const newProviderIndex = this.electNewProviderOrFail(error, retryNumber);
+      this.useProvider(newProviderIndex);
```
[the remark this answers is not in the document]
koonopek: Soultion: isolate provider counting per call. after failure check if index has already changes if so the do nothing otherwise select new provider

packages/rpc-providers/src/ProviderWithAgreement.ts
koonopek: will do

packages/rpc-providers/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+1. If the provider fails on some operation new provider is elected.
+2. If all providers fail on the same operation error is thrown.
+3. When we switch to the next provider, all operations (excluding 1.) are executed by him till it fails, when it does ->
+4. Providers are elected by the sequence given in array. If array ends we start from the beginning.
+
+## ProviderWithAgreement
+
+1. Works like `ProviderWithFallback, with an exception for two methods.
+2. `getBlockNumber` asks all providers for blockNumber and then picks `blockNumber.sort(ASC)[1]`
```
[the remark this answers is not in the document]
koonopek: YES, actually this idea come from you I think, that we should take the second lowest block number as current state

packages/rpc-providers/README.md
```diff
@@ -0,0 +1,19 @@
...
+3. `call` ask all providers for result
+   - if at least `N` answers are the same return call result
+   - ignore all errors
+   - as soon as he will receive two matching responses return. (doesn't wait for the rest of providers to finish)
```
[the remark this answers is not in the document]
koonopek: Changed in whole README.md

packages/rpc-providers/src/ProviderWithFallback.ts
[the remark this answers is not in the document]
koonopek: done

packages/rpc-providers/test/ProviderWithAgreement.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    beforeEach(async () => {
+      providerWithAgreement = new ProviderWithAgreement([
+        falseProvider,
+        new providers.StaticJsonRpcProvider("http://blabla.xd"),
+        hardhat.ethers.provider,
+      ]);
+    });
+
+    it("should read from contract", async () => {
+      const counter = contract.connect(signer.connect(providerWithAgreement));
```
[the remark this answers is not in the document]
koonopek: done also in ProviderWithFallback

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,219 @@
...
+  Listener,
+  Network,
+  Provider,
+  TransactionReceipt,
+} from "@ethersproject/providers";
+import { ProviderWithFallbackBase } from "./ProviderWithFallbackBase";
+
+export type ProviderWithFallbackConfig = {
+  unrecoverableErrors: ErrorCode[];
+  logger: { info: (message: string) => void; warn: (message: string) => void };
```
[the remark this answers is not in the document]
koonopek: Mine, changed it to use https://docs.ethers.org/v5/api/utils/logger/ (don't allow to pass own logger)

packages/rpc-providers/src/ProviderWithAgreement.ts
koonopek: done

packages/rpc-providers/src/ProviderWithAgreement.ts
koonopek: done

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    return blockNumbers.sort((a, b) => a - b).at(1) as number;
+  }
+
+  private async resolveCallPromises(promises: Promise<any>[]) {
+    const agreedResults: CallResult[] = [];
+    let resolvedCalls = 0;
+
+    while (true) {
```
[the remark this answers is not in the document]
koonopek: done

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    return this.resolveCallPromises(promises);
+  }
+
+  private async electBlockNumber(): Promise<number> {
+    let lastError;
+    const blockNumbers: number[] = [];
+
+    // collect block numbers
+    await Promise.all(
```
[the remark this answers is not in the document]
koonopek: done

packages/on-chain-relayer/src/core/contract-interactions/get-last-round-params.ts
```diff
@@ -1,11 +1,16 @@
-import { BigNumber, Contract, utils } from "ethers";
+import { BigNumber, Contract } from "ethers";
 
 const parseBigNumberParam = (valueInBigNumber: BigNumber) =>
-  Number(utils.formatUnits(valueInBigNumber, 0));
+  valueInBigNumber.toNumber();
+
+export type LastRoundParams = {
+  lastRound: number;
+  lastUpdateTimestamp: number;
```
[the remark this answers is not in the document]
koonopek: Resolved all conflicts

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    // collect block numbers
+    const blockNumbersResults = await Promise.allSettled(
+      this.providers.map((provider) => provider.getBlockNumber())
+    );
+
+    const errors = [];
+    for (const blockNumberResult of blockNumbersResults) {
+      if (blockNumberResult.status === "fulfilled") {
+        blockNumbers.push(blockNumberResult.value);
```
hatskier, line 69: I think this algorithm can be simplified if we get rid of the array

hatskier: I would also like to discuss the best way to select the right block number, as taking the second smaller one may be not the best option as well

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,125 @@
...
+      return blockNumbers[0];
+    }
+
+    return blockNumbers.sort((a, b) => a - b).at(1) as number;
+  }
+
+  private resolveCallPromises(callPromises: Promise<string>[]) {
+    return new Promise<string>((resolve, reject) => {
+      const errors: Error[] = [];
+      const agreedResults: string[] = [];
```
hatskier, line 90: I think this algorithm can fail in some cases. E.g. we call 10 promises and require at least 3 same responses. The first answer that we receive is `0x1`, but all the rest is `0x2`

hatskier: I think the proper implementation should have a mapping `(resultHash => {result, counter})`

hatskier (changes requested):

packages/rpc-providers/src/ProviderWithAgreement.ts
koonopek: Yes we discussed that this is okey. To be naive this way, because we are waiting only for two responses. If not we should always wait for all responses or increase quorum to at least 3 same answers.

packages/rpc-providers/src/ProviderWithAgreement.ts
koonopek: That was the option we have chosen when we were discussing implementation, I was proposing median or the lowest block, but you suggested that heuristic to take the second oldest block should be the best. I agreed, and still agree. We didn't wanted to build full quorum provider, we just wanted to have confirmation that at least two nodes agree on the same answer.

koonopek: done

packages/on-chain-relayer/src/core/contract-interactions/get-contract.ts
```diff
@@ -1,17 +1,20 @@
...
+import { abi as mentoAdapterABI } from "../../../artifacts/contracts/custom-integrations/mento/MentoAdapterBase.sol/MentoAdapterBase.json";
 import { IRedstoneAdapter } from "../../../typechain-types";
+import { config } from "../../config";
+import { getProvider, getSigner } from "./get-provider-or-signer";
 
-export const getAdapterContract = () => {
-  const { privateKey, adapterContractAddress } = config;
-  const provider = getProvider();
-  const signer = new Wallet(privateKey, provider);
+export const getAdapterContract = (signerOrProvider?: Signer | Provider) => {
```
cehali, line 10: Is change is necessary?

packages/on-chain-relayer/src/core/contract-interactions/get-provider-or-signer.ts
```diff
@@ -1,16 +1,37 @@
...
+        ErrorCode.NONCE_EXPIRED,
+      ],
+    });
+  }
 };
 
 export const getSigner = () => {
-  const provider = getProvider();
-  const signer = new Wallet(config.privateKey, provider);
+  const signer = new Wallet(config.privateKey, getProvider());
```
cehali, line 35: `export const getSigner = () => new Wallet ...`

packages/rpc-providers/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "@nomiclabs/hardhat-etherscan": "^3.0.0",
+    "@types/chai": "^4.3.4",
+    "@types/mocha": "^10.0.1",
+    "chai": "^4.3.7",
+    "ethers": "5.7.2",
+    "hardhat": "^2.13.0",
+    "sinon": "^15.0.3"
+  },
+  "peerDependencies": {
+    "ethers": "5.x.x"
```
cehali, line 29: You can point the last 5 version

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,147 @@
...
+  blockNumbers: number[],
+  numberOfProviders: number
+): number => {
+  if (blockNumbers.length === 1) {
+    return blockNumbers[0];
+  }
+
+  return blockNumbers
+    .sort((a, b) => a - b)
+    .at(Math.floor(numberOfProviders / 3)) as number;
```
cehali, line 28: Maybe we could take the minimal block number?

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  private resolveCallPromises(callPromises: Promise<string>[]) {
+    return new Promise<string>((resolve, reject) => {
+      const errors: Error[] = [];
+      const results = new Map<string, number>();
+      let handledResults = 0;
+
+      for (const callPromise of callPromises) {
+        callPromise
+          .then((currentResult) => {
```
cehali, line 116: Why not try catch finally?

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,147 @@
...
+            const currentResultCount = results.get(currentResult);
```
cehali, line 117: What could be the value of `currentResult` string? Is it possible that we will have multiple of the same values?

packages/rpc-providers/tsconfig.build.json
```diff
@@ -0,0 +1,22 @@
+{
```
cehali, line 1: Why we need this file?

packages/rpc-providers/src/ProviderWithAgreement.ts
koonopek: In the end I made this configurable and as default function I have proposed to take `floor(provider.length /3 )`, so depending on the providers count: 
```
5 -> blocks[1] (4/5 synced providers)
6 -> blocks[3] (4/6 synced providers)
7 -> blocks[3] (5/7 synced providers)
10 -> blocks[5] (6/10 synced providers)
```
If we would pick median we would eliminate half of the providers on the start, I think this is too restrictive

packages/on-chain-relayer/src/core/contract-interactions/get-contract.ts
koonopek: Such a method is easier to test because we can inject mock provider for example.
Does it break something?

packages/rpc-providers/package.json
koonopek: It is peerDependecies so it is saying like: I should work with every ethers 5 version. I think there is no added value in narrowing this.

packages/rpc-providers/src/ProviderWithAgreement.ts
koonopek: Yes this is expected that different providers on the same block height will return same answer.
This will be result of read method on some contract

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    // we want to avoid situation were 2 calls fail in very short period
+    // if this happens we would choose effectively as new provider this.providerIndex + 2 % this.providers.length
+    // if first request already has changed provider we do nothing
+    if (lastUsedProviderIndex !== this.providerIndex) {
+      return;
+    }
+
+    const nextProviderIndex = (this.providerIndex + 1) % this.providers.length;
```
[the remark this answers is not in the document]
koonopek: If the last provider failed we start from the begging of the array, but for the same request we can try every provider only once.

packages/rpc-providers/tsconfig.build.json
koonopek: We don't wan't to include `tests` in `dist` folder, but still want to type check them :)
Also if we try to use typescript `references` without excluding hardhat tests it will fail. too

packages/rpc-providers/src/ProviderWithAgreement.ts
cehali: Cool, I get it now 👍

packages/on-chain-relayer/src/core/contract-interactions/get-contract.ts
cehali: Noo, just `signerOrProvider ?? getSigner()` seems a little bit strange 😄

packages/rpc-providers/src/ProviderWithAgreement.ts
cehali: Ok, I wasn't aware of this. Thanks for explanation 👍

packages/on-chain-relayer/test/contracts/price-feeds/old/PriceFeed.test.tes
```diff
@@ -82,7 +82,9 @@ describe("PriceFeed", () => {
       const roundData = await contract.getRoundData(roundId);
       expect(roundData.answer).to.be.equal(167099000000);
       expect(roundData.roundId).to.be.equal(roundId);
-      expect(roundData.updatedAt).to.be.equal(expectedTimestamp);
+      expect(
```
hatskier, line 85: I think we can remove this file xD

packages/rpc-providers/README.md
```diff
@@ -0,0 +1,19 @@
...
+4. Providers are elected by the sequence given in array. If array ends we start from the beginning.
+
+## ProviderWithAgreement
+
+1. Works like `ProviderWithFallback, with an exception for two methods.
+2. `getBlockNumber` asks all providers for blockNumber and then picks `blockNumber.sort(ASC)[1]`
+3. `call` ask all providers for result
+   - if at least `N` answers are the same, it return call result
+   - ignore all errors
+   - as soon as it will receive two matching responses returns. (doesn't wait for the rest of the providers to finish)
```
hatskier, line 19: Not two matching responses, but N matching responses

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,147 @@
...
+    this.agreementConfig = {
+      ...defaultConfig,
+      ...this.getProviderWithFallbackConfig(),
+    };
+    if (
+      this.agreementConfig.numberOfNodesWhichHaveToAgree < 2 ||
+      this.agreementConfig.numberOfNodesWhichHaveToAgree > this.providers.length
+    ) {
+      throw new Error(
+        "numberOfNodesWhichHaveToAgree should be >= 2 and > then supplied providers count"
```
hatskier, line 57: numberOfNodesWhichHaveToAgree -> numberOfProvidersWhichHaveToAgree?

packages/rpc-providers/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+1. If the provider fails on some operation new provider is elected.
+2. If all providers fail on the same operation error is thrown.
+3. When we switch to the next provider, all operations (excluding 1.) are executed by it till it fails.
+4. Providers are elected by the sequence given in array. If array ends we start from the beginning.
+
+## ProviderWithAgreement
+
+1. Works like `ProviderWithFallback, with an exception for two methods.
+2. `getBlockNumber` asks all providers for blockNumber and then picks `blockNumber.sort(ASC)[1]`
```
hatskier, line 15: This text doesn't match with the implementation

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,147 @@
...
+              results.set(currentResult, currentResultCount + 1);
+              if (
+                currentResultCount + 1 ===
+                this.agreementConfig.numberOfNodesWhichHaveToAgree
+              ) {
+                resolve(currentResult);
+              }
+            }
+
+            results.set(currentResult, 1);
```
hatskier, line 130: Wouldn't it overwrite the value to 1, even if before we set it to another value? I think it can cause problems with all expectedProvidersToAgree except for 2

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  protected override executeWithFallback(
+    fnName: string,
+    ...args: any[]
+  ): Promise<any> {
+    return this.doExecuteWithFallback(0, this.providerIndex, fnName, ...args);
+  }
+
+  private async doExecuteWithFallback(
+    retryNumber = 0,
```
hatskier, line 160: I would call it alreadyRetriedCount as currently it's difficult to differentiate if from number of allowed retries

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,237 @@
...
+    logger.warn(
+      `Provider ${providerName} failed with error: ${error?.code} ${error?.message}`
+    );
+
+    if (retryNumber === this.providers.length - 1) {
+      logger.warn(`All providers failed to execute action, rethrowing error`);
+      throw error;
+    }
+
+    // we want to avoid situation were 2 calls fail in very short period
```
hatskier, line 223: IMHO this comment is a bit misleading:
- I am not sure I understand this expression and not sure it's valid `this.providerIndex + 2 % this.providers.length`
- I would also shorten it to smth like this `// if another concurrent request already has changed provider we do nothing`

hatskier (changes requested):

packages/on-chain-relayer/test/contracts/price-feeds/old/PriceFeed.test.tes
koonopek: We have a lot of files like this, I think they should be removed in separate PR

packages/rpc-providers/README.md
koonopek: good

packages/rpc-providers/src/ProviderWithAgreement.ts
koonopek: why not

packages/rpc-providers/README.md
koonopek: Yeah, updated

packages/rpc-providers/src/ProviderWithAgreement.ts
koonopek: Yes, I ate `else` statement very good catch :)

packages/rpc-providers/src/ProviderWithFallback.ts
koonopek: Definitely better name!

packages/rpc-providers/src/ProviderWithFallback.ts
koonopek: * I think it is valid I meant that we would jum by two providers ahead effectively instead of 1 (as it is happening now). But I have removed it because it make more trouble then help
* good point

hatskier (changes requested):

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,205 @@
...
+  sleepBetweenBlockSynReq: number;
+  blockNumberCacheTTLInMS: number;
+  electBlockFn: (blocks: number[], numberOfAgreeingNodes: number) => number;
+}
+
+const DEFAULT_ELECT_BLOCK_FN = (
+  blockNumbers: number[],
+  numberOfProviders: number
+): number => {
+  if (blockNumbers.length === 1) {
```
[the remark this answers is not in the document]
koonopek: I tried but this just median i think it can't be simpler.
Remove first `if`

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,205 @@
...
+    providers: JsonRpcProvider[],
+    config: Partial<ProviderWithAgreementConfig> = {}
+  ) {
+    super(providers, config);
+    this.agreementConfig = {
+      ...defaultConfig,
+      ...this.getProviderWithFallbackConfig(),
+    };
+    if (
+      this.agreementConfig.numberOfProvidersWhichHaveToAgree < 2 ||
```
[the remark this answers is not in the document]
koonopek: done

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  private executeCallWithAgreement(
+    transaction: Deferrable<TransactionRequest>,
+    electedBlockTag: BlockTag
+  ) {
+    return new Promise<string>((resolve, reject) => {
+      const electedBlockNumber = convertBlockTagToNumber(electedBlockTag);
+      const errors: Error[] = [];
+      const results = new Map<string, number>();
```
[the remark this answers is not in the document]
koonopek: Why not it has `has` method which is cleaner then calling `obj['abc']` because what if `obj['abc']=0` it is false positive. I'm not sure if storing biggg strings in objects will work properly.
I didn't find any benchamrks in favour of Objects:
1. https://stackoverflow.com/questions/18541940/map-vs-object-in-javascript
2. https://www.zhenghao.io/posts/object-vs-map
3. https://bretcameron.medium.com/how-javascript-maps-can-make-your-code-faster-90f56bf61d9d (don't like the methodology here)

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,205 @@
...
+
+      const sync = async (providerIndex: number) => {
+        if (stop) return;
+
+        while (blockPerProvider[providerIndex] < electedBlockNumber) {
+          blockPerProvider[providerIndex] = await this.providers[providerIndex]
+            .getBlockNumber()
+            // ignore errors try again later
+            .catch();
+          if (stop) {
```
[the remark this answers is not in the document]
koonopek: Good idea

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,205 @@
...
+          }
+        } catch (e: any) {
+          errors.push(e);
+        }
+      };
+
+      const syncThenCall = async (providerIndex: number) => {
+        await sync(providerIndex);
+        await call(providerIndex);
+        handledResults += 1;
```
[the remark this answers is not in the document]
koonopek: Meh, I don't like this syntax, but I don't mind :)

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,205 @@
...
+  private blockNumberCache = { value: 0, lastUpdate: 0n };
+
+  constructor(
+    providers: JsonRpcProvider[],
+    config: Partial<ProviderWithAgreementConfig> = {}
+  ) {
+    super(providers, config);
+    this.agreementConfig = {
+      ...defaultConfig,
+      ...this.getProviderWithFallbackConfig(),
```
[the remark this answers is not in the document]
koonopek: done

packages/rpc-providers/tsconfig.json
```diff
@@ -0,0 +1,31 @@
...
+  "include": [
+    "./src",
+    "./test",
+    "./typechain-types",
+    "package.json",
+    "./contracts",
+    "src/**/*.json",
+    "./artifacts/**/*.json"
+  ],
+  "files": [
```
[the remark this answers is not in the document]
koonopek: Don't know some auto magic - fixed

packages/rpc-providers/tsconfig.build.json
```diff
@@ -0,0 +1,22 @@
...
+    "outDir": "dist",
+    "declaration": true,
+    "declarationMap": true,
+    "incremental": true,
+    "composite": true
+  },
+  "include": ["./src", "package.json", "./contracts", "src/**/*.json"],
+  "files": [
+    "./src/ProviderWithFallback.ts",
+    "./src/ProviderWithFallbackBase.ts"
```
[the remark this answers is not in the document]
koonopek: Fixed

packages/rpc-providers/README.md
```diff
@@ -0,0 +1,24 @@
+# Magic providers, compatible with ethers.Provider interface
+
+**Tested only with JsonRpcProviders**
+
+## ProviderWithFallback
+
+1. If the provider fails on some operation new provider is elected.
+2. If all providers fail on the same operation error is thrown.
+3. When we switch to the next provider, all operations (excluding 1.) are executed by it till it fails.
```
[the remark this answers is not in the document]
koonopek: Hmm good question

packages/rpc-providers/src/common.ts
```diff
@@ -0,0 +1,16 @@
+export const timeout = <T>(prom: Promise<T>, timeoutMS: number): Promise<T> => {
+  let timer: NodeJS.Timeout;
+  return Promise.race<T>([
+    prom,
+    new Promise(
+      (_r, reject) =>
+        (timer = setTimeout(
+          () => reject(`Timeout error ${timeoutMS} [MS]`),
```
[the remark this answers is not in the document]
koonopek: oki doki

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  private async doExecuteWithFallback(
+    alreadyRetriedCount = 0,
+    lastProviderUsedIndex: number,
+    fnName: string,
+    ...args: any[]
+  ): Promise<any> {
+    try {
+      return await (this.currentProvider as any)[fnName](...[...args]);
```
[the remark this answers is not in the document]
koonopek: It shallow copy args.
I think we could leave just `...args`

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -0,0 +1,239 @@
...
+
+    // if another concurrent request already has changed provider we do nothing
+    if (lastUsedProviderIndex !== this.providerIndex) {
+      return;
+    }
+
+    const nextProviderIndex = (this.providerIndex + 1) % this.providers.length;
+    const nextProviderName = this.extractProviderName(nextProviderIndex);
+
+    logger.info(`Fallback in to next provider ${nextProviderName}.`);
```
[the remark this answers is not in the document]
koonopek: Good idea

packages/rpc-providers/test/ProviderWithFallback.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      stubProvider,
+      stubProvider,
+    ]);
+
+    spy.onFirstCall().rejects().onSecondCall().resolves(10);
+
+    expect(await fallbackProvider.getBlockNumber()).to.eq(10);
+  });
+
+  it("should call first provider if second fails, and the second provider is currently used", async () => {
```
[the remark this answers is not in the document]
koonopek: Yep

packages/rpc-providers/test/ProviderWithFallback.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    const stubProvider = new providers.StaticJsonRpcProvider();
+    const spy = sinon.stub(stubProvider, "getBlockNumber");
+
+    const fallbackProvider = new ProviderWithFallback([
+      stubProvider,
+      stubProvider,
+    ]);
+
+    spy.onFirstCall().rejects().onSecondCall().resolves(10);
```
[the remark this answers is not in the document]
koonopek: Good idea, will do

packages/rpc-providers/test/ProviderWithFallback.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    await Promise.allSettled([
+      fallbackProvider.getBlockNumber(),
+      fallbackProvider.getBlockNumber(),
+    ]);
+
+    expect(fallbackProvider.getCurrentProviderIndex()).to.eq(1);
+  });
+});
+
+async function deployCounter() {
```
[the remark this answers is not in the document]
koonopek: ;)

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -0,0 +1,205 @@
...
+      let handledResults = 0;
+
+      const sync = async (providerIndex: number) => {
+        if (stop) return;
+
+        while (blockPerProvider[providerIndex] < electedBlockNumber) {
+          blockPerProvider[providerIndex] = await this.providers[providerIndex]
+            .getBlockNumber()
+            // ignore errors try again later
+            .catch();
```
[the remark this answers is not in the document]
hatskier: Hmm, I am still a bit afraid of never resolved or rejected promise if all rpcs just fail to return block number. They will never be synced but will always retry.

I would propose a small change, e.g. to stop trying to sync for a given rpc if one call has failed. Let's discuss it tomorrow

hatskier (approved):

PR #516, opened 2023-04-06
feat: split price feed contracts to round-less and round-enabled versions

hatskier:


packages/on-chain-relayer/contracts/core/IPermissionlessPriceUpdater.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity ^0.8.4;
+
+interface IPermissionlessPriceUpdater {
+  function getMinIntervalBetweenUpdates() external view returns (uint256);
+  function validateProposedTimestamp(uint256 proposedTimestamp) external view;
```
jakub-wojciechowski, line 7: isProposedTimestampValid? 
(to getter nie funkcja)

packages/on-chain-relayer/contracts/core/PermissionlessPriceUpdater.sol
```diff
@@ -1,44 +1,52 @@
...
-  // We don't use storage variables to avoid problems with upgradable contracts
-  uint256 constant LAST_ROUND_STORAGE_LOCATION =
-    0x919ecb282edbbb41bface801311ec7a6df61da05d3d63b938d35b526a69d4d6d; // keccak256("RedStone.lastRound");
-  uint256 constant LAST_UPDATED_TIMESTAMP_STORAGE_LOCATION =
+contract PermissionlessPriceUpdater is IPermissionlessPriceUpdater {
+  // We don't use storage variables to avoid potential problems with upgradable contracts
+  bytes32 constant LAST_UPDATED_TIMESTAMP_STORAGE_LOCATION =
     0x3d01e4d77237ea0f771f1786da4d4ff757fcba6a92933aa53b1dcef2d6bd6fe2; // keccak256("RedStone.lastUpdateTimestamp");
 
+  uint256 constant MIN_MILLISECONDS_INTERVAL_BETWEEN_UPDATES = 10_000;
```
jakub-wojciechowski, line 19: Doesn't it limit us too much in terms of latency, what if we're running in 1s update mode?

packages/on-chain-relayer/contracts/price-feeds/PriceFeedBase.sol
```diff
@@ -0,0 +1,59 @@
...
+
+  // Below are methods that are not part of the AggregatorV3Interface,
+  // but are still used by some projects integrated with Chainlink (e.g. GMX)
+
+  function latestAnswer() public view returns (int256) {
+    return int256(getPriceFeedAdapter().getValueForDataFeed(dataFeedId));
+  }
+
+  function latestRound() public view virtual returns (uint80) {
+    return 0;
```
jakub-wojciechowski, line 57: Do we support latest round here - shouldn't we throw error?

packages/on-chain-relayer/contracts/price-feeds/PriceFeedsAdapterBase.sol
```diff
@@ -0,0 +1,96 @@
...
+  RedstoneConsumerNumericBase,
+  Ownable,
+  PermissionlessPriceUpdater,
+  IPriceFeedAdapter
+{
+  using EnumerableSet for EnumerableSet.Bytes32Set;
+
+  EnumerableSet.Bytes32Set private dataFeedsIds;
+
+  constructor(bytes32[] memory dataFeedsIds_) {
```
jakub-wojciechowski, line 21: Let's keep all the contract initializable pls.

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -0,0 +1,220 @@
...
+ * @author The Redstone Oracles team
+ * @dev This contract is used to repeatedly push RedStone data to blockchain storage
+ * More details here: https://docs.redstone.finance/docs/smart-contract-devs/get-started/redstone-classic
+ *
+ * Key details about the contract:
+ * - Data feed values can be updated using the `updateDataFeedValues` function
+ * - All data feeds must be updated within a single call, partial updates are not allowed
+ * - There is a configurable minimum interval between updates
+ * - Updaters can be restricted by overriding `requireAuthorisedUpdater` function
+ * - The contract is designed to force values validation, be default it prevents returning zero values
```
cehali, line 19: `be` -> `by`

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -0,0 +1,220 @@
...
+    uint256 dataTimestampFromLatestUpdate = getDataTimestampFromLatestUpdate();
+
+    // We intentionally allow to update data with the same timestamp
+    if (dataPackagesTimestamp < dataTimestampFromLatestUpdate) {
+      revert DataTimestampCanNotBeOlderThanBefore(
+        dataPackagesTimestamp,
+        dataTimestampFromLatestUpdate
+      );
+    }
+  }
```
cehali, line 132: Maybe move it to `RedstoneDefaultsLib` to have all validations there?

packages/on-chain-relayer/src/scripts/price-feeds/deploy-price-feeds-contracts.ts
```diff
@@ -58,7 +58,7 @@ import { config } from "../../config";
       dataPackages[dataFeeds[0]][0].dataPackage.timestampMilliseconds;
 
     const firstRound = 1;
-    const updateTransaction = await wrappedContract.updateDataFeedsValues(
+    const updateTransaction = await wrappedContract.updateDataFeedValues(
```
cehali, line 61: I am not sure of this rename, you are updating values of multiple data feeds

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -0,0 +1,220 @@
...
+  function _unpackTimestamps(uint256 packedTimestamps)
+    internal
+    pure
+    returns (uint128 dataTimestamp, uint128 blockTimestamp)
+  {
+    dataTimestamp = uint128(packedTimestamps >> 128); // first 128 bits
+    blockTimestamp = uint128(packedTimestamps); // last 128 bits
+  }
+
+  function _saveTimestampsOfCurrentUpdate(uint256 dataPackagesTimestamp) private {
```
cehali, line 166: Underscore is `_saveTimestampsOfCurrentUpdate` is necessary?

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  function getValuesForDataFeeds(bytes32[] memory requestedDataFeedIds)
+    public
+    view
+    returns (uint256[] memory)
+  {
+    uint256[] memory values = getValuesForDataFeedUnsafe(requestedDataFeedIds);
+    for (uint256 i = 0; i < requestedDataFeedIds.length; i++) {
+      validateDataFeedValue(requestedDataFeedIds[i], values[i]);
```
cehali, line 191: We validate values when updating, do we need to update them also when getting?

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -0,0 +1,220 @@
...
+
+  function getDataTimestampFromLatestUpdate() public view returns (uint256 lastDataTimestamp) {
+    (lastDataTimestamp, ) = getTimestampsFromLatestUpdate();
+  }
+
+  function getBlockTimestampFromLatestUpdate() public view returns (uint256 blockTimestamp) {
+    (, blockTimestamp) = getTimestampsFromLatestUpdate();
+  }
+
+  function getPackedTimestampsFromLatestUpdate() public view returns (uint256 packedTimestamps) {
```
jakub-wojciechowski, line 142: Are we going to use this function anywhere without unpacking (We produced a lot of code and hoops for relatively simple operation)

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -0,0 +1,220 @@
...
+    blockTimestamp = uint128(packedTimestamps); // last 128 bits
+  }
+
+  function _saveTimestampsOfCurrentUpdate(uint256 dataPackagesTimestamp) private {
+    uint256 blockTimestamp = block.timestamp;
+    assembly {
+      let timestamps := packTwoNumbers(dataPackagesTimestamp, blockTimestamp)
+      sstore(LATEST_UPDATE_TIMESTAMPS_STORAGE_LOCATION, timestamps)
+
+      function packTwoNumbers(num1, num2) -> resultNumber {
```
jakub-wojciechowski, line 172: Cannot we inline this operation?

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -0,0 +1,220 @@
...
+    virtual
+  {
+    preventUpdateWithOlderDataPackages(dataPackagesTimestamp);
+    RedstoneDefaultsLib.validateTimestamp(dataPackagesTimestamp);
+  }
+
+  function preventUpdateWithOlderDataPackages(uint256 dataPackagesTimestamp) internal view {
+    uint256 dataTimestampFromLatestUpdate = getDataTimestampFromLatestUpdate();
+
+    // We intentionally allow to update data with the same timestamp
```
koonopek, line 125: Why?

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -0,0 +1,220 @@
...
+      );
+    }
+  }
+
+  function validateAndUpdateDataFeedValues(
+    bytes32[] memory dataFeedIdsArray,
+    uint256[] memory values
+  ) internal virtual;
+
+  function validateCurrentBlockTimestamp() private view {
```
koonopek, line 94: Maybe `assertMinIntervalBetweenUpdatesPassed` is more meaningful name?

packages/on-chain-relayer/contracts/price-feeds/without-rounds/PriceFeedsAdapterWithoutRounds.sol
```diff
@@ -0,0 +1,34 @@
...
+
+pragma solidity ^0.8.4;
+
+import "../PriceFeedsAdapterBase.sol";
+
+abstract contract PriceFeedsAdapterWithoutRounds is PriceFeedsAdapterBase {
+  bytes32 constant VALUES_MAPPING_STORAGE_LOCATION =
+    0x4dd0c77efa6f6d590c97573d8c70b714546e7311202ff7c11c484cc841d91bfc; // keccak256("RedStone.oracleValuesMapping");
+
+  function updateDataFeedValue(bytes32 dataFeedId, uint256 dataFeedValue) internal override {
```
koonopek, line 11: Shouldn't we revert if dataFeedId is not supported? DataFeedsArray?

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -0,0 +1,220 @@
...
+      sstore(LATEST_UPDATE_TIMESTAMPS_STORAGE_LOCATION, timestamps)
+
+      function packTwoNumbers(num1, num2) -> resultNumber {
+        resultNumber := or(shl(BITS_COUNT_IN_16_BYTES, num1), num2)
+      }
+    }
+  }
+
+  function getValueForDataFeed(bytes32 dataFeedId) public view returns (uint256) {
+    uint256 valueForDataFeed = getValueForDataFeedUnsafe(dataFeedId);
```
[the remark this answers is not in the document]
koonopek: For example in WithoutRounds implementation

packages/on-chain-relayer/contracts/core/PermissionlessPriceUpdater.sol
hatskier: We always can override it in the derived contracts, but IMO it's better to have 10 seconds by default to prevent unefficient gas spending in case of several independent relayers with different private keys

packages/on-chain-relayer/contracts/price-feeds/PriceFeedBase.sol
hatskier: We support only round with id 0. I did it this way intentionally for being able to integrate this "lite" version of the contract even if the project does something weird, but always relies on the latest round (e.g. something like `getRoundData(latestRound())`

packages/on-chain-relayer/contracts/price-feeds/PriceFeedsAdapterBase.sol
hatskier: Will do

packages/on-chain-relayer/contracts/price-feeds/data-services/PriceFeedsAdapterWithoutRoundsMainDemo.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity ^0.8.4;
+
+import "../without-rounds/PriceFeedsAdapterWithoutRounds.sol";
+
+contract PriceFeedsAdapterLiteMainDemo is PriceFeedsAdapterWithoutRounds {
```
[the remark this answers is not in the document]
hatskier: Good catch, thx!

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
hatskier: It would be difficult and IMO not really clear, as RedstoneDefaultsLib is located in a separate package and (again, my opinion) should not contain logic related to our classic model

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
hatskier: It's a good practice to add underscores to internal and private functions in solidity

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
hatskier: In my opinion, we should, as it increases security and also prevent accidental problems (e.g. if we forgot to set some value it's 0 by default, then it will not be returned). This validation is also really cheap, so IMO it's worth it

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
hatskier: Yes, we use it on PriceFeedsAdapters (esp. the one with rounds)

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
hatskier: My original thought was that sometimes we may want to update prices more frequently than our oracle nodes update them (e.g. in case of Mento Adapter we may want to push prices every 5 seconds, but the data is updated by nodes every 10 seconds). But as @jakub-wojciechowski noticed, it introduces a significant security risk, becasue potential attackers would be able to combine different subsets of data packages with the same timestamps to manipulate prices (if we use 3 of 5 signers for example). So I will set requiremment of the bigger timestamps again

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
hatskier: I agree, this name is much better

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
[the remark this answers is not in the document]
hatskier: I agree, adding this function will signifcantly increase security and reduce a number of potential weird edge cases. Will implement it

packages/on-chain-relayer/contracts/price-feeds/without-rounds/PriceFeedsAdapterWithoutRounds.sol
hatskier: We should, will implement it

hatskier: Actually, in this case, maybe we should not check if the dataFeedId is supported, because this function is internal and is called only for supported data feed ids received from the `getDataFeedIds` function

packages/on-chain-relayer/contracts/core/IPermissionlessPriceUpdater.sol
hatskier: I would like to stay with the current approach to be consistent with other validation functions that we use in the core model and here for updaters authorisation

packages/on-chain-relayer/contracts/price-feeds/PriceFeedBase.sol
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  function latestRoundData() public view override returns (uint80 roundId, int256 answer, uint256 startedAt, uint256 updatedAt, uint80 answeredInRound) {
+    roundId = latestRound();
+    answer = latestAnswer();
+
+    (uint128 dataTimestamp, uint128 blockTimestamp) = getPriceFeedAdapter()
+      .getTimestampsFromLatestUpdate();
+
+    startedAt = dataTimestamp / 1000; // convert to seconds
```
koonopek, line 36: I think, this should be somewhere well documented because it doesn't align with Chainlink Aggregator logic

packages/on-chain-relayer/contracts/price-feeds/without-rounds/SinglePriceFeedAdapter.sol
```diff
@@ -0,0 +1,74 @@
...
+
+  function getValueForDataFeedUnsafe(bytes32) public view virtual override returns (uint256 dataFeedValue) {
+    uint160 dataFeedValueCompressed;
+    assembly {
+      dataFeedValueCompressed := sload(DATA_FROM_LATEST_UPDATE_STORAGE_LOCATION)
+    }
+    return uint256(dataFeedValueCompressed);
+  }
+
+  function getTimestampsFromLatestUpdate() public view virtual override returns (uint128 dataTimestamp, uint128 blockTimestamp) {
```
koonopek, line 46: Why are overrding this function from RedstoneAdapterBase?

packages/on-chain-relayer/contracts/price-feeds/without-rounds/SinglePriceFeedAdapterWithClearing.sol
```diff
@@ -0,0 +1,67 @@
...
+  function _updateDataFeedValue(bytes32 dataFeedId, uint256 dataFeedValue) internal override {
+    validateDataFeedValue(dataFeedId, dataFeedValue);
+    uint256 blockTimestampCompressedAndShifted = block.timestamp << 208; // Move value to the first 48 bits
+    assembly {
+      // Save timestamp and data feed value
+      let timestampAndValue := or(blockTimestampCompressedAndShifted, dataFeedValue)
+      sstore(DATA_FROM_LATEST_UPDATE_STORAGE_LOCATION, timestampAndValue)
+
+      // Clear temp data timestamp, it refunds 19.9k gas
+      sstore(TEMP_DATA_TIMESTAMP_STORAGE_LOCATION, 0)
```
koonopek, line 58: Maybe instead storing it and then clearing we could pass it as parameter, or there is not difference in gas in this two ways?

packages/on-chain-relayer/contracts/price-feeds/PriceFeedsAdapterBase.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity ^0.8.4;
+
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "../core/RedstoneAdapterBase.sol";
+
+abstract contract PriceFeedsAdapterBase is RedstoneAdapterBase, Initializable {
```
koonopek, line 8: What is the purpose of this class?

koonopek: I mean that all of this could be in PriceFeedsAdapter, don't see any pros in extract it to "Base" class.

packages/on-chain-relayer/contracts/price-feeds/with-rounds/PriceFeedsAdapterWithRounds.sol
```diff
@@ -0,0 +1,100 @@
...
+    _updatePackedTimestampsForLatestRound();
+
+    for (uint256 i = 0; i < dataFeedsIdsArray.length; i++) {
+      bytes32 dataFeedId = dataFeedsIdsArray[i];
+      _updateDataFeedValue(dataFeedId, values[i]);
+    }
+  }
+
+  function _updateDataFeedValue(bytes32 dataFeedId, uint256 dataFeedValue) internal virtual override {
+    validateDataFeedValue(dataFeedId, dataFeedValue);
```
koonopek, line 28: validateDataFeedValue -> validateDataFeed?

packages/on-chain-relayer/hardhat.config.ts
```diff
@@ -1,8 +1,37 @@
...
+import "@openzeppelin/hardhat-upgrades";
+import "hardhat-gas-reporter";
 import "@nomicfoundation/hardhat-toolbox";
+import "@nomiclabs/hardhat-waffle";
+import "@nomiclabs/hardhat-ethers";
+import { HardhatUserConfig } from "hardhat/config";
+
+dotenv.config();
 
 const config: HardhatUserConfig = {
```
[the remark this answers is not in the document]
koonopek: For example, deployment costs decrease with the decrease in "runs"—which specifies how often each opcode will be executed over the life of a contract. So we can even increase this number to something like 100,000,000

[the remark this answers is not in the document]
koonopek: https://github.com/smartcontractkit/chainlink/blob/18edbf4fb1f3601d75808ffde246112b8687114a/contracts/hardhat.config.ts#L12

packages/on-chain-relayer/test/contracts/price-feeds/common/price-feed-utils.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+interface PriceFeedTestsContracts {
+  adapter: IRedstoneAdapter;
+  priceFeed: PriceFeedBase;
+}
+
+chai.use(chaiAsPromised);
+
+export const describeCommonPriceFeedTests = ({
```
koonopek, line 22: NEAT!

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -0,0 +1,223 @@
...
+    }
+  }
+
+  function validateAndUpdateDataFeedsValues(bytes32[] memory dataFeedIdsArray, uint256[] memory values) internal virtual;
+
+  function _assertMinIntervalBetweenUpdatesPassed() private view {
+    uint256 currentBlockTimestamp = block.timestamp;
+    uint256 blockTimestampFromLatestUpdate = getBlockTimestampFromLatestUpdate();
+    uint256 minIntervalBetweenUpdates = getMinIntervalBetweenUpdates();
+    if (currentBlockTimestamp < blockTimestampFromLatestUpdate + minIntervalBetweenUpdates) {
```
koonopek, line 106: I have test this on arbitrum mainnet (by fetching blocks from provider) and it comes out that blocks are quite frequently produced with same timestamp.
This two consecutive block have same timestamps. Maybe we should switch to using block.number
part of the testing script output:
```
{
  blockDiff: '84437183-84437184',
  prev: 1682500112,
  curr: 1682500112,
  delta: 0
}
This two consecutive block have same timestamps
{
  blockDiff: '84437184-84437185',
  prev: 1682500112,
  curr: 1682500112,
  delta: 0
}
Summary 75 of 100 blocks has same timestamp
```

packages/on-chain-relayer/test/contracts/price-feeds/common/price-feed-utils.ts
```diff
@@ -0,0 +1,138 @@
...
+      curBlockTime = prevBlockTime + 10;
+      mockDataTimestamp = prevBlockTime * 1000;
+      await time.setNextBlockTimestamp(curBlockTime);
+
+      const wrappedContract = (await WrapperBuilder.wrap(
+        contracts.adapter
+      ).usingSimpleNumericMock({
+        mockSignersCount: 2,
+        timestampMilliseconds: mockDataTimestamp,
+        dataPoints: [{ dataFeedId: "BTC", value: 42 }],
```
koonopek, line 98: Shouldn't we somehow verify that decimals precision that we are sending is equal to decimals precision described in priceFeed?
 Or at least put somewhere big warning that developers should check that.

koonopek: Yeah I think warning is enough

packages/on-chain-relayer/contracts/price-feeds/with-rounds/PriceFeedWithRounds.sol
```diff
@@ -0,0 +1,28 @@
...
+    return uint80(getPriceFeedAdapterWithRounds().getLatestRoundId());
+  }
+
+  function getRoundData(uint80 requestedRoundId) public view override returns (uint80 roundId, int256 answer, uint256 startedAt, uint256 updatedAt, uint80 answeredInRound) {
+    (uint256 dataFeedValue, uint128 roundDataTimestamp, uint128 roundBlockTimestamp) = getPriceFeedAdapterWithRounds().getRoundData(
+      getDataFeedId(),
+      requestedRoundId
+    );
+    roundId = requestedRoundId;
+    answer = int256(dataFeedValue);
```
koonopek, line 23: This will overflow (in Solidity 0.7.x and older) or throw an exception (in Solidity 0.8+) for values between 2^255 (int max value) and 2^256-1 (uint max value).

koonopek: As I undestarnd we don't use all 32 bytes to store number, so it will never flow

packages/on-chain-relayer/contracts/price-feeds/PriceFeedBase.sol
hatskier: What do you mean?

packages/on-chain-relayer/contracts/price-feeds/without-rounds/SinglePriceFeedAdapterWithClearing.sol
hatskier: Can you elaborate how you'd do it?

packages/on-chain-relayer/contracts/price-feeds/PriceFeedsAdapterBase.sol
koonopek: I think that PriceFeedsAdapter is the only class using it.
And even if duplication is only one for loop.
It is sometimes hard to read what is the current implementation if we have to traverse so deep hierarchy of inheritance.
https://gordonc.bearblog.dev/dry-most-over-rated-programming-principle/ just the first article after googling `why DRY is not always good`

packages/on-chain-relayer/contracts/price-feeds/PriceFeedBase.sol
koonopek: The startedAt and updatedAt fields represent different things in our impl of AggregatorV3Interface and Chainlink. If users use it in their contracts in the same way as they did with chainlink, they could be surprised :)

cehali (dismissed):

cehali (approved):

packages/on-chain-relayer/contracts/price-feeds/PriceFeedsAdapterBase.sol
hatskier: Done

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -0,0 +1,204 @@
...
+    uint256 valueForDataFeed = getValueForDataFeedUnsafe(dataFeedId);
+    validateDataFeedValue(dataFeedId, valueForDataFeed);
+    return valueForDataFeed;
+  }
+
+  function getValuesForDataFeeds(bytes32[] memory requestedDataFeedIds) public view returns (uint256[] memory) {
+    uint256[] memory values = getValuesForDataFeedUnsafe(requestedDataFeedIds);
+    for (uint256 i = 0; i < requestedDataFeedIds.length; i++) {
+      bytes32 dataFeedId = requestedDataFeedIds[i];
+      getDataFeedIndex(dataFeedId); // will revert if data feed id is not supported
```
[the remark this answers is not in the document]
hatskier: Yes, optimiser will handle thigs like this, so I would leave it as is, as IMHO it's better for the code readability

packages/on-chain-relayer/contracts/price-feeds/PriceFeedBase.sol
hatskier: I agree, we need to document it very well

packages/on-chain-relayer/contracts/price-feeds/PriceFeedsAdapterBase.sol
hatskier: But we already have many PriceFeedsAdapters: The version with rounds and the version without ones. If we don't have this logic here - we would need to duplicate it

packages/on-chain-relayer/contracts/price-feeds/with-rounds/PriceFeedsAdapterWithRounds.sol
hatskier: IMHO it's not clear what it means to validate data feed, we validate its value and only use the data feed id to log more details in case the validation fails

packages/on-chain-relayer/hardhat.config.ts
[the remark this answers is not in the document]
hatskier: I agree, will do it in a separate PR

packages/on-chain-relayer/test/contracts/price-feeds/common/price-feed-utils.ts
hatskier: Thank you 😊

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
hatskier: I agree, that we should be ready for this change, as we discussed I will abstract the logic of getting the block time identifier (which can return block.timestamp or block.number, depending on the needs)

packages/on-chain-relayer/test/contracts/price-feeds/common/price-feed-utils.ts
hatskier: This is a mock wrapper, with the real wrapper we have no control over precision, because the oracle-nodes sign the data with values, which are already prepared with the correct precision

packages/on-chain-relayer/contracts/price-feeds/data-services/VSTPriceFeedsAdapter.sol
```diff
@@ -0,0 +1,56 @@
...
+    return 2; // 2 out of 3
+  }
+
+  function aggregateValues(uint256[] memory values) public pure override returns (uint256) {
+    if (values.length != 2) {
+      revert InvalidSignersCount(values.length);
+    }
+
+    _checkCircuitBreaker(values[0]);
+    _checkCircuitBreaker(values[0]);
```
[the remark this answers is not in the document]
hatskier: Yes, it's a PoC, I agree that we should clear the flag after, will do it in a separate PR

packages/on-chain-relayer/benchmarks/zkEVM.txt
```diff
@@ -0,0 +1,62 @@
+zkEVM:
```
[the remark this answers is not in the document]
hatskier: Why not

packages/on-chain-relayer/contracts/price-feeds/PriceFeedsAdapterBase.sol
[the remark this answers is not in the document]
hatskier: It's not unnecessary, as we want to have upgradable contracts

packages/on-chain-relayer/contracts/price-feeds/without-rounds/PriceFeedWithoutRounds.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.4;
+
+import "../PriceFeedBase.sol";
+
+abstract contract PriceFeedWithoutRounds is PriceFeedBase {
+  error UnsupportedRoundId(uint80 requestedRoundId);
```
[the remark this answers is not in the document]
hatskier: I will add it in a separate PR

PR #520, opened 2023-04-06
feat: replace LevelDB with SQLite

cehali:


koonopek:
Sorry for the interruption :) 
But maybe it would be profitable to use SQLite in WAL mode, is more efficient in this type of cases:
https://www.sqlite.org/wal.html
Code how to enable it using `knex`:
```js
  const nodeDb = knex({
    client: "better-sqlite3",
    connection: {
      filename: `sqlite/node.sqlite`,
    },
    useNullAsDefault: true,
    pool: {
      afterCreate: (conn, cb) => {
        // https://github.com/knex/knex/issues/4971#issuecomment-1030701574
        conn.pragma("journal_mode = WAL");
        cb();
      },
    },
  });
  ``

File: packages/oracle-node/src/db/local-db/index.ts [hunk 1]
```diff
@@ -23,62 +38,58 @@ interface LastPrices {
...
 };
 
-export const closeLocalLevelDB = async () => {
-  await pricesSublevel.close();
-  await db.close();
+export const closeLocalDB = () => {
+  db.close();
 };
 
 export const getPrices = async (
```
hatskier, line 65: Should it be async?

hatskier (changes requested):

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -1,14 +1,29 @@
...
-import { roundTimestamp } from "../../utils/timestamps";
 
-const PRICES_SUBLEVEL = "prices";
-const DEFAULT_LEVEL_OPTS = {
-  keyEncoding: "utf8",
-  valueEncoding: "json",
-};
+const PRICES_TABLE = "prices";
+
+const CREATE_TABLE_STATEMENT = `CREATE TABLE IF NOT EXISTS ${PRICES_TABLE} (symbol TEXT, prices TEXT)`;
```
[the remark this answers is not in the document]
cehali: Because it is stringified array of prices from the last 15 minutes. Sqlite doesn't have an array type and if we want to go with each row as a new price we will have to add some background process to delete prices older than 15 minutes.

packages/oracle-node/src/db/local-db/index.ts
cehali: Nope, it is not async: https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#close---this

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -1,14 +1,29 @@
...
+
+const CREATE_INDEX_STATEMENT = `CREATE UNIQUE INDEX IF NOT EXISTS idx_prices_symbol ON ${PRICES_TABLE} (symbol)`;
+
+const getSelectBySymbolStatement = (symbols: string[]) =>
```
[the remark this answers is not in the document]
cehali: It is how SQLite works, we need a placeholder for each symbol we want to retrieve. It doesn't work with join array as a single string passed to the query, more info: https://github.com/WiseLibs/better-sqlite3/issues/125#issuecomment-1190753970.

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -23,62 +37,58 @@ interface LastPrices {
...
+  db.close();
 };
 
 export const getPrices = async (
   symbols: string[]
 ): Promise<PriceValuesInLocalDB> => {
-  const valuesForSymbols = await pricesSublevel.getMany(symbols);
+  const statement = db.prepare(getSelectForSymbolsStatement(symbols));
 
+  const valuesForSymbols: ReturnPricesFromDb = statement.all(symbols);
```
hatskier, line 69: I think we need to do performance benchmarks to verify is the sync operations on sqlite (with AWS ECS + EFS) will not slow down the oracle-node

PR #529, opened 2023-04-07
feat: add stETH token, minor Curve fetcher improvement

cehali:


packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -68,10 +68,8 @@ export class CurveFetcher extends DexOnChainFetcher<Response> {
   calculateSpotPrice(assetId: string, response: Response): number {
     const pairedTokenPrice = this.getPairedTokenPrice(assetId);
     const { ratio } = response;
-
-    const spotPrice = ratio.mul(utils.parseEther("1.0")).div(pairedTokenPrice);
-
-    return parseFloat(utils.formatUnits(spotPrice, DEFAULT_DECIMALS));
+    const ratioAsNumber = Number(utils.formatUnits(ratio, DEFAULT_DECIMALS));
+    return ratioAsNumber * pairedTokenPrice;
```
hatskier, line 72: How could it work before? 😄

hatskier: If it passed previous tests, we definitely need to consider implementing better tests and also creating a policy for writing tests, that doesn't allow to write tests only for pairs with stablecoins

packages/oracle-node/manifests/dev/main-wide-support.json
```diff
@@ -826,6 +826,12 @@
     "JGLP": {
       "source": ["camelot-usdc"]
     },
+    "WSTETH": {
+      "source": ["coingecko", "coinmarketcap"]
+    },
+    "STETH": {
+      "source": ["coingecko", "coinmarketcap"]
```
hatskier, line 833: Do we want to add the curve-eth and curve-weth sources here?

hatskier (approved):

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
cehali: Totally agree

PR #540, opened 2023-04-17
chore: update main wide support tokens

cehali:


hatskier:
We would like to only remove lens data feeds, and it's already done in another PR

Issue #548, opened 2023-04-18
Implement same timestamps feature in cache services


koonopek:
We want to implement feature that forces querying all data packages with the same timestamps. It is very useful in on-chain relayer code.

We can move the old approach (which is a bit faster) to a new endpoint, but current endpoint (e.g. https://oracle-gateway-2.a.redstone.finance/data-packages/latest/redstone-avalanche-prod) should return latest data packages with the same timestamps, as it will significantly simplify integration with on-chain-relayers

koonopek:
@hatskier Why not implement new endpoint for this?

koonopek:
Ahh probably caching :)

koonopek:
![Image](https://user-images.githubusercontent.com/24699136/232785897-e0dae1fa-229f-4a00-ba37-57334bf8d067.png)

koonopek:
On production database it is 350 ms (old impl) and 300 ms new impl (50ms faster)

PR #549, opened 2023-04-18
feat: implement same timestamps feature in cache service

koonopek:
Benchmark -  https://github.com/redstone-finance/redstone-oracles-monorepo/issues/548#issuecomment-1513091444

### DIFF
--- a/packages/cache-service/src/data-packages/data-packages.controller.ts
+++ b/packages/cache-service/src/data-packages/data-packages.controller.ts
@@ -16,7 +16,6 @@ import { Cache } from "cache-manager";
 import type { Response } from "express";
 import { Serializable } from "redstone-protocol";
 import { DataPackagesRequestParams } from "redstone-sdk";
-import { BundlrService } from "../bundlr/bundlr.service";
 import config from "../config";
 import { duplexStream } from "../utils/streams";
 import { ReceivedDataPackage } from "./data-packages.interface";
@@ -104,7 +103,7 @@ export class DataPackagesController {
     @Param("DATA_SERVICE_ID") dataServiceId: string
   ): Promise<DataPackagesResponse> {
     await this.validateDataServiceId(dataServiceId);
-    return this.dataPackagesService.getAllLatestDataWithCache(
+    return this.dataPackagesService.getLatestDataPackagesWithCache(
       dataServiceId,
       this.cacheManager
     );
@@ -138,7 +137,7 @@ export class DataPackagesController {
   @Get("latest")
   @Header("Cache-Control", "max-age=5")
   async getLatest(@Query() query: GetLatestDataPackagesQuery) {
-    return await this.dataPackagesService.getDataPackages(
+    return await this.dataPackagesService.queryLatestDataPackages(
       this.prepareDataPackagesRequestParams(query),
       this.cacheManager
     );
--- a/packages/cache-service/src/data-packages/data-packages.controller.ts
+++ b/packages/cache-service/src/data-packages/data-packages.controller.ts
@@ -110,6 +110,18 @@ export class DataPackagesController {
     );
   }
 
+  @Get("most-recent/:DATA_SERVICE_ID")
+  @Header("Cache-Control", "max-age=5")
+  async getMostRecent(
+    @Param("DATA_SERVICE_ID") dataServiceId: string
+  ): Promise<DataPackagesResponse> {
+    await this.validateDataServiceId(dataServiceId);
+    return this.dataPackagesService.getMostRecentDataPackagesWithCache(
+      dataServiceId,
+      this.cacheManager
+    );
+  }
+
   @Get("historical/:DATA_SERVICE_ID/:TIMESTAMP")
   @Header("Cache-Control", "max-age=5")
   async getByTimestamp(
--- a/packages/cache-service/src/data-packages/data-packages.controller.ts
+++ b/packages/cache-service/src/data-packages/data-packages.controller.ts
@@ -103,13 +103,13 @@ export class DataPackagesController {
     @Param("DATA_SERVICE_ID") dataServiceId: string
   ): Promise<DataPackagesResponse> {
     await this.validateDataServiceId(dataServiceId);
-    return this.dataPackagesService.getLatestDataPackagesWithCache(
+    return this.dataPackagesService.getLatestDataPackagesWithSameTimestampWithCache(
       dataServiceId,
       this.cacheManager
     );
   }
 
-  @Get("most-recent/:DATA_SERVICE_ID")
+  @Get("latest-not-aligned-by-time/:DATA_SERVICE_ID")
   @Header("Cache-Control", "max-age=5")
   async getMostRecent(
     @Param("DATA_SERVICE_ID") dataServiceId: string
--- a/packages/cache-service/src/data-packages/data-packages.service.ts
+++ b/packages/cache-service/src/data-packages/data-packages.service.ts
@@ -71,7 +71,7 @@ export class DataPackagesService {
     await DataPackage.insertMany(dataPackages);
   }
 
-  async getAllLatestDataWithCache(
+  async getLatestDataPackagesWithCache(
     dataServiceId: string,
     cacheManager: Cache
   ): Promise<DataPackagesResponse> {
@@ -153,6 +153,7 @@ export class DataPackagesService {
           dataPoints: { $first: "$dataPoints" },
           dataServiceId: { $first: "$dataServiceId" },
           dataFeedId: { $first: "$dataFeedId" },
+          isSignatureValid: { $first: "$isSignatureValid" },
         },
       },
       {
@@ -213,7 +214,7 @@ export class DataPackagesService {
         },
       },
       {
-        $sort: { count: -1, timestamp: -1 },
+        $sort: { count: -1, "_id.timestampMilliseconds": -1 },
       },
       {
         $limit: 1,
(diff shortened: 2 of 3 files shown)


packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -123,35 +123,49 @@ export class DataPackagesService {
...
       },
     ]);
 
     // Parse DB response
-    for (const dataPackage of groupedDataPackages) {
-      const { _id, __v, ...rest } = dataPackage;
-      __v;
-      const dataFeedId = _id.dataFeedId;
+    const dataPackagesWithSameTimestamp = groupedDataPackages[0];
+    for (let i = 0; i < dataPackagesWithSameTimestamp.count; i++) {
```
koonopek, line 147: ugly bust fast ~5ms

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -123,35 +123,49 @@ export class DataPackagesService {
...
+          signatures: { $push: "$signature" },
+          dataPoints: { $push: "$dataPoints" },
+          dataFeedIds: { $push: "$dataFeedId" },
+          signerAddress: { $push: "$signerAddress" },
+          isSignatureValid: { $push: "$isSignatureValid" },
         },
       },
       {
-        $sort: { timestampMilliseconds: -1 },
+        $sort: { count: -1, timestamp: -1 },
```
koonopek, line 138: So we take the "group" which contains the biggest number of packages if two groups contains the same value of packages we take the newer one.

It is an important piece of code. We could also take the newest pacakge if it doesnt deviate more then 2 dataFeeds or some other custom logic @hatskier  what do you think.

I would also want to test it on prod db and compare with production cache services

koonopek:
Sorry for force push, however I fucked my git locally and had to do it

packages/cache-service/src/data-packages/data-packages.service.ts
koonopek: However in the scenario where we have 18 data packages (in 3 MIN interval). And for example, the 18th packages (oldest) contain the most number of dataFeeds, we will select this one and thus serve old data. If losing data feeds between node iterations never occurs, then this is not the issue.

To solve this potential issue we could ALWAYS select the second group (second timestamp). Or the first one if dataFeedId.count is the same in the first and second (all datapacakges for given timestamp are already in the database so the is no need to wait pick the older one).

@hatskier WDYT?

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -104,7 +103,19 @@ export class DataPackagesController {
     @Param("DATA_SERVICE_ID") dataServiceId: string
   ): Promise<DataPackagesResponse> {
     await this.validateDataServiceId(dataServiceId);
-    return this.dataPackagesService.getAllLatestDataWithCache(
+    return this.dataPackagesService.getLatestDataPackagesWithCache(
```
hatskier, line 106: It's not so clear how `Latest` is different from `MostRecent` 😅
Can we rename the function to smth like `getLatestDataPackagesWithSameTimestampWithCache`. I know the name is long, but IMO it will still be better and more readable than now

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -104,7 +103,19 @@ export class DataPackagesController {
...
+      dataServiceId,
+      this.cacheManager
+    );
+  }
+
+  @Get("most-recent/:DATA_SERVICE_ID")
```
hatskier, line 112: Maybe change `most-recent` to smth like `latest-not-aligned-by-time`? WDYT?

packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -126,7 +137,7 @@ export class DataPackagesController {
   @Get("latest")
   @Header("Cache-Control", "max-age=5")
   async getLatest(@Query() query: GetLatestDataPackagesQuery) {
-    return await this.dataPackagesService.getDataPackages(
+    return await this.dataPackagesService.queryLatestDataPackages(
```
hatskier, line 140: Why do you change this endpoint in this PR?

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -71,39 +71,61 @@ export class DataPackagesService {
     await DataPackage.insertMany(dataPackages);
   }
 
-  async getAllLatestDataWithCache(
+  async getLatestDataPackagesWithCache(
     dataServiceId: string,
     cacheManager: Cache
   ): Promise<DataPackagesResponse> {
-    // Checking if data packages for this data service are
```
hatskier, line 78: Just curious, why have you decided to remove this comment?

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -158,14 +180,86 @@ export class DataPackagesService {
...
+          },
+        },
+      },
+      {
+        $group: {
+          _id: {
+            timestampMilliseconds: "$timestampMilliseconds",
+          },
+          count: { $count: {} },
+          dataServiceId: { $first: "$dataServiceId" },
```
hatskier, line 208: Do we really need this line `dataServiceId: { $first: "$dataServiceId" },`? I think we we already know the value we'll receive (because of the $match statement above

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -158,14 +180,86 @@ export class DataPackagesService {
...
+            $gte: Date.now() - MAX_ALLOWED_TIMESTAMP_DELAY,
+          },
+        },
+      },
+      {
+        $group: {
+          _id: {
+            timestampMilliseconds: "$timestampMilliseconds",
+          },
+          count: { $count: {} },
```
hatskier, line 207: Do we really need to count in on the DB level? Can't we do counting and sorting on the JS side?

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -158,14 +180,86 @@ export class DataPackagesService {
...
+          dataServiceId: { $first: "$dataServiceId" },
+          signatures: { $push: "$signature" },
+          dataPoints: { $push: "$dataPoints" },
+          dataFeedIds: { $push: "$dataFeedId" },
+          signerAddress: { $push: "$signerAddress" },
+          isSignatureValid: { $push: "$isSignatureValid" },
+        },
+      },
+      {
+        $sort: { count: -1, "_id.timestampMilliseconds": -1 },
```
hatskier, line 217: Can we do sorting on the JS side to decrease load on DB?

hatskier: I am not sure if it's a better option actually. But would be cool to discuss it

hatskier (changes requested):

packages/cache-service/src/data-packages/data-packages.controller.ts
koonopek: This is the same endpoint as `latest` thus I wanted to be consistent, by this I mean "every GET `/latest` should work the same"

packages/cache-service/src/data-packages/data-packages.service.ts
koonopek: Cause it precisely describes what 2 lines of code are doing, I don't see the added value in this comment.
Also I have introduced a similar function, and repeating this comment would have sens in my opinion so I have deleted it in two places.

packages/cache-service/src/data-packages/data-packages.service.ts
koonopek: Good point!

packages/cache-service/src/data-packages/data-packages.service.ts
koonopek: This is always an option :) But lets discuss:

The cons:
1. DBs are optimized for this kind of operation (use better runtime then node).
2. They are easier to scale than services
3. Reduce the amount of data that has to be sent over the wire by 18 times in this case
4. Increase load on service (also this is blocking operation so when server will be manipulating data he won't be able to )
The pros:
1.  Decrease load on DB.

When I have disabled counting sorting and limit, the query executes in ~380ms so is almost 2 times slower (200ms). Probably cause of the data volume send. (tested on WIFI in our biuro)

packages/cache-service/src/data-packages/data-packages.service.ts
koonopek: Answered here: https://github.com/redstone-finance/redstone-oracles-monorepo/pull/549#discussion_r1172162664

packages/cache-service/src/data-packages/data-packages.controller.ts
koonopek: Totally, I was having some hard time with this names :)

packages/cache-service/src/data-packages/data-packages.controller.ts
koonopek: Agree

packages/cache-service/src/data-packages/data-packages.service.ts
hatskier: I am convinced, didn't take into account the 3rd point before, but now noticed the limit: 1. So it makes much more sense

packages/cache-service/src/data-packages/data-packages.service.ts
hatskier: I think it will never be the case that 18th timestamp has the biggest number of packages

PR #550, opened 2023-04-18
feat: prepare node for price limits

cehali:


packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -282,6 +298,12 @@ export default class PricesService {
 
       if (deviationPercent > deviationWithRecentValues.maxPercent) {
         reason = `Value is too deviated (${deviationPercent}%)`;
+      } else if (
+        priceLimits &&
+        (value >= priceLimits.upper || value <= priceLimits.lower)
```
hatskier, line 303: Maybe > and < instead of >= and <=? IMO it's more clear if the limit is allowed to be reached

packages/oracle-node/test/PricesService.spec.ts
```diff
@@ -372,12 +375,43 @@ describe("PricesService", () => {
...
-          emptyManifest.deviationCheck
+          emptyManifest.deviationCheck,
+          {}
         )
       ).toThrow(
         "Invalid price for symbol ETH. Reason: Value is too deviated (114.28571428571428%)"
       );
     });
+
+    it("should throw when price out of lower limit", async () => {
```
hatskier, line 386: Let's also add price hard limits check in the integration tests

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -173,6 +185,7 @@ export default class PricesService {
     prices: PriceDataBeforeAggregation[]
   ): Promise<PriceDataAfterAggregation[]> {
```
hatskier, line 186: Please update the comment to this function `calculateAggregatedValues`, as not it's also based on hard limits

hatskier (approved):

PR #563, opened 2023-04-20
feat: add check for last price timestamp

cehali:


packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -119,8 +121,22 @@ const setLastPrices = (prices: PriceDataAfterAggregation[]) => {
...
+export const getLastPrice = (
+  symbol: string
+): PriceValueInLocalDB | undefined => {
+  const currentTimestamp = Date.now();
+  const lastPrice = lastPrices[symbol];
+
+  if (lastPrice) {
+    const timeDiff = currentTimestamp - lastPrice.timestamp;
+    if (timeDiff >= MAX_PRICE_IN_DB_TIME_DIFF) {
+      throw new Error(
```
hatskier, line 133: Or maybe we should return undefined in this case? WDYT?

cehali: I think if we have a price in the local cache for a particular token and it is obsolete it means something is not right and we should give information about it. We could assume that it would come up somewhere. Let's discuss it tomorrow.

hatskier: Ok, it makes sense

hatskier (approved):

PR #565, opened 2023-04-21
feat: add tradfi indexes

cehali:


packages/oracle-node/src/fetchers/twelve-data-etf/TwelveDataEtfFetcher.ts
```diff
@@ -0,0 +1,53 @@
+import _ from "lodash";
+import axios, { AxiosResponse } from "axios";
+import { config } from "../../config";
+import {
+  TWELVE_DATA_PRICE_URL,
+  TwelveDataFetcher,
+  TwelveDataResponse,
+} from "../twelve-data/TwelveDataFetcher";
+import { getRequiredPropValue } from "../../utils/objects";
```
hatskier, line 9: I think we can abstract much more common logic between TwelveDataEtfFetcher and TwelveDataFetcher, to only specify the differences (.e.g we can have a TwelveDataFetcherBase) and only set the params that are different

packages/oracle-node/src/fetchers/twelve-data-etf/TwelveDataEtfFetcher.ts
```diff
@@ -0,0 +1,53 @@
...
+interface Tokens {
+  [symbol: string]: string;
+}
+
+interface Config {
+  country: string;
+  tokens: Tokens;
+}
+
+export class TwelveDataEtfFetcher extends TwelveDataFetcher {
```
hatskier, line 20: I think we should add tests for this fetcher

packages/oracle-node/src/fetchers/twelve-data-etf/TwelveDataEtfFetcher.ts
```diff
@@ -0,0 +1,53 @@
...
+import axios, { AxiosResponse } from "axios";
+import { config } from "../../config";
+import {
+  TWELVE_DATA_PRICE_URL,
+  TwelveDataFetcher,
+  TwelveDataResponse,
+} from "../twelve-data/TwelveDataFetcher";
+import { getRequiredPropValue } from "../../utils/objects";
+
+interface Tokens {
```
hatskier, line 11: IMHO `Tokens` sounds too generic

hatskier (changes requested):

hatskier (approved):
LGTM!

PR #574, opened 2023-04-24
feat: make more params optional in data-service-wrapper

koonopek:


packages/protocol/src/data-services-urls.ts
```diff
@@ -0,0 +1,25 @@
...
+    "https://oracle-gateway-1.a.redstone.finance",
+    "https://oracle-gateway-2.a.redstone.finance",
+  ],
+  "redstone-custom-urls-demo": ["https://d1zm8lxy9v2ddd.cloudfront.net"],
+  "redstone-main-demo": [DEMO_URL],
+  "redstone-rapid-demo": [DEMO_URL],
+  "redstone-stocks-demo": [DEMO_URL],
+  "redstone-twaps-demo": [DEMO_URL],
+  "mock-data-service": ["http://valid-cache.com"],
+};
```
koonopek, line 14: All examples, however KydService was missing in examples

packages/evm-connector/contracts/data-services/RedstoneDataServiceConsumer.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity ^0.8.4;
+
+import "../core/RedstoneConsumerNumericBase.sol";
+
+abstract contract RedstoneDataServiceConsumer is RedstoneConsumerNumericBase {
```
koonopek, line 7: New abstract class for dataServices

cehali: We don't need to new abstract class to add one minor function. Maybe we could add it to `RedstoneConsumerBase`. We don't plan not to use data service id anywhere.

packages/evm-connector/src/wrappers/BaseWrapper.ts
```diff
@@ -14,11 +14,14 @@ interface OverwriteFunctionArgs {
...
 
 export abstract class BaseWrapper {
+  protected contract!: Contract;
+
   abstract getBytesDataForAppending(
     params?: ParamsForDryRunVerification
   ): Promise<string>;
 
   overwriteEthersContract(contract: Contract): Contract {
+    this.contract = contract;
```
cehali, line 24: Is it used somewhere outside this function?

koonopek: yes

koonopek: To fetch dataServiceId from this.contract

packages/evm-connector/contracts/data-services/RedstoneDataServiceConsumer.sol
koonopek: Hmm I think that it could make confusion to developers which will try extend `RedstoneConsumerNumericBase` because they will have to override this method and it could be problematic for them to decide which value to use.
However if you think that this is not an issue I can add it to RedstoneConsumerBase

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -6,17 +6,54 @@ import { BaseWrapper, ParamsForDryRunVerification } from "./BaseWrapper";
 import { parseAggregatedErrors } from "../helpers/parse-aggregated-errors";
 import { runDryRun } from "../helpers/run-dry-run";
 import { version } from "../../package.json";
+import { resolveDataServiceUrls } from "redstone-protocol";
 
-interface DryRunParamsWithUnsignedMetadata extends ParamsForDryRunVerification {
+type Optional<T, K extends keyof T> = Omit<T, K> & {
```
cehali, line 11: Maybe use `Partial` type: https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype?

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -6,17 +6,54 @@ import { BaseWrapper, ParamsForDryRunVerification } from "./BaseWrapper";
...
+    "dataServiceId"
+  >;
+
   constructor(
-    private dataPackagesRequestParams: DataPackagesRequestParams,
-    private urls: string[]
+    dataPackagesRequestParams: Optional<
+      DataPackagesRequestParams,
+      "dataServiceId" | "uniqueSignersCount"
+    >,
```
cehali, line 38: Why not `DataPackagesRequestInput`?

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -6,17 +6,54 @@ import { BaseWrapper, ParamsForDryRunVerification } from "./BaseWrapper";
...
 }
 
+export type DataPackagesRequestInput = Optional<
+  DataPackagesRequestParams,
+  "dataServiceId" | "uniqueSignersCount"
+>;
+
 export class DataServiceWrapper extends BaseWrapper {
+  private _urls?: string[];
+  private readonly dataPackagesRequestParams: Optional<
```
cehali, line 29: Why it is different than `dataPackagesRequestParams` and `DataPackagesRequestInput`?

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -6,17 +6,54 @@ import { BaseWrapper, ParamsForDryRunVerification } from "./BaseWrapper";
...
+    DataPackagesRequestParams,
+    "dataServiceId"
+  >;
```
cehali, line 32: Why it is different than `dataPackagesRequestParams` and `DataPackagesRequestInput`?

cehali: Ohhh, I see

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -6,17 +6,54 @@ import { BaseWrapper, ParamsForDryRunVerification } from "./BaseWrapper";
...
 import { runDryRun } from "../helpers/run-dry-run";
 import { version } from "../../package.json";
+import { resolveDataServiceUrls } from "redstone-protocol";
 
-interface DryRunParamsWithUnsignedMetadata extends ParamsForDryRunVerification {
+type Optional<T, K extends keyof T> = Omit<T, K> & {
+  [P in K]?: T[P];
+};
+
+const DEFAULT_UNIQUE_SIGNERS_COUNT = 2;
```
cehali, line 15: I am not sure 2 is the best choice here. In the prod env we have 3 out of 5 nodes and in dev we have only 1 node per data service.

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -6,17 +6,54 @@ import { BaseWrapper, ParamsForDryRunVerification } from "./BaseWrapper";
...
+      ...dataPackagesRequestParams,
+    };
+  }
+
+  private get urls(): string[] {
+    if (this._urls) {
+      return this._urls;
+    }
+    this._urls = resolveDataServiceUrls(
+      this.dataPackagesRequestParams.dataServiceId as string
```
cehali, line 54: I would leave it as the default type and add undefined as possible type to `resolveDataServiceUrls`. There you have to check to throw an error if we don't find the URL.

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
koonopek: Partial make all Properties optional, `Optional` make only given properties Optional

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
koonopek: I would leave it as is, it is more type safe (better to catch something in compiler error then in runtime or tests)
In this case, the developer must explicitly say that something is a string. 
And `undefined|null` is not valid input to `resolveDataServicesUrls` of course, if it is used only in this place this is not a difference, but this function can be also used somewhere else.

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
koonopek: Ohh but you are right we are actually making all fields optional here, good point, will fix

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
koonopek: Yeah, good point! Will do

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
cehali: iIf we call `urls()` before `getUnsignedMetadata()`  `dataServiceId` would be undefined. In theory it shouldn't happen but could be misleading in the future. Maybe we should store URLs as class variable and just use it in `getUnsignedMetadata` without storing it? `resolveDataServiceUrls` is just getting from the object so we could run it every time

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -6,17 +6,41 @@ import { BaseWrapper, ParamsForDryRunVerification } from "./BaseWrapper";
...
   constructor(
-    private dataPackagesRequestParams: DataPackagesRequestParams,
-    private urls: string[]
+    dataPackagesRequestParams: DataPackagesRequestInput,
+    urls?: string[]
   ) {
     super();
+    this._urls = urls;
+    this.dataPackagesRequestParams = {
+      uniqueSignersCount: DEFAULT_UNIQUE_SIGNERS_COUNT,
```
cehali, line 31: I think we can remove `DEFAULT_UNIQUE_SIGNERS_COUNT` completely and handle it the same as dataServiceId, I believe it is already handled that way.

koonopek: true

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
koonopek: You mean `getBytesDataForAppending()` not `getUnsignedMetadata()`?

cehali: yes `getBytesDataForAppending`

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -31,10 +36,38 @@ export class DataServiceWrapper extends BaseWrapper {
     const disablePayloadsDryRun = Boolean(
       this.dataPackagesRequestParams.disablePayloadsDryRun
     );
+
+    const { urls } = await this.loadDefaultParams();
```
koonopek, line 40: Hack for typscript cause:
`When you check or assert the type of some value so as to narrow its apparent type, the compiler uses control flow analysis to keep track of where and when such a narrowing should apply. But this analysis only works inside a single function scope. It does not cross function boundaries.`
Thanks @cehali for reference

packages/evm-connector/contracts/data-services/CustomUrlsConsumerBase.sol
```diff
@@ -5,17 +5,17 @@ pragma solidity ^0.8.4;
...
   }
 
-  function getAuthorisedSignerIndex(address signerAddress)
-    public
-    view
-    virtual
-    override
-    returns (uint8)
-  {
+  function getAuthorisedSignerIndex(
```
hatskier, line 16: Please avoid doing formatting changes in contracts

packages/evm-connector/contracts/data-services/KydServiceConsumerBase.sol
```diff
@@ -8,17 +8,17 @@ contract KydServiceConsumerBase is RedstoneConsumerNumericBase {
...
   }
 
-  function getAuthorisedSignerIndex(address signerAddress)
-    public
-    view
-    virtual
-    override
-    returns (uint8)
-  {
+  function getAuthorisedSignerIndex(
```
hatskier, line 19: Please don't modify formatting

packages/protocol/src/data-services-urls.ts
```diff
@@ -0,0 +1,25 @@
+const DEMO_URL = "https://d33trozg86ya9x.cloudfront.net";
+
+export const REDSTONE_DATA_SERVICES_URLS: Record<string, string[]> = {
```
hatskier, line 3: IMO these URLs should be placed in sdk, not in the protocol

packages/evm-connector/src/wrappers/BaseWrapper.ts
```diff
@@ -14,11 +14,14 @@ interface OverwriteFunctionArgs {
 }
 
 export abstract class BaseWrapper {
+  protected contract!: Contract;
```
hatskier, line 17: Why exclamation mark?

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -6,15 +6,20 @@ import { BaseWrapper, ParamsForDryRunVerification } from "./BaseWrapper";
...
 }
 
+export type DataPackagesRequestInput = Partial<DataPackagesRequestParams>;
+
 export class DataServiceWrapper extends BaseWrapper {
   constructor(
-    private dataPackagesRequestParams: DataPackagesRequestParams,
-    private urls: string[]
+    private readonly dataPackagesRequestParams: DataPackagesRequestInput,
+    private urls?: string[]
```
hatskier, line 22: Just an idea: WDYT about keeping urls also in DataPackagesRequestParams to have only one object with optional and non-optional params?

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -31,10 +36,38 @@ export class DataServiceWrapper extends BaseWrapper {
...
+        await this.getDataServiceIdFromContract();
+    }
+
+    if (!this.urls) {
+      this.urls = resolveDataServiceUrls(
+        this.dataPackagesRequestParams.dataServiceId
+      );
+    }
+
+    return { ...this.dataPackagesRequestParams, urls: this.urls };
```
hatskier, line 70: Why do we return the object if we only use urls?

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -31,10 +36,38 @@ export class DataServiceWrapper extends BaseWrapper {
...
+  }
+
+  private async loadDefaultParams() {
+    if (!this.dataPackagesRequestParams.uniqueSignersCount) {
+      this.dataPackagesRequestParams.uniqueSignersCount =
+        await this.getUniqueSignersThresholdFromContract();
+    }
+
+    if (!this.dataPackagesRequestParams.dataServiceId) {
+      this.dataPackagesRequestParams.dataServiceId =
```
hatskier, line 60: IMHO it's not very readable when function for loading params also produces side effects (like updating few props of `dataPackagesRequestParams`).

Maybe we can rename it so it indicates what it does (e.g. `setDefaultParamsIfNeeded`)

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -67,9 +101,36 @@ export class DataServiceWrapper extends BaseWrapper {
...
     return requestRedstonePayload(
-      this.dataPackagesRequestParams,
+      this.dataPackagesRequestParams as DataPackagesRequestParams,
       urls,
       unsignedMetadataMsg
     );
   }
+
+  private async getDataServiceIdFromContract(): Promise<string> {
+    try {
```
hatskier, line 111: I would propose to have only one `try .. catch`, e.g. in the `loadDefaultParams` and here I would implement only "unsafe" logic of fetching details from the contract

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -21,6 +21,16 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
...
   /* ========== VIRTUAL FUNCTIONS (MAY BE OVERRIDDEN IN CHILD CONTRACTS) ========== */
 
+  /**
+   * @dev This function must be implemented by the child consumer contract.
+   * It should return dataServiceId which DataServiceWrapper will use if not provided explicite .
+   * If not overriden, value will always have to be provided explicite in DataServiceWrapper.
+   * @return dataServiceId being consumed by contract
+   */
+  function getDataServiceId() public view virtual returns (string memory) {
+    return "";
```
hatskier, line 31: I would propose to revert here with an error, as we consider "" as invalid value, so maybe we can simply throw the error here

hatskier (changes requested):

packages/protocol/src/data-services-urls.ts
koonopek: image
:)

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
koonopek: Good idea done

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
koonopek: This is also used in differnet places where urls is not required, however I can add it to urls

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
koonopek: Then we miss more precise error message

koonopek: We don't know exactly which param we cant fetch

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
koonopek: Valid point, refactored

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
koonopek: I have refactored this function

packages/evm-connector/src/wrappers/BaseWrapper.ts
koonopek: To tell typescript compiler that this function will be for sure initialized (but outside of constructor)

packages/protocol/src/data-services-urls.ts
koonopek: But I also think that it more fits in SDK :)

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
koonopek: Changed my mind and done as you suggested

packages/protocol/src/data-services-urls.ts
koonopek: moved

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
hatskier: I think we can remove the case with checking if the data service id has been overriden if we revert with error if it's not. I think we can still throw an error inside these functions, but have only a single try .. catch block in the function that uses all of them. Let's discuss it in person :)

packages/evm-connector/contracts/core/RedstoneConsumerBase.sol
```diff
@@ -21,6 +21,16 @@ abstract contract RedstoneConsumerBase is CalldataExtractor {
...
   /* ========== VIRTUAL FUNCTIONS (MAY BE OVERRIDDEN IN CHILD CONTRACTS) ========== */
 
+  /**
+   * @dev This function must be implemented by the child consumer contract.
+   * It should return dataServiceId which DataServiceWrapper will use if not provided explicite .
+   * If not overriden, value will always have to be provided explicite in DataServiceWrapper.
+   * @return dataServiceId being consumed by contract
+   */
+  function getDataServiceId() public view virtual returns (string memory) {
+    revert("getDataServiceId not implemented");
```
hatskier, line 31: Let's use custom errors to be consistent with the rest of the code

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -25,16 +30,52 @@ export class DataServiceWrapper extends BaseWrapper {
...
+      unsignedMetadataMsg,
+      dataPackagesRequestParams,
+      ...dryRunParams,
+    });
+  }
+
+  private async resolveDataPackagesRequestParams(): Promise<
+    Required<DataPackagesRequestInput>
+  > {
+    const fetchedParams = {} as Required<DataPackagesRequestInput>;
```
hatskier, line 60: Why do you set it as `Required<DataPackagesRequestInput>`, despite it can not have this type?

packages/sdk/src/data-services-urls.ts
```diff
@@ -0,0 +1,25 @@
...
+  ],
+  "redstone-custom-urls-demo": ["https://d1zm8lxy9v2ddd.cloudfront.net"],
+  "redstone-main-demo": [DEMO_URL],
+  "redstone-rapid-demo": [DEMO_URL],
+  "redstone-stocks-demo": [DEMO_URL],
+  "redstone-twaps-demo": [DEMO_URL],
+  "mock-data-service": ["http://valid-cache.com"],
+};
+
+export function resolveDataServiceUrls(dataServiceId: string): string[] {
```
hatskier, line 16: Why `function` notation, not the fancy one (with arrows)?

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
hatskier: Okay, I see the value of more etailed error messages for end developers

hatskier (approved):

PR #588, opened 2023-05-02
feat: improved sources for steth

hatskier:
- Added `kaiko` and `uniswap-v2-on-chain-weth` sources for stETH
- Removed coingecko source for stETH
- Increased `interval` in the `kaiko` fetcher as it didn't work for stETH with smaller interval


cehali (approved):
LGTM

packages/oracle-node/manifests/data-services/kaiko.json
```diff
@@ -13,13 +13,13 @@
     "ETH": {},
```
koonopek, line 13: I'm not sure if this is related to the current PR, but I noticed that there are two instances of kaiko.json - one here and another in single-sources. It appears to be redundant.

packages/oracle-node/src/fetchers/kaiko/KaikoFetcher.ts
```diff
@@ -13,14 +13,10 @@ const KAIKO_REQUEST_HEADERS = {
 };
 const KAIKO_REQUEST_PARAMS = {
   page_size: 10,
-  interval: "1m",
+  interval: "5m",
```
koonopek, line 16: Or we sure? I think that we should prefer the lowest possible interval

koonopek:
Dry run still failing (VST token is missing)

packages/oracle-node/src/fetchers/kaiko/KaikoFetcher.ts
koonopek: Or maybe not, cause it would be more fragile to manipulation, however still interested why are u changing this?

hatskier: I've answered in the description to this PR ;)

PR #592, opened 2023-05-03
integration tests classic model

rpiszczatowski:


.github/workflows/monorepo-integration-tests.yml
```diff
@@ -12,7 +12,18 @@ jobs:
         working-directory: packages/integration
     steps:
       - uses: actions/checkout@v3
-      - name: Run tests
+      - name: Run core tests
         run: bash integration-test-core.sh
         env:
           NODE_ENV: test
+  classic:
```
hatskier, line 19: Would the tests run in parallel? If not, I would propose to separate these 2 files

hatskier: This looks a bit misleading

Screenshot 2023-05-03 at 15 45 19

packages/integration/integration-test-classic.sh
```diff
@@ -0,0 +1,174 @@
...
+  [ -n "$onChainRelayerPid" ] && kill $onChainRelayerPid
+  [ -n "$hardhatNodePid" ] && kill $hardhatNodePid
+  [ -n "$oracleNodePid" ] && kill $oracleNodePid
+  [ -n "$cacheLayerPid" ] && kill $cacheLayerPid
+  [ -n "$mongoDbPid" ] && kill $mongoDbPid
+}
+
+trap 'stopAll' EXIT
+
+lazilyInstallNPMDeps() {
```
hatskier, line 27: What do you think about moving duplicated functions to a separate bash file and includng it using the `source` command on top?

packages/on-chain-relayer/.gitignore
```diff
@@ -13,3 +13,6 @@ artifacts
 .openzeppelin/unknown-*.json
 
 hardhat-deployed-contracts.json
+
+# integration tests temporary files
+adapter_contract_address.txt
```
hatskier, line 18: Can you rename to `adapter-contract-address` and use dashes instead of underscores to be consistent with other filenames?

packages/on-chain-relayer/contracts/mocks/PriceFeedsAdapterWithoutRoundsMock.sol
```diff
@@ -7,12 +7,14 @@ import "../price-feeds/without-rounds/PriceFeedsAdapterWithoutRounds.sol";
...
-    dataFeedIds = new bytes32[](1);
+    dataFeedIds = new bytes32[](3);
     dataFeedIds[0] = bytes32("BTC");
+    dataFeedIds[1] = bytes32("ETH");
+    dataFeedIds[2] = bytes32("AAVE");
   }
 
   function getUniqueSignersThreshold() public view virtual override returns (uint8) {
-    return 2;
+    return 1;
```
hatskier, line 17: Let's please create a separate contract file for the integration test in the same way we do it in evm-connector and core tests

packages/on-chain-relayer/test/monorepo-integration-tests/deploy-adapter.ts
```diff
@@ -0,0 +1,10 @@
+import { ethers } from "hardhat";
```
hatskier, line 1: Let's use `scripts` folder for this file as it's a hardhat script,  not a test

packages/on-chain-relayer/test/monorepo-integration-tests/verify-prices.ts
```diff
@@ -0,0 +1,24 @@
+import { expect } from "chai";
```
hatskier, line 1: I would also make a hardhat script from this file as it's not a normal isolated test, because there is an interaction with already deployed contract. We can still use `expect` in a hardhat script

hatskier (changes requested):

.github/workflows/monorepo-integration-tests.yml
rpiszczatowski: Separate jobs by default run in parallel, so yes.
I'll adjust the naming.

packages/integration/integration-test-classic.sh
```diff
@@ -0,0 +1,174 @@
+#!/bin/bash
```
koonopek, line 1: Sorry for the interrupt, but why we are not writing this kind of tests in Node.js.
I think that in long term it will be hard to maintain them, and oh boy we want to maintain them :)

rpiszczatowski: This is a rather philosophical question :) The main argument is that the primary job done by the script is to start all the components in the background, schedule some 'standard' tasks they perform and then run some verifications on the state of those components and then clean up. As far as I am concerned, the best tools for such a work are scripting languages. Also, please note that for tasks strictly related to interacting with our components I used hardhat, and the code is written in typescript (deploying contract, on-chain prices verification). If we have more interactions with the system they will be written in TS and only started from this shell script.
And the other argument is that tests for the core component were already written in bash :)

Having said that, I am open to switching to Node.js even though I am less experienced with this tool compared to bash.

koonopek: What is a script is rather a philosophical question :) java__SCRIPT__ :) (JOKING)

In my opinion, maintaining a bash code base is more difficult compared to TS. When using oracle-ci-node, we encountered problems with processes that were not being terminated, a situation that would have been easier to catch in JS. Additionally, adding new assertions would be simpler using JS. ( I think I'm repeating myself sorry)

(PS. I use JS TS interchangeably)

rpiszczatowski: resolving as non-conclusive at the moment. We can discuss it further in a broader group.

koonopek: @hatskier @cehali @Lukasz2891 what do you think?

cehali: I remember that Alex had some good reasons why he didn't want to do this in node.js but I am not sure what was it 😄

Lukasz2891: The question #0 - why don't we use a prepared `.env` file instead of copying `.env.example` and setting then the variables in the script? 

For example `.env.on-chain-relayer` with filled values could be attached to the invocation of the node/yarn script, I think ;) Couldn't it? It sth related to `NODE_ENV` variable, but I'm not sure it completely transparent or it's needed to have a piece of code in a `.ts' file to read it.

Lukasz2891: Shell is nothing special for me. I use `Makefile`s :D The main advantage of using shell is that we don't need `node` installed on a machine the tests are running on. But we do it for running the tests anyway...

So for me, for now, the argument "we have sth working" is totally ok too ;)

packages/integration/integration-test-classic.sh
```diff
@@ -0,0 +1,114 @@
...
+  # Compile on-chain-relayer
+  cd ../on-chain-relayer
+  lazilyInstallNPMDeps
+  yarn compile
+  lazilyBuildTypescript
+
+  # Launch local EVM instance with hardhat
+  runWithLogPrefixInBackground "yarn start-node" "hardhat-node"
+  hardhatNodePid=$!
+  sleep 10 # wait for hardhat to start blockchain instance
```
hatskier, line 85: I would propose to implement some kind of active waiting, because it's possible that hardhat will not launch a local blockchain instance on a slower machine and the tests will fail in a non-deterministic way

packages/integration/integration-test-classic.sh
```diff
@@ -0,0 +1,114 @@
...
+  updateDotEnvFile "PRIVATE_KEY" "$HARDHAT_MOCK_PRIVATE_KEY"
+  updateDotEnvFile "ADAPTER_CONTRACT_ADDRESS" "$ADAPTER_CONTRACT_ADDRESS"
+  updateDotEnvFile "DATA_FEEDS" '["BTC", "ETH", "AAVE"]'
+  updateDotEnvFile "DATA_SERVICE_ID" "mock-data-service"
+  updateDotEnvFile "CACHE_SERVICE_URLS" '["http://localhost:3000"]'
+  updateDotEnvFile "HEALTHCHECK_PING_URL" ""
+  cat .env
+  runWithLogPrefixInBackground "yarn start" "on-chain-relayer"
+  onChainRelayerPid=$!
+  sleep 15 # wait for relayer to put data on-chain
```
hatskier, line 103: Same, I would propose to avoid sleep commands with hardcoded time

packages/on-chain-relayer/test/monorepo-integration-tests/scripts/deploy-adapter.ts
```diff
@@ -0,0 +1,10 @@
+import { ethers } from "hardhat";
+import fs from "fs";
```
hatskier, line 2: Can we rename this file so it indicates that it deploys mock adapter for integration tests

packages/on-chain-relayer/test/monorepo-integration-tests/scripts/verify-prices.ts
```diff
@@ -0,0 +1,24 @@
+import { expect } from "chai";
+import { BigNumber } from "ethers";
+import { formatBytes32String } from "ethers/lib/utils";
+import { ethers } from "hardhat";
```
hatskier, line 4: Can you please rename this script so it indicates that it is used in monorepo integration tests. You can also add few details in the comments inside the file

packages/on-chain-relayer/test/monorepo-integration-tests/scripts/verify-prices.ts
```diff
@@ -0,0 +1,24 @@
...
+import { PriceFeedsAdapterWithoutRounds } from "../../../typechain-types";
+
+(async () => {
+  const ContractFactory = await ethers.getContractFactory(
+    "PriceFeedsAdapterWithoutRoundsOneSignerMock"
+  );
+  let contract: PriceFeedsAdapterWithoutRounds = ContractFactory.attach(
```
hatskier, line 11: Can we change `let` -> `const`?

hatskier (changes requested):

packages/on-chain-relayer/test/monorepo-integration-tests/scripts/deploy-adapter.ts
rpiszczatowski: I added word 'mock'. But 'integration test' part is already in folder structure. Do you still want to repeat it in file name?
@hatskier ?

packages/on-chain-relayer/test/monorepo-integration-tests/scripts/verify-prices.ts
rpiszczatowski: same question as above regarding file name: isn't it enough to have it in the folder "monorepo-integration-tests"?
@hatskier ?

packages/cache-service/scripts/analyze-data-packages.ts
```diff
@@ -57,6 +57,7 @@ interface DataPackagesGroupedBySigner {
 main();
 
 async function main() {
+  mongoose.set('strictQuery', true);
```
hatskier, line 60: Why do you add it in few places? Can we do it in one place?

hatskier (approved):

packages/cache-service/scripts/analyze-data-packages.ts
rpiszczatowski: Hmm. I see no common code where we could put it. I don't think it is worth complicating those scripts by creating some common one just yet. Especially because the explicit setting of the strictQuery parameter is rather temporary. This is to make the code compatible with the next mongoose version where the default will change to true.

packages/on-chain-relayer/test/monorepo-integration-tests/scripts/deploy-adapter.ts
hatskier: You are right

hatskier (approved):

PR #596, opened 2023-05-04
fix: node version in CI

cehali:


cehali:
Based on https://github.com/NomicFoundation/hardhat/issues/3877

.github/workflows/sdk-tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       run:
         working-directory: packages/sdk
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
koonopek, line 17: Missing `actions/setup-node`?

koonopek:
So some specific node version was the issue, or the issue was that is wasn't specified?

koonopek (approved):
One minior question

.github/workflows/sdk-tests.yml
cehali: We need to specify it only in packages where we use hardhat

cehali:
> So some specific node version was the issue, or the issue was that is wasn't specified?

It seems that 18.16 is breaking something but I assumed that when we use `actions/checkout@v2` we use an older version than 18 because we had warnings that we should switch to v3 to use a newer node.

hatskier (approved):

.github/workflows/sdk-tests.yml
koonopek: I think that this is good practice to specify it everywhere to be sure what node are we using

PR #599, opened 2023-05-04
chore: update and unify dependencies

cehali:


cehali:
Based on https://github.com/NomicFoundation/hardhat/issues/3877

koonopek (approved):
Nice!

hatskier (commented):
Why does it have so many changes that are not related to deps?

cehali:
> Why does it have so many changes that are not related to deps?

@hatskier Updated dependencies changed their interfaces or removed some functions and I had to resolve those issues. The only change is that waffle within hardhat doesn't allow passing parameters in the `revertedWith` function. I added `.withArgs` to make it work.

PR #578, opened 2023-04-25
feat: add dry run test for stocks manifest

cehali:

PR #601, opened 2023-05-05
chore: clean manifests for monitoring

cehali:


packages/oracle-node/src/fetchers/ccxt/symbol-to-id/binancecoinm.json
```diff
@@ -1,43 +1,43 @@
...
-  "ICX": "ICX/USD",
-  "ALGO": "ALGO/USD",
-  "RUNE": "RUNE/USD",
-  "APE": "APE/USD",
-  "VET": "VET/USD",
-  "ZIL": "ZIL/USD",
-  "KNC": "KNC/USD",
-  "XMR": "XMR/USD",
-  "GMT": "GMT/USD"
+  "BTC": "BTC/USD:BTC",
```
koonopek, line 2: I would appreciate it if you could provide me with an explanation of this change so that I can better understand it.

packages/oracle-node/src/fetchers/ccxt/CcxtFetcher.ts
```diff
@@ -81,7 +81,11 @@ export class CcxtFetcher extends BaseFetcher {
 
   serializePairSymbol(pairSymbol: string) {
     if (pairSymbol.endsWith("/USDT:USDT")) {
-      return pairSymbol.replace("/USDT:USDT", "USDT");
+      if (this.name === "bybit") {
```
koonopek, line 84: Do you think we could provide a comment explaining what is happening here, looks pretty magical?

koonopek:
What about:
binanceusdm 
liquid 
probit 
from https://gist.github.com/koonopek/5210470143aa6a7b6bf3d733cfb34932

packages/oracle-node/src/fetchers/ccxt/symbol-to-id/binancecoinm.json
cehali: `binancecoinm` fetcher from `ccxt` returns for token `ETH` object like below. So you call function `fetchTickers(["ETH"])` and receive an object with key and symbol value `ETH/USD:ETH`. Probably a newer version of `ccxt` changed it.
image

packages/oracle-node/src/fetchers/ccxt/CcxtFetcher.ts
cehali: It turns out it is no longer needed, removed `serializePairSymbol`

packages/oracle-node/manifests/data-services/stocks.json
```diff
@@ -19,7 +19,6 @@
         "binanceus",
```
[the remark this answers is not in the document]
cehali: It is in this PR: https://github.com/redstone-finance/redstone-oracles-monorepo/pull/578

packages/oracle-node/src/fetchers/ccxt/symbol-to-id/binancecoinm.json
koonopek: Wired stuff, thanks for explaining this

packages/oracle-node/manifests/single-source/bybit-v5.json
```diff
@@ -0,0 +1,15 @@
+{
+  "interval": 60000,
+  "priceAggregator": "median",
+  "defaultSource": ["bybit"],
```
koonopek, line 4: "bybit-v5"?

packages/oracle-node/manifests/data-services/arbitrum-vesta.json
```diff
@@ -35,7 +34,6 @@
     "USDT": {
       "source": [
         "binanceus",
-        "bitfinex",
```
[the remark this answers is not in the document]
koonopek: nvm :)

packages/oracle-node/src/fetchers/index.ts
```diff
@@ -74,6 +75,7 @@ export default {
   newyorkfed: new NewyorkfedFetcher(),
   "arbitrum-evm-fetcher": new ArbitrumEvmFetcher(arbitrumProvider),
   "non-usd-based": new NonUsdBasedFetcher(),
+  "bybit-v5": new CcxtFetcher("bybit"),
```
hatskier, line 78: Why?

packages/oracle-node/manifests/single-source/idex.json
```diff
@@ -9,7 +9,5 @@
       "maxDelayMilliseconds": 300000
     }
   },
-  "tokens": {
-    "DAI": {}
-  }
+  "tokens": {}
```
hatskier, line 12: What's the need of this manifest if it has no tokens?

packages/oracle-node/src/fetchers/ccxt/CcxtFetcher.ts
```diff
@@ -78,11 +77,4 @@ export class CcxtFetcher extends BaseFetcher {
     const stableCoinSymbol = pairSymbol.slice(-4);
     return getLastPrice(stableCoinSymbol)?.value;
   }
-
-  serializePairSymbol(pairSymbol: string) {
```
hatskier, line 82: Why do you remove this function?

hatskier (changes requested):

packages/oracle-node/manifests/single-source/bybit-v5.json
cehali: If we have in the same manifest token `CMP` which only works with bybit v5 and other tokens like `AVAX` which works with older API versions, then we don't get `CMP`. This is why we need a separate manifest for `CMP`, I hope they will fix it in the future.

packages/oracle-node/src/fetchers/index.ts
cehali: If we have in the same manifest token `CMP` which only works with bybit v5 and other tokens like `AVAX` which works with older API versions, then we don't get `CMP`. This is why we need a separate manifest for `CMP`, I hope they will fix it in the future.

packages/oracle-node/src/fetchers/ccxt/CcxtFetcher.ts
cehali: I changed files with symbol-to-id in a way that we don't need this function. It was only used in bybit and binanceusdm.

packages/oracle-node/manifests/single-source/idex.json
cehali: Deleted

PR #604, opened 2023-05-05
feat: add relayer addresses to deployed config

koonopek:


hatskier (commented):
@koonopek WDYT about keeping these addresses in our remote config? I also think there will be more of them, because we will also have backup addresses

koonopek:
@hatskier yeah that was my initial idea, when we were discussing this on last daily. However, I thought that @cehali suggested that it would be better if we keep all the info in one place( for example we don't have to duplicate info about provider)I thought that we agreed on this

We are using array here, so we allow more then one address.

hatskier (approved):

koonopek:
Will be in remote config (discussed offline)

PR #623, opened 2023-05-11
feat: add steak hut tokens

cehali:


packages/oracle-node/.jest/set-env-vars.js
```diff
@@ -9,7 +9,6 @@ process.env.OVERRIDE_PRICE_CACHE_SERVICE_URLS = `["http://mock-price-cache-servi
 process.env.ETHERSCAN_API_URL = "http://mock-etherscan-api-url";
 process.env.ETHERSCAN_API_KEY = "test-api-key";
 process.env.LEVEL_DB_LOCATION = "level-db-for-tests";
-process.env.AVALANCHE_RPC_URL = "http://mock-avalanche-rpc-url";
```
hatskier, line 12: Why?

packages/oracle-node/src/fetchers/evm-chain/shared/utils/get-tokens-prices-from-local-cache.ts
```diff
@@ -0,0 +1,14 @@
+import Decimal from "decimal.js";
+import { getLastPrice } from "../../../../db/local-db";
+
+export const getTokensPricesFromLocalCache = (tokenNames: string[]) => {
```
hatskier, line 4: IMO it's much safer to return an object `{symbol: value}` instead of array. WDYT@ @cehali @koonopek

cehali: Changed to object

packages/oracle-node/.jest/set-env-vars.js
cehali: Change to condition

hatskier (approved):

koonopek (approved):
GJ 👍

PR #628, opened 2023-05-14
fix: generating blockSequence in MultiBlockCurveFetcher and add it to main

koonopek:


hatskier (approved):

hatskier:
Please name pull requests according to conventional commits convention

PR #634, opened 2023-05-15
test: add upgrades tests to on-chain-relayer contracts

koonopek:


packages/on-chain-relayer/test/contracts/price-feeds/common/price-feed-utils.ts
```diff
@@ -127,12 +129,66 @@ export const describeCommonPriceFeedTests = ({
...
     it("should properly upgrade the contract (change data feed adapter)", async () => {
-      expect(1).to.be.equal(1);
+      const updateContractFactory = await ethers.getContractFactory(
+        "PriceFeedsUpdatedMock"
+      );
+
+      const updatedContract = (await upgrades.upgradeProxy(
+        contractV1,
+        updateContractFactory
+      )) as PriceFeedBase;
```
cehali, line 168: This part is the same here and in `should properly upgrade the contract (change data feed id)` test. It seems that in both cases you upgrade the price feed contract

cehali: ohhh, my bad sorry

cehali: But still, this part is the same, maybe put it in one test?

packages/on-chain-relayer/contracts/mocks/PriceFeedsUpdatedMock.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity ^0.8.4;
+
+import "../price-feeds/with-rounds/PriceFeedWithRounds.sol";
+
+contract PriceFeedsUpdatedMock is PriceFeedWithRounds {
```
hatskier, line 7: PriceFeedsUpdatedMock -> PriceFeedUpdatedMock?

packages/on-chain-relayer/test/contracts/price-feeds/common/price-feeds-adapter-utils.ts
```diff
@@ -104,20 +104,68 @@ export const describeCommonPriceFeedsAdapterTests = ({
...
+
+      expect(dataFeeds.length).to.eq(2);
+      expect(dataFeeds).to.deep.eq([
+        "0x4554480000000000000000000000000000000000000000000000000000000000",
+        "0x4254430000000000000000000000000000000000000000000000000000000000",
+      ]);
+    });
+
+    it("should properly upgrade the contract (change authorized updaters)", async () => {
+      const updatedContractFactory = await ethers.getContractFactory(
```
hatskier, line 154: Can we check another authorised signer before the contract update to be sure that it has changed?

hatskier (commented):
Can you please also implement tests for access control for the upgrades? We need to be able to specify who is the "owner" (who can update the contracts)

packages/on-chain-relayer/test/contracts/price-feeds/common/price-feed-utils.ts
koonopek: Good point
Refactored common part to new describe block `before`

packages/on-chain-relayer/test/contracts/price-feeds/common/price-feeds-adapter-utils.ts
koonopek: will do

koonopek:
> Can you please also implement tests for access control for the upgrades? We need to be able to specify who is the "owner" (who can update the contracts)

Hmm but where such tests should be placed, because it is more like testing `ProxyAdmin`, `TransparentProxy` and plugin by openzeppelin then our contracts?

hatskier (approved):

PR #642, opened 2023-05-16
feat: add avalanche dry run test

cehali:


packages/oracle-node/test/dry-run/dry-run-tests-configs.ts
```diff
@@ -25,6 +27,11 @@ const config: Record<DryRunTestType, DryRunTestConfig> = {
     manifest: stocksManifest,
     nodeIterations: 1,
   },
+  [DryRunTestType.avalanche]: {
+    manifest: avalancheManifest,
+    nodeIterations: 4,
+    additionalCheck: assertAllAvalancheTokensAreProperlyFetched,
```
koonopek, line 33: I think we are duplicating here with `assertMainRequiredTokensAreProperlyFetched`
Maybe we should just pass manifest here :)
And `assertAllAreProperlyFetched` should happen in dry-run "main" function

cehali: Good catch, fixed

koonopek (dismissed):
GJ 🥇

hatskier (approved):

hatskier (approved):

PR #651, opened 2023-05-17
chore: on-chain relayer contracts improvements after abdk audit

hatskier:


cehali (dismissed):
LGTM 👍

packages/on-chain-relayer/contracts/price-feeds/PriceFeedBase.sol
```diff
@@ -1,32 +1,100 @@
...
 
-  function latestRoundData() public view override returns (uint80 roundId, int256 answer, uint256 startedAt, uint256 updatedAt, uint80 answeredInRound) {
+
+  /**
+   * @notice Returns details of the latest successful update round
+   * @dev It uses few helpful functions to abstract logic of getting
+   * latest round id and value
+   * @return roundId The number of the latest round
+   * @return answer The latest reported value
+   * @return startedAt We do not have the exact analogue for the Chainlink value
```
cehali, line 81: Recently we got info regarding `startedAt` from one of the projects:  `What’s the significance of different values of startedAt and updatedAt. As per my understanding they should be same after chainlink’s OCR update. Please let me know if we need to take special care for these values.`

packages/on-chain-relayer/contracts/price-feeds/PriceFeedBase.sol
```diff
@@ -35,15 +103,31 @@ abstract contract PriceFeedBase is IPriceFeed, Initializable {
...
+
+    // We want to be compatible with Chainlink's interface
+    // And in our case the roundId is always equal to answeredInRound
     answeredInRound = roundId;
   }
 
+  /**
+   * @notice Old Chainlink function for getting the number of latest round
+   * @return latestRound The number of the latest update round
+   */
```
cehali, line 115: Wrong description?

packages/on-chain-relayer/contracts/price-feeds/data-services/VSTPriceFeedsAdapter.sol
```diff
@@ -1,12 +1,12 @@
...
+pragma solidity ^0.8.14;
 
-import "../without-rounds/SinglePriceFeedAdapter.sol";
+import {SinglePriceFeedAdapter} from "../without-rounds/SinglePriceFeedAdapter.sol";
 
 contract VSTPriceFeedsAdapter is SinglePriceFeedAdapter {
-
-  uint256 internal constant BIT_MASK_TO_CHECK_CIRCUIT_BREAKER_FLAG = 0x0000000000000000000000000100000000000000000000000000000000000000;
+  uint256 internal constant BIT_MASK_TO_CHECK_CIRCUIT_BREAKER_FLAG =
+    0x0000000000000000000000000100000000000000000000000000000000000000;
```
cehali, line 9: Unnecessary new line?

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -159,66 +232,142 @@ abstract contract RedstoneAdapterBase is RedstoneConsumerNumericBase, IRedstoneA
...
       sstore(LATEST_UPDATE_TIMESTAMPS_STORAGE_LOCATION, timestamps)
     }
   }
 
+  /**
+   * @notice Returns the latest properly reported value of the data feed
+   * @param dataFeedId The identifier of the requested data feed
+   * @return value The latest value of the given data feed
+   */
   function getValueForDataFeed(bytes32 dataFeedId) public view returns (uint256) {
```
cehali, line 314: I understand that we have validation and then `getValueForDataFeedUnsafe` in this function but maybe we could write something about it in the description as it could a little bit misleading that we use 'unsafe' function here

koonopek (dismissed):
LGTM! 
(was looking at source code changes not comments)

packages/on-chain-relayer/contracts/price-feeds/PriceFeedBase.sol
hatskier: Good catch. Also I really didn't like this dividing by 1000, it would not work properly with block numbers instead of timestamps

hatskier: Changed

packages/on-chain-relayer/contracts/price-feeds/PriceFeedBase.sol
hatskier: Good catch!

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
hatskier: Added a comment

cehali (approved):
LGTM

koonopek (approved):

packages/on-chain-relayer/contracts/core/RedstoneAdapterBase.sol
```diff
@@ -43,26 +46,50 @@ abstract contract RedstoneAdapterBase is RedstoneConsumerNumericBase, IRedstoneA
...
+    // By default, anyone can update data feed values, but it can be overridden
   }
 
-  // This function can be overriden to reduce gas costs
+  /**
+   * @notice Returns identifiers of all data feeds supported by the Adapter contract
+   * @dev this function must be implemented in derived contracts
+   * @return An array of data feed identifiers
+   */
+  function getDataFeedIds() public view virtual returns (bytes32[] memory);
```
koonopek, line 67: @Lukasz2891 here is fix already

PR #654, opened 2023-05-18
feat: change effective date parsing in new york fed

cehali:


cehali:
Based on calculation from Voltz
image

hatskier (approved):

packages/oracle-node/src/fetchers/newyorkfed/NewyorkfedFetcher.ts
```diff
@@ -92,8 +91,8 @@ export class NewyorkfedFetcher extends BaseFetcher {
 
   // We want effective date as timestamp with hour set to 8:00am EDT (New York timezone)
```
rpiszczatowski, line 92: Technically, this comment is not exactly correct. EDT is incorrect here as in New York we have EST or EDT depending on season.

PR #657, opened 2023-05-18
fix: uncomment passing gasLimit to transaction

koonopek:


packages/on-chain-relayer/src/core/contract-interactions/update-prices.ts
```diff
@@ -68,7 +69,7 @@ const updatePricesInPriceFeedsAdapter = async ({
 }: UpdatePricesArgs): Promise<TransactionResponse> => {
   return await wrapContract(adapterContract).updateDataFeedsValues(
     proposedTimestamp,
-    // TX_CONFIG
+    TX_CONFIG
```
koonopek, line 72: :))))

cehali (approved):

PR #663, opened 2023-05-21
feat: Add support for datapackges with custom decimals

koonopek:


packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
```diff
@@ -14,12 +17,17 @@ export const valueDeviationCondition = (
     for (const { dataPackage } of dataPackages[dataFeedId]) {
       for (const dataPoint of dataPackage.dataPoints) {
         const valueFromContract = valuesFromContract[dataFeedId];
-        const valueFromFetchedDataPackage = (
-          dataPoint.toObj() as INumericDataPoint
-        ).value;
+        const fetchedDataPackage = dataPoint.toObj() as INumericDataPoint;
```
hatskier, line 20: Maybe fetchedDataPackage -> dataPointObj?

hatskier (dismissed):

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
koonopek: done

hatskier (approved):

PR #665, opened 2023-05-22
feat: add contract details config validation

cehali:


packages/oracle-node/src/fetchers/evm-chain/shared/utils/get-contract-details-from-config.ts
```diff
@@ -0,0 +1,10 @@
+export const getContractDetailsFromConfig = <T extends string, K>(
+  contractDetailsConfig: Record<T, K>,
+  id: T
+) => {
+  const contractDetails = contractDetailsConfig?.[id];
+  if (!contractDetails) {
+    throw new Error(`Misconfiguration for ${id}, cannot get contract details`);
```
koonopek, line 7: `terminateWithManifestConfigError`?

cehali: It isn't manifest error

koonopek: I think it is because, we configured in manifest fetcher for token which is not supported by this fetcher. We are missing some details about it. Also if we don't terminate(proccess.exit) here we still won't get information at the start of the node that some fetcher won't ever report prices for some token.

koonopek (changes requested):
Comment for github guys

packages/oracle-node/src/fetchers/evm-chain/shared/utils/get-contract-details-from-config.ts
cehali: ok, convinced

koonopek (approved):

koonopek (approved):

PR #667, opened 2023-05-23
feat: add bootstraping on value deviation

koonopek:


hatskier:
We decided to not use this approach (as it's not really clean) and simply run the first iteration manually using a script

PR #675, opened 2023-05-25
feat: pass values as strings to cache service

cehali:


packages/oracle-node/src/aggregated-price-handlers/DataPackageBroadcastPerformer.ts
```diff
@@ -137,14 +137,10 @@ export class DataPackageBroadcastPerformer
     await this.performBroadcast(promises, "data package");
   }
 
-  private priceToDataPoint(price: PriceDataAfterAggregation): NumericDataPoint {
-    return new NumericDataPoint({
+  private priceToDataPoint(price: PriceDataAfterAggregation): StringDataPoint {
+    return new StringDataPoint({
```
koonopek, line 141: I think we still want to use `NumericDataPoint`, we don't want to abandon this logic:
```js
    const decimals = useDefaultIfUndefined(
      numericDataPointArgs.decimals,
      DEFAULT_NUM_VALUE_DECIMALS
    );
    const valueByteSize = useDefaultIfUndefined(
      numericDataPointArgs.valueByteSize,
      DEFAULT_NUM_VALUE_BS
    );
    const valueBytes = convertNumberToBytes(
      numericDataPointArgs.value,
      decimals,
      valueByteSize
    );
```
But just `NumericDataPoint` implementation to accept string representing number. Also we should avoid in the implementation parsing to `Number()` like it is happening now in this function: `convertNumberToBytes``  let stringifiedNumber = roundFractionalComponentIfExceedsDecimals
    ? Number(value).toFixed(decimals)
    : String(value);`

koonopek (changes requested):

packages/oracle-node/src/aggregated-price-handlers/DataPackageBroadcastPerformer.ts
cehali: I will provide another changes and I have to reconsider this but I see your point

PR #680, opened 2023-05-26
feat: add SafeBigNumber and process price as big number

cehali:


packages/oracle-node/src/numbers/SafeBigNumber.ts
```diff
@@ -0,0 +1,92 @@
...
+  add(other: NumberArg): SafeBigNumber {
+    return new SafeBigNumber(this._value.add(other.toString()), this._decimals);
+  }
+
+  sub(other: NumberArg): SafeBigNumber {
+    return new SafeBigNumber(this._value.sub(other.toString()), this._decimals);
+  }
+
+  div(other: NumberArg): SafeBigNumber {
+    return new SafeBigNumber(this._value.div(other.toString()), this._decimals);
```
koonopek, line 47: What happens if we do: `SafeBigNumber.from(1).div(10)` ?

packages/oracle-node/test/numbers/SafeBigNumber.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    describe("div", () => {
+      test("should div", () => {
+        const result = SafeBigNumber.from(100, DEFAULT_DECIMALS).div(100);
+        expect(result).toEqual(SafeBigNumber.from(1, DEFAULT_DECIMALS));
+      });
+
+      test("should return 0 if float < 0", () => {
+        const result = SafeBigNumber.from(10, DEFAULT_DECIMALS).div(100);
+        expect(result).toEqual(SafeBigNumber.from(0, DEFAULT_DECIMALS));
```
koonopek, line 109: Why is this ok?

packages/oracle-node/src/numbers/SafeBigNumber.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  abs(): SafeBigNumber {
+    return new SafeBigNumber(this._value.abs(), this._decimals);
+  }
+
+  add(other: NumberArg): SafeBigNumber {
+    return new SafeBigNumber(this._value.add(other.toString()), this._decimals);
+  }
+
+  sub(other: NumberArg): SafeBigNumber {
```
koonopek, line 42: What happens if `other` is float in any case?

packages/oracle-node/src/numbers/SafeBigNumber.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  assertNonNegative() {
+    utils.assert(
+      !this._value.isNegative(),
+      `${this._value.toString()} is smaller than 0`
+    );
+  }
+
+  unsafeToNumber() {
+    return this._value.toNumber();
```
koonopek, line 82: Should we divde by `10**this.decimals` here

packages/oracle-node/test/numbers/SafeBigNumber.spec.ts
cehali: It is the same as in BigNumber.js

packages/oracle-node/src/numbers/SafeBigNumber.ts
cehali: It will throw an error, just like BigNumber.js

PR #682, opened 2023-05-28
feat: implemented an extended version of db comparison script

hatskier:


packages/cache-service/scripts/compare-dbs-alex.ts
```diff
@@ -0,0 +1,428 @@
...
+const PAGE_SIZE_MILLISECONDS = 24 * 3600 * 1000;
+const EXCEPTIONAL_DEVIATION = 4242;
+const DATA_SERVICE_ID_1 = "redstone-avalanche-demo";
+const DATA_SERVICE_ID_2 = "redstone-avalanche-prod";
+const DB_URL_1 = getRequiredEnv("MONGO_DB_DEV");
+const DB_URL_2 = getRequiredEnv("MONGO_DB_PROD");
+
+// Mappings from a signer address to an expected number of data packages
+const EXPECTED_SIGNERS_1 = {
+  "0x3a7d971De367FE15D164CDD952F64205F2D9f10c": 1,
```
koonopek, line 27: EXPECTED_SIGNERS_1 -> EXPECTED_SIGNERS_DEMO
EXPECTED_SIGNERS_2 -> EXPECTED_SIGNERS_PROD
?

packages/cache-service/scripts/compare-dbs-alex.ts
```diff
@@ -0,0 +1,428 @@
...
+      startTimestamp: currentTimestamp,
+      endTimestamp: Math.min(
+        currentTimestamp + pageSizeMilliseconds,
+        endTimestamp
+      ),
+    });
+  }
+
+  // We reverse the array to start loading from the newest data packages
+  intervals.reverse();
```
koonopek, line 337: I think we could achive the same in for loop, but I don't care in case of scripts like this

packages/cache-service/scripts/compare-dbs-alex.ts
```diff
@@ -0,0 +1,428 @@
...
+
+  return allTimestamps;
+}
+
+function checkMissesInSets<T>(
+  context: string,
+  set1: Set<T>,
+  set2: Set<T>
+): void {
+  const onlyInFirst = [...set1].filter((element) => !set2.has(element));
```
koonopek, line 371: Why shallow copy here?

packages/cache-service/scripts/compare-dbs-alex.ts
```diff
@@ -0,0 +1,428 @@
...
+  for (const dataPackagesForManyFeeds of Object.values(dataPackages)) {
+    for (const dataFeedId of Object.keys(dataPackagesForManyFeeds)) {
+      allDataFeedIds.add(dataFeedId);
+    }
+  }
+
+  return allDataFeedIds;
+}
+
+function extractAllTimestamps(dataPackages: DataPackagesResponse): Set<number> {
```
koonopek, line 356: `return new Set(Object.keys(dataPackages).map(Number))`

packages/cache-service/scripts/compare-dbs-alex.ts
```diff
@@ -0,0 +1,428 @@
...
+  });
+}
+
+function calcDeviation(value1: number, value2: number): number {
+  if (value1 <= 0 || value2 <= 0) {
+    throw new Error(
+      `Values must be positive. Received: ${value1} and ${value2}`
+    );
+  }
+  return (Math.abs(value1 - value2) / Math.min(value1, value2)) * 100;
```
koonopek, line 389: I think this is wrong way of calucating deviation it is unfair
`Math.min(value1,value2)` => `value2` 
and add before `if(value2 === 0) return Number.MAX_SAFE_INTEGER`

packages/cache-service/scripts/compare-dbs-alex.ts
```diff
@@ -0,0 +1,428 @@
...
+
+function getMaxDeviation(values1: number[], values2: number[]): number {
+  if (values1.length === 0 || values2.length === 0) {
+    console.log(
+      `Deviation can not be calculated correctly. One of the arrays is empty.`
+    );
+    return EXCEPTIONAL_DEVIATION;
+  }
+
+  return Math.max(
```
koonopek, line 216: I think `calcDeviation(Math.min(...values1), Math.max(...values2))` is enough, what do you think that we have to calculate `deviation` for every possible worst case.

koonopek (changes requested):
Very clean script, pleasure to read ;)
Left some comments and please rename script name to something more meanigful

packages/cache-service/scripts/compare-dbs-alex.ts
hatskier: I didn't want to make this assumption that something is always and demo and the other thins is prod

packages/cache-service/scripts/compare-dbs-alex.ts
hatskier: It's not a shallow copy, it's converting a set to an array and then filtering
More info here: https://stackoverflow.com/questions/33234666/how-to-map-reduce-filter-a-set-in-javascript

packages/cache-service/scripts/compare-dbs-alex.ts
hatskier: This way of calculating deviation jsut returns max value. It works like max(MikeDeviation(val1, val2), MikeDeviation(val2, val1)) and in my opinion it is not so bad in the case of this script

packages/cache-service/scripts/compare-dbs-alex.ts
hatskier: Because e.g. it may be possible that values1 have a very big value outlier

packages/cache-service/scripts/compare-dbs-alex.ts
hatskier: Yes, it's way nicer 👍

cehali (approved):

PR #686, opened 2023-05-31
feat: prepare primary manifest

hatskier:


packages/oracle-node/manifests/data-services/primary.json
```diff
@@ -0,0 +1,90 @@
...
+        "kucoin",
+        "phemex"
+      ],
+      "skipSigning": true
+    },
+    "FRAX": {
+      "source": [
+        "coingecko",
+        "coinmarketcap",
+        "curve-usdc",
```
koonopek, line 79: Do we want `curve-usdc-multi-block` and `curve-usdc` here?
I think that it doesn't have sense in long term

koonopek (approved):

packages/oracle-node/manifests/data-services/primary.json
hatskier: We want to have only multi-block in the primary data service. Otherwise the whole advantage of additional manipulation protection goes away

PR #695, opened 2023-06-05
feat: remove reporting lost precision in JsNativeSafeNumber

koonopek:


packages/oracle-node/src/numbers/JsNativeSafeNumber.ts
```diff
@@ -216,13 +215,3 @@ const parseToSafeNumber = (value: utils.NumberLike) => {
 
   return number;
 };
-
-const reportIfLoosingPrecision = (number: string) => {
-  const [_, decimals] = number.split(".");
-
-  if (decimals && decimals.length > JsNativeSafeNumberConfig.MAX_DECIMALS) {
```
rpiszczatowski, line 223: @koonopek Why do we remove that feature?

koonopek: Cause it doesn't add any value and puts a lot of trash in logs.
Also we we introduced it for later analyses which were conducted and it doesn't seem to be a problem.
https://github.com/orgs/redstone-finance/projects/5/views/16?filterQuery=assignee%3Akoonopek+num&pane=issue&itemId=28541707

PR #697, opened 2023-06-05
feat: add swETH token

cehali:


packages/oracle-node/src/fetchers/balancer/balancer-fetchers-config.json
```diff
@@ -1,8 +1,20 @@
 {
   "balancer-dai": {
-    "baseToken": "DAI"
+    "OHM": {
```
hatskier, line 3: Why did you decide to provide this change? Looks like now we will  need to duplicate `pairedToken` for each asset in each `balancer-xxx`, instead of sepcicifying it once (as we did before)

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -61,38 +67,22 @@ export class BalancerFetcher extends DexOnChainFetcher<BalancerResponse> {
...
+    const lastPriceFromCache = getLastPrice(pairedToken);
     if (!lastPriceFromCache) {
-      throw new Error(`Cannot get last price from cache for: ${tokenToGet}`);
+      throw new Error(`Cannot get last price from cache for: ${pairedToken}`);
     }
     return lastPriceFromCache.value;
   }
 
   protected getPoolIdForAssetId(assetId: string) {
-    const poolFound = balancerPools.find((pool) => {
```
hatskier, line 83: Nice!

packages/oracle-node/src/fetchers/balancer/BalancerFetcher.ts
```diff
@@ -6,14 +6,20 @@ import {
...
 import { getLastPrice } from "../../db/local-db";
-import balancerPools from "./balancer-pools.json";
 import { config } from "../../config";
 
 const balancerConfig: BalancerSdkConfig = {
   network: Network.MAINNET,
   rpcUrl: config.ethMainRpcUrl as string,
 };
 
+export interface BalancerPoolConfig {
```
hatskier, line 16: I don't have access to this code to comment
`calcSpotPrice(pool.tokens[1].address, pool.tokens[0].address)`

But looks like it can work incorrectly for pools with tokens.length > 2, which is the case for the SWETH pool

packages/oracle-node/src/fetchers/balancer/balancer-fetchers-config.json
cehali: I provide this change to merge two config files into one, make it similar to the ones we have for other DEXes and simplify code like in here https://github.com/redstone-finance/redstone-oracles-monorepo/pull/697#discussion_r1218391292

cehali: Sure, will fix it

hatskier (dismissed):

hatskier (dismissed):

hatskier (approved):

PR #698, opened 2023-06-06
feat: add maverick fetcher (for SWETH)

koonopek:


packages/oracle-node/src/fetchers/maverick/maverick-fetcher-config.ts
```diff
@@ -0,0 +1,13 @@
+import { ethereumProvider } from "../../utils/blockchain-providers";
+
+export default {
+  // this contract is stateless
+  poolInformationAddress: "0xadc6ced7666779ede88e82c95e363450ac59bfd3",
+  provider: ethereumProvider,
+  tokens: {
+    SWETH: {
+      pairedToken: "WETH",
```
rpiszczatowski, line 9: In other fetchers, the term `pairedToken` is used for a different purpose. Maybe we should stay consistent.

hatskier: I agree

packages/oracle-node/src/fetchers/maverick/MaverickFetcher.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  override calculateLiquidity(
+    _assetId: string,
+    _response: MaverickResponse
+  ): number {
+    throw new Error(`Calculating liquidity not implemented for ${this.name}`);
+  }
+
+  protected getPairedTokenPrice(id: string): Decimal {
+    let tokenToGet = id === "WETH" ? "ETH" : id;
```
hatskier, line 59: I think we should get rid of this logic and be consistent with other fetchers where we just specify the pairedTokenPrice

packages/oracle-node/src/fetchers/maverick/maverick-fetcher-config.ts
koonopek: done

hatskier (approved):

packages/contract-deployments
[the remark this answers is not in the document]
hatskier: We'll remove it in a separate PR

PR #702, opened 2023-06-06
feat: add test for unique signers when fetching data packages from cache service

cehali:


packages/cache-service/test/e2e/data-packages.spec.ts
```diff
@@ -229,6 +231,28 @@ describe("Data packages (e2e)", () => {
     );
   });
 
+  it("/data-packages/latest (GET) return packages with unique signers if multiple packages with repeated singers", async () => {
+    await DataPackage.insertMany(mockDataPackagesForUniqueSigners);
+
+    const dpTimestamp = mockDataPackages[0].timestampMilliseconds;
+    Date.now = jest.fn(() => dpTimestamp);
```
koonopek, line 238: Can we clear that mock after test, to avoid polluting global objects

koonopek (changes requested):
<3 one small comment

packages/cache-service/test/e2e/data-packages.spec.ts
cehali: Done

packages/cache-service/test/e2e/data-packages.spec.ts
```diff
@@ -98,7 +98,10 @@ describe("Data packages (e2e)", () => {
     await DataPackage.insertMany(dataPackagesToInsert);
   });
 
-  afterEach(async () => await dropTestDatabase());
+  afterEach(async () => {
+    jest.clearAllMocks();
```
koonopek, line 102: I think It could work if we do something like:
`jest.spy(Date,'now').mockImplementation(() = > 1200)`
But it won't work for `Date.now = jest.fn(() => 1200)`, because we didn't use `jest` mock mechanics in first place thus jest doesn't know what is original function.

koonopek (changes requested):

packages/cache-service/test/e2e/data-packages.spec.ts
koonopek: I think, I have hard time to explain what I mean, maybe code will be self explanatory
image

cehali: New change should work 
image

koonopek: Exactly!

koonopek (dismissed):
LGTM!

cehali:
@koonopek please reapprove 😃

koonopek (approved):

packages/cache-service/test/e2e/data-packages.spec.ts
```diff
@@ -428,24 +455,32 @@ describe("Data packages (e2e)", () => {
...
+    const dpTimestamp = mockDataPackages[0].timestampMilliseconds;
+    jest.spyOn(Date, "now").mockImplementation(() => dpTimestamp);
     await performPayloadTests((response) => {
       return response.body;
     }, "raw");
   });
 
   it("/data-packages/payload (GET) - should return payload in bytes format", async () => {
+    const dpTimestamp = mockDataPackages[0].timestampMilliseconds;
+    jest.spyOn(Date, "now").mockImplementation(() => dpTimestamp);
```
koonopek, line 475: I think you could write it in `performPayloadTests` then you wouldnt need to repeat so many times :)

koonopek (approved):

PR #713, opened 2023-06-07
feat: filter tokens with skip signing when validating big package

cehali:


packages/oracle-node/src/validators/validate-data-feed-for-big-package.ts
```diff
@@ -1,17 +1,40 @@
...
+  dataPoints: DataPoint[],
+  allTokens: TokensConfig
+) =>
+  dataPoints.filter(
+    (dataPoint) => !allTokens[dataPoint.dataFeedId].skipSigning
+  );
+
+const countTokensWithoutSkipSigning = (allTokens: TokensConfig) =>
+  Object.values(allTokens).reduce(
+    (count, config) => (config.skipSigning ? count : ++count),
```
hatskier, line 38: config -> tokenConfig?

hatskier: I would also change `++count` to count + 1, as we don't need to increase the value of the variable here, we only need to return the new value

cehali: Done

packages/oracle-node/src/validators/validate-data-feed-for-big-package.ts
```diff
@@ -1,17 +1,40 @@
...
   dataPoints: DataPoint[],
-  allTokenCount?: number
+  allTokensConfig?: TokensConfig
 ) => {
-  if (!allTokenCount) {
-    throw new Error(
-      `Cannot get token count from manifest.`
-    );
+  if (!allTokensConfig) {
+    throw new Error(`Cannot get token config from manifest.`);
```
koonopek, line 10: Could we use `terminateWithManifestConfigError` here?

packages/oracle-node/src/validators/validate-data-feed-for-big-package.ts
```diff
@@ -1,17 +1,40 @@
...
   }
-  const validDataPointsPercentage = (dataPoints.length / allTokenCount) * 100;
+
+  const dataPointsWithoutSkipSigning = filterDataPointsWithoutSkipSigning(
+    dataPoints,
+    allTokensConfig
+  );
+  const tokensWithoutSkipSigningCount =
```
koonopek, line 17: What should happen if `tokensWithoutSkipSigningCount === 0`, should we throw or return `false`?

koonopek (changes requested):

packages/oracle-node/src/validators/validate-data-feed-for-big-package.ts
cehali: I think we should return `false` and not create big data package

koonopek (changes requested):

koonopek (approved):

PR #717, opened 2023-06-07
fix: do not run all conditions in conditions object

cehali:


hatskier (dismissed):

cehali:
Without function, whenever the `checkConditionByName` object was used, it run all conditions and produces errors. I had this situation when I wanted to run relayer based only on `time` but the `value-deviation` check was also triggered.

packages/on-chain-relayer/src/core/update-conditions/should-update.ts
```diff
@@ -22,9 +22,7 @@ export const shouldUpdate = (context: Context): ConditionCheckResponse => {
...
 
 const checkConditionByName = (context: Context) => ({
-  time: timeUpdateCondition(context.lastUpdateTimestamp),
-  "value-deviation": valueDeviationCondition(
-    context.dataPackages,
-    context.valuesFromContract
-  ),
+  time: () => timeUpdateCondition(context.lastUpdateTimestamp),
+  "value-deviation": () =>
+    valueDeviationCondition(context.dataPackages, context.valuesFromContract),
```
koonopek, line 27: I would even go on step further with this change, make this functions async and for  example for `valueDeviationCondition` would fetchLastPrice from contract only if `valueDeviationCondition` is set. Now it is handled by weird if (if i remenber correctly)

koonopek (changes requested):

packages/on-chain-relayer/src/core/update-conditions/should-update.ts
cehali: Good idea, changed

packages/on-chain-relayer/src/core/update-conditions/should-update.ts
```diff
@@ -22,9 +24,10 @@ export const shouldUpdate = (context: Context): ConditionCheckResponse => {
...
 
 const checkConditionByName = (context: Context) => ({
-  time: timeUpdateCondition(context.lastUpdateTimestamp),
-  "value-deviation": valueDeviationCondition(
-    context.dataPackages,
-    context.valuesFromContract
-  ),
+  time: () => timeUpdateCondition(context.lastUpdateTimestamp),
+  "value-deviation": async () =>
+    await valueDeviationCondition(
```
koonopek, line 29: Can we do the same with `getLastRoundParamsFromContract` and `time` ?

cehali: `lastUpdateTimestamp` is used not only in `timeUpdateCondition` but also in `updatePrices` function

koonopek (approved):
Nice!

PR #723, opened 2023-06-12
Feat-make-electing-block-number-great-again

koonopek:
* Tunned AgrrementProvider according to recent test and consulting with @hatskier
* As the next step I propose to use it with one of Nodes (primary or demo) on set-up with 7/8 best rpcs for given chain (I have stats for that)


koonopek:
duplicated

PR #730, opened 2023-06-13
fix: use terminateWithManifestConfigError for handling configuration …

rpiszczatowski:
…errors of DEX fetchers


packages/oracle-node/src/fetchers/dex-on-chain/DexOnChainFetcher.ts
```diff
@@ -5,17 +5,22 @@ import {
...
 export interface Responses<T> {
   [spotAssetId: string]: T;
 }
 
 export abstract class DexOnChainFetcher<T> extends MultiRequestFetcher {
-  abstract calculateLiquidity(assetId: string, response: T): number;
-  calculateSlippage(assetId: string, response: T): number {
-    throw new Error("method not implemented!");
+  calculateLiquidity(_assetId: string, _response: T): number {
+    terminateWithManifestConfigError(`liquidity calculation not implemented for ${this.getName()}`);
```
hatskier, line 16: What do you think about renaming the function to smth like `terminateNodeWithError`? IMO currently it's a bit misleading why we use it here, since these errors are not related to the manifest

koonopek: Why do you think the are not related to manifest? In.my opinion they are caused because someone used "toke" in manifest which is not supported by node

rpiszczatowski: You can think of those errors as manifest-related because manifest is the place where you declare the combination of a token name (e.g. OHM_xxx_liquidity) and source that does not support liquidity calculation. This is not a problem with a source that it does not support liquidity calculation.

cehali: I agree with Mike, it is related to a misconfiguration in the manifest. We can change the error message slightly to show there is a problem in manifest like: `liquidity calculation not implemented for ${this.getName()}, you defined liquidity for a token in the manifest with fetcher which doesn't support liquidity`.

koonopek (approved):
Small step for humanity giant leap for redstone ~ Mike Armstrong

PR #734, opened 2023-06-14
feat: Add gas optimized transaction deliver

koonopek:
Tested on goerli, works stable even with gas spiking 5 times. Here is the contract I was writing to:
https://goerli.etherscan.io/address/0xddededae617cec1b1c7b60c467ac6b9787047b9f

In default setup we prefer time of delivery over low cost

### DIFF
--- a/packages/on-chain-relayer/src/core/contract-interactions/update-prices.ts
+++ b/packages/on-chain-relayer/src/core/contract-interactions/update-prices.ts
@@ -9,6 +9,7 @@ import {
 } from "../../custom-integrations/mento/mento-utils";
 
 import { getSortedOraclesContractAtAddress } from "./get-contract";
+import { TransactionDeliver } from "redstone-rpc-providers";
 
 interface UpdatePricesArgs {
   adapterContract: Contract;
@@ -17,6 +18,10 @@ interface UpdatePricesArgs {
 }
 
 const TX_CONFIG = { gasLimit: config.gasLimit };
+
+const deliveryMan = new TransactionDeliver({
+  expectedDeliveryTimeMs: config.expectedTxDeliveryTimeInMS,
+});
 
 export const updatePrices = async (
   dataPackages: DataPackagesResponse,
@@ -67,10 +72,22 @@ const updatePricesInPriceFeedsAdapter = async ({
   wrapContract,
   proposedTimestamp,
 }: UpdatePricesArgs): Promise<TransactionResponse> => {
-  return await wrapContract(adapterContract).updateDataFeedsValues(
-    proposedTimestamp,
-    TX_CONFIG
+  const wrappedContract = wrapContract(adapterContract);
+
+  const deliveryResult = await deliveryMan.deliver(
+    wrappedContract,
+    "updateDataFeedsValues",
+    [proposedTimestamp],
+    Number(TX_CONFIG.gasLimit)
   );
+
+  if (!deliveryResult) {
+    throw new Error(
+      "Transaction with sane nonce was delivered by someone else"
+    );
+  }
+
+  return deliveryResult;
 };
 
 const updatePricesInMentoAdapter = async ({
--- a/packages/on-chain-relayer/src/core/contract-interactions/update-prices.ts
+++ b/packages/on-chain-relayer/src/core/contract-interactions/update-prices.ts
@@ -78,7 +78,7 @@ const updatePricesInPriceFeedsAdapter = async ({
     wrappedContract,
     "updateDataFeedsValues",
     [proposedTimestamp],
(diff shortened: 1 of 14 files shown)


packages/rpc-providers/src/TransactionDeliver.ts
```diff
@@ -0,0 +1,254 @@
...
+
+type FeeStructure = {
+  maxFeePerGas: number;
+  maxPriorityFeePerGas: number;
+};
+
+const unsafeBnToNumber = (bn: BigNumber) => Number(bn.toString());
+
+const ethGasTrackerOracle: GasOracleFn = async (apiKey: string = "") => {
+  const response = // rate limit is 5 seconds
```
hatskier, line 59: Won't we hit the rate limit many times then?

hatskier: Shouldn't we implement caching here?

packages/rpc-providers/src/TransactionDeliver.ts
```diff
@@ -0,0 +1,254 @@
...
+};
+
+type FeeStructure = {
+  maxFeePerGas: number;
+  maxPriorityFeePerGas: number;
+};
+
+const unsafeBnToNumber = (bn: BigNumber) => Number(bn.toString());
+
+const ethGasTrackerOracle: GasOracleFn = async (apiKey: string = "") => {
```
hatskier, line 58: Can you name function as a verb please?

packages/rpc-providers/src/TransactionDeliver.ts
```diff
@@ -0,0 +1,254 @@
...
+} as Record<number, GasOracleFn | undefined>;
+
+const DEFAULT_TRANSACTION_DELIVERY_OPTS = {
+  maxAttempts: 5,
+  priorityFeePerGasMultiplier: 1.125, // 112,5%
+  percentileOfPriorityFee: 75,
+  logger: (text: string) => console.log(`[${TransactionDeliver.name}] ${text}`),
+};
+
+export class TransactionDeliver {
```
hatskier, line 91: Maybe TransactionDeliveryMan?

packages/rpc-providers/src/TransactionDeliver.ts
```diff
@@ -0,0 +1,254 @@
...
+  public async deliver<T extends Contract, M extends keyof T>(
+    contract: T,
+    method: M,
+    params: Parameters<T[M]>,
+    gasLimit?: number
+  ): Promise<TransactionResponse | undefined> {
+    const provider = contract.provider as providers.JsonRpcProvider;
+    const address = await contract.signer.getAddress();
+
+    let lastAttempt = undefined as LastDeliveryAttempt | undefined;
```
hatskier, line 107: I would use `let lastAttempt: LastDeliveryAttempt | undefined = undefined;` according to typescript best practices

packages/rpc-providers/src/TransactionDeliver.ts
koonopek: Yep

packages/rpc-providers/src/TransactionDeliver.ts
koonopek: Yes! I was thinking about it ;)

packages/rpc-providers/src/TransactionDeliver.ts
koonopek: Good point :)

packages/rpc-providers/src/TransactionDeliver.ts
koonopek: Good idea :) 5s caching should solve all problems

koonopek: But it still wouldn't hit the rate limit cause we call this function once every 15 seconds there is no point IN ETH blockchain to call it more frequently

packages/on-chain-relayer/src/core/contract-interactions/update-prices.ts
```diff
@@ -67,10 +72,22 @@ const updatePricesInPriceFeedsAdapter = async ({
...
+  const wrappedContract = wrapContract(adapterContract);
+
+  const deliveryResult = await deliveryMan.deliver(
+    wrappedContract,
+    "updateDataFeedsValues",
+    [proposedTimestamp],
+    TX_CONFIG.gasLimit ? Number(TX_CONFIG.gasLimit) : undefined
   );
+
+  if (!deliveryResult) {
```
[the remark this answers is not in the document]
koonopek: Yeah good point :)

hatskier (approved):

PR #740, opened 2023-06-16
chore: added relayer manifests

hatskier:


koonopek:
Hmm that means that we have to update it every time we deploy new service. What do you think 2 alternative solutions:
* link (as dependency) contract-deployments repo to monitoring (both are private so we are not putting on risk anything)
* publish contract-deployments/deployments on public github pages (info there is public anyway trough blockchain)

cehali (approved):

koonopek (approved):

PR #754, opened 2023-06-21
feat: improve mock fetcher and add USDf token

cehali:


packages/oracle-node/manifests/data-services/main.json
```diff
@@ -854,6 +854,10 @@
         "xt"
       ]
     },
+    "USDf": {
+      "source": ["mock"],
+      "fixedValue": 1
```
koonopek, line 859: I don't get this dualism that we don't have single source of truth github or manifest but rather both of them

packages/oracle-node/src/fetchers/mock-fetcher/mock-fetcher.ts
```diff
@@ -1,19 +1,24 @@
...
   }
 
-  async fetchData() {
+  async fetchData(_ids: string[], opts: FetcherOpts) {
     const isMockPricesUrl = config.mockPricesUrlOrPath.startsWith("http");
     if (isMockPricesUrl) {
       return (await axios.get(config.mockPricesUrlOrPath)).data;
     }
+
+    if (this.isAnyTokenWithFixedValue(opts.manifest.tokens)) {
```
koonopek, line 18: 1. This can be done before `axios.get` I think.
2. Also why we check `if any` and then execute for every. What if one of tokens doesnt have set `fixedValue`

koonopek (changes requested):

packages/oracle-node/manifests/data-services/main.json
cehali: It is a different situation, previously we used a mock fetcher for some tests which require frequent changes. Now we want to constantly have `USDf` as 1. I can easily imagine a situation where someone deletes a repo or file with mocked values. Here it will be hard to remove this value as it needs to pass PR and redeploy the node.

packages/oracle-node/src/fetchers/mock-fetcher/mock-fetcher.ts
cehali: 1. Explained in the previous comment, overengineering in this scenario and prone to human error. 
2. It means that we have misconfiguration in manifest, we want all of the to be from URL or with `fixedValue`

koonopek: Still don't get it because now we can one feature "mock-url" or second "fixed value in manifest", we can use two feature in the same node at the same time. And as you said earlier this are two distinct things, so why are we trying to fit them in one fetcher?
(Maybe i don't understand something)

koonopek: 1. If I setup in env `mockPricesUrlOrPath` to some URL I will never used value from `manfiest[token].fixedValue`.
2. And if I setup `mockPricesUrlOrPath` to fs path I will return prices only for tokens which have setup `fixedValue`.
3. And last case if I don't set in any token `fixedValue` and set `mockPricesUrlOrPath` to fs path I will return prices from json file.

Pretty complicated stuff, do you think we can simplify it? Or It has to be like that?

A.D 2. So maybe we should throw error if this is misconfiguration :)

packages/oracle-node/src/fetchers/fixed-value-fetcher/fixed-value-fetcher.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  async fetchData(_ids: string[], opts: FetcherOpts) {
+    return this.getFixedTokenValuesFromManifest(opts.manifest.tokens);
+  }
+
+  extractPrices(response: any, ids: string[]): PricesObj {
+    const result: PricesObj = {};
+
+    for (const id of ids) {
+      result[id] = response[id] ?? response.__DEFAULT__;
```
koonopek, line 17: What is this `__DEFAULT` thing

koonopek (approved):

packages/oracle-node/src/fetchers/fixed-value-fetcher/fixed-value-fetcher.ts
cehali: Leftover from mock fetcher, removed

koonopek (approved):
GJ!

PR #755, opened 2023-06-21
feat: add retry to curve multi block fetcher

cehali:


koonopek (dismissed):
Please veryfi if check for retrying is overriden properly (as default it checks if response !== undefined) if I remember correctly

cehali:
Good point, I will verify it

cehali:
@koonopek I checked validation and it turns out that it required additional changes

koonopek (approved):

PR #758, opened 2023-06-21
feat: add request validation to cache-service

koonopek:


packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -11,56 +11,34 @@ import {
...
-    dataServiceId: string;
-  };
-}
-
 const CONTENT_TYPE_OCTET_STREAM = "application/octet-stream";
 const CONTENT_TYPE_TEXT = "text/html";
 const CONTENT_TYPE_JSON = "application/json";
 
 @Controller("data-packages")
+@UsePipes(new ValidationPipe({ transform: true }))
```
cehali, line 41: Maybe add it at the application level: https://docs.nestjs.com/techniques/validation#auto-validation ?

koonopek: Then you also have to remember to add it in every test.
Like this you have to only add it one per every controller:
```js
    const moduleFixture: TestingModule = await Test.createTestingModule({
      imports: [AppModule],
    }).compile();

    app = moduleFixture.createNestApplication();
    // app.useGlobalPipe(new ValidationPipe())
    await app.init();
```
Also allow to pass custom option per controller thus more granular control.

packages/cache-service/src/data-packages/data-packages.interface.ts
```diff
@@ -1,5 +1,90 @@
...
+  SignedDataPackagePlainObj,
+} from "redstone-protocol";
+import { IsNumberOrString } from "../utils/IsNumberOrString";
+import { CachedDataPackage } from "./data-packages.model";
 
-export interface ReceivedDataPackage extends SignedDataPackagePlainObj {
-  sources?: object;
+export class ReceivedDataPackage implements SignedDataPackagePlainObj {
+  @IsString()
+  signature: string;
```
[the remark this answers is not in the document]
koonopek: We can but then mock-signatures and stuff like this won't work. Do we want to be so strict?

packages/cache-service/src/data-packages/data-packages.interface.ts
```diff
@@ -1,5 +1,90 @@
...
+  @IsOptional()
+  sources?: Record<string, string | number>;
+}
+
+export class ReceivedDataPoint implements IStandardDataPoint {
+  @IsString()
+  dataFeedId: string;
+
+  @Validate(IsNumberOrString)
+  value: string;
```
cehali, line 37: If we validate it to be string or number, we should stick with this typing probably?

packages/cache-service/src/data-packages/data-packages.interface.ts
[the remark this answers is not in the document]
koonopek: done, I hope we wont break something :)

hatskier (approved):

hatskier (approved):

PR #759, opened 2023-06-21
feat: add metadata to DataPoint

koonopek:


packages/protocol/src/data-point/DataPoint.ts
```diff
@@ -7,16 +7,31 @@ import { IStringDataPoint } from "./StringDataPoint";
...
+  metadata?: Metadata;
 }
-export type DataPointPlainObj =
-  | INumericDataPoint
-  | IStandardDataPoint
-  | IStringDataPoint;
+export type DataPointPlainObj = IStandardDataPoint | INumericDataPoint;
+
+export enum MetadataDataPointType {
+  PRICE_V1,
```
koonopek, line 15: Not sure about naming here.
Maybe serialization format would be better like `BASE64_FLOAT`

packages/protocol/src/data-point/DataPoint.ts
```diff
@@ -7,16 +7,31 @@ import { IStringDataPoint } from "./StringDataPoint";
...
+}
+
+export interface PriceMetadata {
```
hatskier, line 18: What do you think about not specifying types for metadata and just assuming that anything can be there, to make it easier to work with it in future?

koonopek: Hmm maybe I will add signature [x: string]: any | undefined (so it will be future proof).
But still keep types just for DX.

koonopek: (Added it)

hatskier: Okay, but I would go with removing `PriceMetadata` interface and would just add `metadata?: Record<string, any>` to data package

koonopek: Lets do this, I don't have opinion on that

packages/protocol/src/data-point/DataPoint.ts
```diff
@@ -7,16 +7,23 @@ import { IStringDataPoint } from "./StringDataPoint";
...
   value: string; // base64-encoded bytes
+  metadata?: Metadata;
 }
-export type DataPointPlainObj =
-  | INumericDataPoint
-  | IStandardDataPoint
-  | IStringDataPoint;
+export type DataPointPlainObj = IStandardDataPoint | INumericDataPoint;
+
+export type MetadataType = "HEX_BIG_INT";
```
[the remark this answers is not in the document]
hatskier: Why HEX_BIG_INT, not BASE64?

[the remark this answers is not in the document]
koonopek: Because it is base64(hex(int)). Base64 will be used always, we are using hex(int) for prices currently.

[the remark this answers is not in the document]
hatskier: Hmm, I am not sure I understand. Hex and Base64 are different ways of representing data. Hex kinda uses 16 "digits" and Base64 kinda uses 64 "digits". Looks like we either have a misunderstanding here or we are doing it quite inefficiently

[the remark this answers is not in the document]
koonopek: I think we are doing quite inefficiently. Let me explain:
In the context of node we have same number like `100.123` represented as `Number` or `String` it doesn't really matter. Currently, we are doing this `convertNumberToBytes` underneath (in shortcut):
```js
): Uint8Array => {
  let stringifiedNumber = roundFractionalComponentIfExceedsDecimals
    ? Number(value).toFixed(decimals)
    : String(value);
  const bigNumberValue = parseUnits(stringifiedNumber, decimals);
  const bytesValue = arrayify(bigNumberValue.toHexString());
```
Thus, we encode the number as hex and then cast it to Uint8Array, which is then encoded using base64 so yes, we are quite unefficent here.
I'm unsure if this will still be the case after @cehali PR. It depends on how he will convert number to DataPoint.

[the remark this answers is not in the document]
rpiszczatowski: Let me add that from what I see in node, in transport layer and in cache we do not use DataPoint objects but rather those DataPointPlainObj which contain values in their JS form, neither hexlified nor base64'd. We only use DataPoint type to create signature in node and then we create it again in cache service to verify the signature. We probably use it further in relayers, but I haven't checked that part yet.

hatskier (approved):

rpiszczatowski (approved):

PR #768, opened 2023-06-23
tmp

cehali:


packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -162,6 +162,7 @@ export class DataPackagesController {
       );
     }
 
+    console.log(body);
```
rpiszczatowski, line 165: accidentally committed?

packages/oracle-node/src/aggregated-price-handlers/DataPackageBroadcastPerformer.ts
```diff
@@ -137,10 +133,10 @@ export class DataPackageBroadcastPerformer
     await this.performBroadcast(promises, "data package");
   }
 
-  private priceToDataPoint(price: PriceDataAfterAggregation): NumericDataPoint {
-    return new NumericDataPoint({
+  private priceToDataPoint(price: PriceDataAfterAggregation): DataPoint {
+    return new DataPoint({
       dataFeedId: price.symbol,
-      value: price.value.unsafeToNumber(),
+      value: convertStringToBytes32(price.value.toString()),
```
rpiszczatowski, line 139: Do we use the correct function here? Shouldn't we use `formatBytes32String` instead? Currently, in case of input longer than 31 bytes, we would get a 'random' value. `formatBytes32String` would throw and thus more clearly state our intention here.

packages/oracle-node/src/aggregators/lwap-aggregator/get-liquidities.ts
```diff
@@ -21,9 +20,14 @@ export const getTickLiquidities = (
       );
     }
     const theOnlySourceValue = Object.values(liquidity.source)[0];
+    if (theOnlySourceValue.isNaN() || theOnlySourceValue.isZero()) {
```
rpiszczatowski, line 23: Is this new validation related to switching numbers to decimals or is it just a precaution not related to the change?

packages/oracle-node/src/fetchers/BaseFetcher.ts
```diff
@@ -138,7 +139,7 @@ export abstract class BaseFetcher implements Fetcher {
           continue;
         }
 
-        pricesObj[pricePair.id] = pricePair.value as number;
+        pricesObj[pricePair.id] = pricePair.value as Decimal;
```
rpiszczatowski, line 142:
```suggestion
        pricesObj[pricePair.id] = pricePair.value!;
```

packages/protocol/src/data-point/DataPoint.ts
```diff
@@ -1,43 +1,56 @@
...
-  | INumericDataPoint
-  | IStandardDataPoint
-  | IStringDataPoint;
 
 export class DataPoint extends Serializable {
-  constructor(
-    public readonly dataFeedId: ConvertibleToBytes32,
-    public readonly value: Uint8Array
-  ) {
+  dataFeedId: string;
```
rpiszczatowski, line 22: readonly?

packages/protocol/src/data-point/DataPoint.ts
```diff
@@ -1,43 +1,56 @@
...
 
+export interface DataPointArgs {
+  dataFeedId: ConvertibleToBytes32;
+  value: Uint8Array;
+  decimals?: number;
+}
 export interface IStandardDataPoint {
   dataFeedId: ConvertibleToBytes32;
   value: string; // base64-encoded bytes
+  decimals?: number;
```
rpiszczatowski, line 18: I don't like the idea that 'standard' data point contains the property called 'decimals'.

packages/protocol/src/data-point/DataPoint.ts
```diff
@@ -1,43 +1,56 @@
...
 export class DataPoint extends Serializable {
-  constructor(
-    public readonly dataFeedId: ConvertibleToBytes32,
-    public readonly value: Uint8Array
-  ) {
+  dataFeedId: string;
+  value: Uint8Array;
+  decimals: number;
+
+  constructor(private readonly dataPointsArgs: DataPointArgs) {
```
rpiszczatowski, line 26: Do we need to have `dataPointsArgs` as a property on `DataPoint` type?

packages/protocol/src/data-point/DataPoint.ts
```diff
@@ -7,6 +7,7 @@ import { IStringDataPoint } from "./StringDataPoint";
 export interface IStandardDataPoint {
   dataFeedId: ConvertibleToBytes32;
   value: string; // base64-encoded bytes
```
rpiszczatowski, line 9: It seems to me that the encoded value is not shifted by decimals places to the right. We used to do it is the previous version of the code.

packages/oracle-node/src/signers/EvmPriceSigner.ts
```diff
@@ -1,25 +1,23 @@
...
 /** IMPORTANT: This function as side effect convert Class instances to pure objects */
 const sortDeepObjects = <T>(arr: T[]): T[] => sortDeepObjectArrays(arr);
 
-const serializePriceValue = (value: number | ISafeNumber): number => {
+const serializePriceValue = (value: number | Decimal): number => {
   if (typeof value === "number") {
     return Math.round(value * 10 ** 8);
-  } else if (value instanceof JsNativeSafeNumber) {
-    return Math.round(value.unsafeToNumber() * 10 ** 8);
+  } else if (value instanceof Decimal) {
```
rpiszczatowski, line 19: I am wondering whether this is the correct approach. We introduce decimals to support big numbers and here we assume that this calculation will return proper value.

packages/protocol/test/plain-obj-serialization.test.ts
```diff
@@ -1,4 +1,4 @@
-import { toUtf8Bytes } from "ethers/lib/utils";
+import { arrayify, formatBytes32String, toUtf8Bytes } from "ethers/lib/utils";
```
rpiszczatowski, line 1: ???

packages/oracle-node/src/aggregators/lwap-aggregator/get-liquidities.ts
cehali: Second think, I saw errors related to that when testing

packages/protocol/src/data-point/DataPoint.ts
cehali: It can have decimals if we know that value would be a number, in other cases we don't have to pass it

packages/protocol/src/data-point/DataPoint.ts
cehali: Now it is

packages/oracle-node/src/signers/EvmPriceSigner.ts
cehali: We rather know that the result of aggregation would be `Decimal`, currently we don't support other options

rpiszczatowski: My point is that the expression under `Math.round` call may be out of range for `number` type. And in general method `serializePriceValue` effectively converts Decimal to `number`. Such a conversion may not be possible in general. The return value may be out of 'safe range' of JS numbers. Maybe this is not a problem for any reason in this particular code but if it is so we should probably add a comment explaining why it is safe to perform this operation.

Decimals-based implementation (as opposed to safe-number implementation) assumes arbitrarily big numbers are possible (maybe up to 32 bytes long, but this is much more than number can handle)

packages/protocol/src/common/utils.ts
```diff
@@ -82,6 +84,14 @@ export const convertIntegerNumberToBytes = (
 export const convertBytesToNumber = (bytes: Uint8Array): number =>
   BigNumber.from(bytes).toNumber();
 
+export const convertAndSerializeBytesToNumber = (
+  bytes: Uint8Array,
+  decimals?: number
```
koonopek, line 89: just neat pick decimals = DEFAULT_NUM_VALUE_DECIMALS

packages/protocol/src/data-point/DataPoint.ts
koonopek: I think that it should be part of metadata (still have to discuss if we should type it)

koonopek (approved):

packages/protocol/src/data-point/DataPoint.ts
cehali: I agree, that should go to metadata and then it will be used only if the data point has a numeric value

packages/oracle-node/src/fetchers/BaseFetcher.ts
cehali: I am not convinced, in essence, it is the same

packages/oracle-node/src/signers/EvmPriceSigner.ts
cehali: Agree that we could add a comment. Changes with `SafeNumber` will be in separate PR and `EvmPriceSigner` module will be deprecated, I hope soon

packages/protocol/src/common/utils.ts
cehali: Done

packages/oracle-node/src/aggregated-price-handlers/DataPackageBroadcastPerformer.ts
```diff
@@ -137,14 +133,15 @@ export class DataPackageBroadcastPerformer
...
-      value: price.value.unsafeToNumber(),
-      decimals: ManifestHelper.getDataFeedDecimals(
+  private priceToDataPoint(price: PriceDataAfterAggregation): DataPoint {
+    const decimals =
+      ManifestHelper.getDataFeedDecimals(
         this.manifestDataProvider.latestManifest!,
         price.symbol
-      ),
-    });
+      ) ?? 8;
```
koonopek, line 141: Could we use const for 8 decimals

packages/integration/integration-test-core.sh
koonopek, line 1: What happened here?
image

koonopek (dismissed):

packages/oracle-node/src/aggregated-price-handlers/DataPackageBroadcastPerformer.ts
koonopek: We have defined it somewhere :)

packages/integration/integration-test-core.sh
cehali: Change in permissions, I changed it in order to run it locally

packages/oracle-node/src/aggregated-price-handlers/DataPackageBroadcastPerformer.ts
cehali: Done

packages/integration/integration-test-core.sh
koonopek: thanks

packages/oracle-node/src/fetchers/BaseFetcher.ts
rpiszczatowski: The effect in this case is the same, but in case of exclamation mark it is obvious that your intention is to say: I know this value cannot be null and I promise it. This is an idiomatic way of saying this. In your code, on the other hand reader has to understand that pricePair.value is `Decimal?` and you just want to drop this `?`.

rpiszczatowski (approved):

PR #769, opened 2023-06-23
feat: create utils packages start porting functionalities

koonopek:


packages/utils/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * as SafeNumber from "./ISafeNumber";
```
koonopek, line 1: This is awsome that you can create Namespaces like this :)

packages/cache-service/package.json
```diff
@@ -44,7 +44,8 @@
     "reflect-metadata": "^0.1.13",
     "rimraf": "^3.0.2",
     "rxjs": "^7.2.0",
-    "ts-jest": "27.0.3"
+    "ts-jest": "27.0.3",
+    "redstone-utils": "*"
```
cehali, line 48: We should discuss how we want to pass internal packages dependencies, via reference or `package.json`. Maybe let's discuss it on daily tomorrow

koonopek: I pass it in both ways If i pass it only via tsconfig reference, then typescript is screaming at me

packages/utils/README.md
```diff
@@ -0,0 +1 @@
+# Redstone Utils
```
hatskier, line 1: Can you please add few more words about this module?

packages/utils/src/Math/index.ts
```diff
@@ -0,0 +1,34 @@
+import Decimal from "decimal.js";
+import { ISafeNumber, createSafeNumber } from "../ISafeNumber";
+import * as ISafeNumberMath from "../ISafeNumber";
+
+export type ConvertibleToISafeNumber = number | string | Decimal | ISafeNumber;
```
hatskier, line 5: Why did you decide to name the `Math` folder with the capital letter?

packages/cache-service/package.json
hatskier: It should be possible to pass to all modules that are not released as NPM packages using only tsconfig reference. We had it before e.g. with sdk or evm-connector

koonopek: You are right, I don't know I think that you have to first build it and then it works (that is why I though it doesnt work)

packages/utils/src/Math/index.ts
koonopek: Dont remember, renamed to `math`

packages/utils/README.md
koonopek: Yep!

packages/utils/src/ISafeNumber/ISafeNumberMath.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+  const result = calculateSum(numbers).div(numbers.length);
+
+  return result;
+};
+
+export const calculateDeviationPercent = (args: {
+  prevValue: ISafeNumber;
+  currValue: ISafeNumber;
+}) => {
```
rpiszczatowski, line 19: What is the advantage of having `args` object here as a parameter instead of two numbers?

koonopek: I don't know. It was like that i didnt wanted to change every usage.

packages/utils/src/ISafeNumber/ISafeNumberMath.ts
```diff
@@ -0,0 +1,45 @@
...
+  const { prevValue: measuredValue, currValue: trueValue } = args;
```
rpiszczatowski, line 20: I am not sure about this naming convention.

To me, we normally have some previous/old value and some current/new value and we want to calculate how the NEW value is different from the OLD value, i.e. abs(new-old)/old*100.

I also wonder why you renamed prev to measured and current to true. I would go the other way around if anything.

Maybe the name BASE would be good for the value we divide by. What do you think?

packages/utils/src/ISafeNumber/ISafeNumberMath.ts
rpiszczatowski: I'd say if you agree with my opinion we should change occurrences but if to you this API looks ok, then let's leave it as is.

koonopek: I don't have opinion on that, dont see thr value in change

packages/utils/src/ISafeNumber/ISafeNumberMath.ts
koonopek: Hmm I actually did the other way around I renamed measured -> prev and true -> current.
Base is okey but how to name other value then :)

koonopek: Automatic rename left old names just in this one place

packages/sdk/package.json
```diff
@@ -15,7 +15,8 @@
   "dependencies": {
     "axios": "^0.27.2",
     "redstone-oracles-smartweave-contracts": "*",
-    "redstone-protocol": "*"
+    "redstone-protocol": "*",
+    "redstone-utils": "*"
```
cehali, line 19: Why do we leave `redstone-utils` only in sdk?

packages/utils/tsconfig.build.json
```diff
@@ -0,0 +1,18 @@
+{
```
cehali, line 1: Do you know if we need this file? I think it is used in scenarios when we want to extend `tsconfig.json` but in this small package we probably don't need this

koonopek: Yes! Will remove

packages/sdk/package.json
koonopek: My bad fixed

PR #773, opened 2023-06-27
feat: refactor seralizing safeNumbers in signers and add tests

koonopek:
* Implemented missing tests
* Narrowed types (was `any`) thus typescript could not catch very obvious errors at compilation time (now it accepts only numbers but can be easily extended to string)


cehali:
I am not sure why we need this change, especially removing `SafeNumber` from deserialization. We plan to pass through the node process the price value as `SafeNumber` (underhood decimals.js)

koonopek:
> I am not sure why we need this change, especially removing `SafeNumber` from deserialization. We plan to pass through the node process the price value as `SafeNumber` (underhood decimals.js). Adding a new test is cool though 😄

We do not remove the serialization of ISafeNumber. Instead, let's perform it earlier in the "prepareForSigning" function. This approach is much cleaner since it guarantees that when we pass something to the "sign(xxx)" function, xxx won't be implicitly converted. We can be certain that what we pass will be signed. Previously, when we passed ISafeNumber, we had to read the code to see how it would be serialized or casted, which could have been confusing.

cehali (approved):

cehali (approved):
LGTM

hatskier (approved):

PR #779, opened 2023-06-28
feat: make redstone payload parser operate on DataPoint instead of NumericDataPoint

koonopek:


packages/cache-service/test/e2e/data-packages.spec.ts
```diff
@@ -420,10 +423,14 @@ describe("Data packages (e2e)", () => {
...
     );
 
-    const dataPoints: any[] = signedDataPackage.dataPackage.dataPoints;
+    const dataPoints: DataPoint[] = signedDataPackage.dataPackage.dataPoints;
     expect(dataPoints.length).toBe(2);
     expect(
-      dataPoints.map((dataPoint) => dataPoint.numericDataPointArgs)
+      dataPoints.map(
+        (dataPoint) =>
+          (convertDataPointToNumericDataPoint(dataPoint) as any)
```
cehali, line 431: Can't we cast to `NumericDataPoint`

packages/protocol/src/redstone-payload/RedstonePayloadParser.ts
```diff
@@ -16,6 +15,7 @@ import { DataPackage } from "../data-package/DataPackage";
 import { SignedDataPackage } from "../data-package/SignedDataPackage";
 import { DataPoint } from "../data-point/DataPoint";
 import { NumericDataPoint } from "../data-point/NumericDataPoint";
+import { BigNumber } from "ethers";
```
cehali, line 18: Why moved down there? 😄

packages/protocol/src/redstone-payload/RedstonePayloadParser.ts
```diff
@@ -177,3 +176,12 @@ export class RedstonePayloadParser {
     return this.bytesData.slice(start, end);
   }
 }
+
+export const convertDataPointToNumericDataPoint = (
+  dataPoint: DataPoint,
+  decimals = 8
```
cehali, line 182: Maybe we should pass decimals to `NumericDataPoint` ?

hatskier, line 182: Can we replace `8` with a constant (AFAIR, we have it somewhere)?

packages/cache-service/test/e2e/data-packages.spec.ts
koonopek: Nope, cause we are accessing numerDataPointArgs which is private field (this was the case before implementation too)

packages/protocol/src/redstone-payload/RedstonePayloadParser.ts
koonopek: Hmm probably removed this dependency and then had to import it again

packages/cache-service/test/e2e/data-packages.spec.ts
koonopek: Dont it another way, no casting at all! Thanks for comment

packages/protocol/src/redstone-payload/RedstonePayloadParser.ts
koonopek: This is not so obvious because asserts can fail then :(
NumerDataPoint actually can  have decimals = undefined and we have to have in mind if he do so it is = 8 (i don't like this logic too)
I have changed it to save decimals if we pass them :)

cehali (approved):

PR #782, opened 2023-06-28
feat: add ALUSD token

cehali:


packages/oracle-node/test/fetchers/avalanche-evm.spec.ts
```diff
@@ -412,41 +412,34 @@ describe("Avalanche EVM fetcher", () => {
     });
   });
 
-  describe("Curve Token", () => {
+  describe.only("Curve Token - crvUSDBTCETH", () => {
```
koonopek, line 415: Remove `.only`

packages/oracle-node/src/fetchers/evm-chain/shared/request-handlers/CurveRequestHandlers.ts
```diff
@@ -0,0 +1,111 @@
...
+interface TokenContractDetails {
+  erc20Address: string;
+  abi: string;
+  erc20abi: string;
+  poolAddress: string;
+  tokens: Tokens[];
+}
+type CurveTokensContractsDetails = Record<string, TokenContractDetails>;
+
+export class CurveRequestHandlers implements IEvmRequestHandlers {
```
koonopek, line 26: Why are we not using `curverFetcher` here? I think we are duplicating.

packages/oracle-node/test/fetchers/avalanche-evm.spec.ts
cehali: Done

packages/oracle-node/src/fetchers/evm-chain/shared/request-handlers/CurveRequestHandlers.ts
cehali: Not really. We use pools with multiple tokens and f.g. we calculate the price of `ALUSD` from the `ALUSD-crvFRAX` pool. We can calculate a ratio of ALUSD in terms of `crvFRAX` but we don't know the price of `crvFRAX`. Then we need to calculate `crvFRAX` based on balances of tokens inside the pool. Again we cannot use a simple curve fetcher as it could calculate the price of one token in terms of another, in the case of `crvFRAX` it would be `USDC` to `FRAX` or another way around. Also, we use 3 token pools like `3Crv` which has `DAI`, `USDC` and `USDT`

packages/oracle-node/test/fetchers/avalanche-evm.spec.ts
koonopek: Remove .only ;)

koonopek (approved):

packages/oracle-node/src/fetchers/evm-chain/shared/request-handlers/CurveRequestHandlers.ts
koonopek: I'm not sure if i understand your message correctly.
But agree that we can't use 3 token pools with CurveFetcher.
However I think we still can use it for crv-frax pair

packages/oracle-node/src/fetchers/evm-chain/ethereum/evm-fetcher/sources/curve-lp-tokens/curveTokensContractsDetails.ts
```diff
@@ -0,0 +1,26 @@
+import abi from "../../../../shared/abis/CurveToken.abi.json";
+import erc20abi from "../../../../shared/abis/CurveErc20.abi.json";
+
+export const curveTokensContractsDetails = {
+  crvFRAX: {
+    erc20Address: "0x3175Df0976dFA876431C2E9eE6Bc45b65d3473CC",
+    erc20abi,
```
koonopek, line 7: I think there is no value in repeating here erc20abi

packages/oracle-node/src/fetchers/evm-chain/shared/request-handlers/CurveRequestHandlers.ts
cehali: In my opinion, you can use it to calculate the price of CRV based on FRAX price or FRAX price based on CRV price but not crvFRAX token itself. What is your idea of using `CurveFetcher` to get the price of crvFRAX? Maybe I am not aware of something here.

packages/oracle-node/src/fetchers/evm-chain/ethereum/evm-fetcher/sources/curve-lp-tokens/curveTokensContractsDetails.ts
cehali: Refactored

packages/oracle-node/src/fetchers/evm-chain/avalanche/evm-fetcher/sources/curve-lp-tokens/curveTokensContractsDetails.ts
```diff
@@ -2,11 +2,11 @@ import abi from "../../../../shared/abis/CurveToken.abi.json";
 import erc20abi from "../../../../shared/abis/CurveErc20.abi.json";
 
 export const curveTokensContractsDetails = {
+  erc20abi,
```
koonopek, line 5: I meant that we can have one `erc20` abi for alllllll contracts, don't have to divde it to CurverErc20 or MavericErc20, abi will still be the same ;)

koonopek (approved):

PR #786, opened 2023-06-30
chore: sanitize AAVE sources

cehali:


packages/oracle-node/manifests/dev/main-wide-support.json
```diff
@@ -683,30 +683,30 @@
...
-        "chainlink",
+        "bybit",
         "coinbaseprime",
         "coinbasepro",
         "coingecko",
         "currencycom",
-        "dia",
         "hitbtc",
         "huobipro",
-        "kaiko",
```
koonopek, line 701: Why do we remove kaiko?

koonopek: Hmm why do remove some of the sources?

koonopek: image
For me not checked sources from picture above have very deviated prices

koonopek (dismissed):
Please answer one comment, just for my better understanding

packages/oracle-node/manifests/dev/main-wide-support.json
cehali: Chainlink and Cex are removed in this PR. Binance is not working correctly in the US, we will move our infra to Europe. Bittrex was working correctly when applying those changes, we need to monitor it. Kaiko should be there, my bad, will fix it in a moment. The rest of the removed sources were not providing bad values or they are other oracles, which we don't want to use in more prod env.

koonopek (dismissed):

PR #788, opened 2023-06-30
fix: issue with single data point in twelve data fetcher

cehali:


rpiszczatowski (dismissed):
test approve

packages/oracle-node/src/fetchers/twelve-data/TwelveDataFetcher.ts
```diff
@@ -46,9 +46,19 @@ export class TwelveDataFetcher extends BaseFetcher {
...
+  extractPrices(
+    response: AxiosResponse<TwelveDataResponse>,
+    ids: string[]
+  ): PricesObj {
     const twelveDataResponse = response.data;
 
+    // If we require only 1 data feed price, response would be { price: xxx }
+    if (ids.length === 1) {
+      const id = ids[0];
+      const value = (twelveDataResponse as unknown as { price: number }).price;
```
koonopek, line 58: You can type it as `TwelveDataResponse[string]` instead of `{price: number}`

koonopek (approved):

PR #789, opened 2023-06-30
feat: add LUSD token

cehali:


koonopek (dismissed):

koonopek (approved):

packages/oracle-node/src/fetchers/evm-chain/uniswap-v3-on-chain/all-uniswap-v3-fetchers.ts
```diff
@@ -16,4 +18,14 @@ for (const [fetcherName, fetcherConfig] of Object.entries(
   );
 }
 
+for (const [fetcherName, fetcherConfig] of Object.entries(
```
koonopek, line 21: I think that pattern when you provide a provider in configuration is better than this one ;)

koonopek (dismissed):

packages/oracle-node/src/fetchers/evm-chain/uniswap-v3-on-chain/all-uniswap-v3-fetchers.ts
cehali: The problem is that we have a first-time situation when we get the price of the same token inside the same fetcher from different chains. If we try to pass the provider in the configuration we will have an object with two keys `LUSD`

koonopek (approved):

PR #795, opened 2023-07-03
feat: implemented a mechanism for getting payload for manual usage

hatskier:
This PR implements a mechanism of passing redstone payload as the last argument of a contract function. It can be way more convenient in many cases, such as:
- Using multisig, where it's difficult to wrap contract calls
- Passing payload across multiple contracts (now it is possible without messy CalldataExtractor and manual results parsing)


packages/evm-connector/src/wrappers/BaseWrapper.ts
```diff
@@ -16,9 +17,52 @@ interface OverwriteFunctionArgs {
...
+    // Redstone payload can be passed as the last argument of the contract function
+    // But it needs to have a length that is a multiplicity of 32, otherwise zeros
+    // will be padded right and contract will revert with `CalldataMustHaveValidPayload`
+    if (shouldBePreparedForManualUsage) {
+      const originalPayload = RedstonePayload.prepare(
+        signedDataPackages,
+        unsignedMetadata
+      );
+      // Calculating the number of bytes in the hex representation of payload
+      const originalPayloadLength = originalPayload.length / 2;
```
jakub-wojciechowski, line 53: What if originalPayload is even (not divisible by 2)? Is it possible?

packages/evm-connector/src/wrappers/BaseWrapper.ts
```diff
@@ -16,9 +17,52 @@ interface OverwriteFunctionArgs {
...
+
+      // Number of bytes that we want to add to unsigned metadata so that
+      // payload byte size becomes a multiplicity of 32
+      const bytesToAdd =
+        Math.ceil(originalPayloadLength / 32) * 32 - originalPayloadLength;
+
+      // Adding spaces to the end of the metadata string, each space uses one byte
+      unsignedMetadata += " ".repeat(bytesToAdd);
```
jakub-wojciechowski, line 61: How about any other character, spaces could be trimmed somehow by printing libraries and cause unexpected issues

packages/evm-connector/src/wrappers/BaseWrapper.ts
```diff
@@ -16,9 +17,52 @@ interface OverwriteFunctionArgs {
...
+      const originalPayload = RedstonePayload.prepare(
+        signedDataPackages,
+        unsignedMetadata
+      );
+      // Calculating the number of bytes in the hex representation of payload
+      const originalPayloadLength = originalPayload.length / 2;
+
+      // Number of bytes that we want to add to unsigned metadata so that
+      // payload byte size becomes a multiplicity of 32
+      const bytesToAdd =
```
jakub-wojciechowski, line 57: Hmm, how about
 32 - (originalPayloadLength % 32)
, for me it's easier to read

packages/evm-connector/test/data-packages-wrapper/data-packages-wrapper.test.ts
```diff
@@ -44,6 +39,23 @@ describe("DataPackagesWrapper", () => {
...
+    const wrappedContract =
+      WrapperBuilder.wrap(contract).usingDataPackages(dataPackagesResponse);
+
+    const tx = await wrappedContract.save2ValuesInStorage(dataFeedIds);
+    await tx.wait();
+    await checkExpectedValues();
+  });
+
+  it("Should work properly with manual payload", async () => {
+    const dataPackagesResponse = getValidDataPackagesResponse();
```
jakub-wojciechowski, line 51: As discussed let's move refactoring that to the next phase

packages/evm-connector/src/wrappers/BaseWrapper.ts
hatskier: No, it's not possible, because it's a hex string with an integer number of bytes, so if bytes size of the payload is equal `n`, then `originalPayload.length` is equal to `2 * n`

packages/evm-connector/src/wrappers/BaseWrapper.ts
hatskier: I just selected the most neutral char, which consumes 1 byte with UTF-8 representation. I think it's better to have space than e.g. 0, because sometimes we may want to store a number in the end of the metadata (e.g. version, or a timestamp)

hatskier: I will replace it with underscore

packages/evm-connector/test/mock-wrapper/numbers-many-data-feeds.test.ts
```diff
@@ -77,6 +84,18 @@ describe("SampleRedstoneConsumerNumericMockManyDataFeeds", function () {
     await testShouldPass(mockNumericPackages, ["BTC", "ETH"]);
   });
 
+  it("Should work properly with manual payload", async () => {
+    const mockWrapper = new MockWrapper(mockNumericPackages);
+    const payload = await mockWrapper.getRedstonePayloadForManualUsage();
+    const dataFeedIds: ("ETH" | "BTC")[] = ["ETH", "BTC"];
```
koonopek, line 90: `"[ETH","BTC"]" as const` the same but more idomatic :)

packages/evm-connector/src/wrappers/BaseWrapper.ts
```diff
@@ -16,9 +17,52 @@ interface OverwriteFunctionArgs {
...
+    const payloadWithoutZeroExPrefix = await this.getBytesDataForAppending(
+      params,
+      shouldBePreparedForManualUsage
+    );
+    return "0x" + payloadWithoutZeroExPrefix;
+  }
+
+  async getBytesDataForAppending(
+    params?: ParamsForDryRunVerification,
+    shouldBePreparedForManualUsage = false
```
koonopek, line 39: It would be much cleaner if there was seperate private function for this, not this param IMO. API will be cleaner end dev don't have think wtf is this param :)

koonopek (approved):
As I said earlier I think that this should be in seperate class/function

koonopek (commented):
I'm okey with merging it now, but would refactor to seperate class/function before releasing. I think we could avoid chaning BaseWrapper and method extending after it compltly. And just add new feature because this is what we are essentaily doing here

packages/evm-connector/src/wrappers/BaseWrapper.ts
```diff
@@ -16,9 +17,52 @@ interface OverwriteFunctionArgs {
...
+  ): Promise<string> {
+    const signedDataPackages = await this.getDataPackagesForPayload(params);
+    let unsignedMetadata = this.getUnsignedMetadata();
+
+    // Redstone payload can be passed as the last argument of the contract function
+    // But it needs to have a length that is a multiplicity of 32, otherwise zeros
+    // will be padded right and contract will revert with `CalldataMustHaveValidPayload`
+    if (shouldBePreparedForManualUsage) {
```
cehali, line 47: We could extract it to separate function with all comments

cehali (dismissed):
Don't we want to create a separate wrapper for it?

cehali (commented):
Don't we want to create a separate wrapper for it?

Answer (Alex): We want to have ability in different wrappers to get the manual payload (e.g. payload from mock wrapper for tests and payload from data packages wrapper to use in prod)

packages/evm-connector/src/wrappers/BaseWrapper.ts
hatskier: I will apply suggested refactoring in a separate PR

packages/evm-connector/src/wrappers/BaseWrapper.ts
hatskier: Will apply this change in a separate PR

PR #797, opened 2023-07-04
feat: adjust relayer to standard data point

cehali:


packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
```diff
@@ -22,12 +22,12 @@ export const valueDeviationCondition = (
...
     for (const { dataPackage } of dataPackages[dataFeedId]) {
       for (const dataPoint of dataPackage.dataPoints) {
-        const dataPointObj = dataPoint.toObj() as INumericDataPoint;
+        const dataPointObj = dataPoint.toObj() as IStandardDataPoint;
 
         const valueFromContractAsDecimal = Number(
           formatUnits(
             valueFromContract.toString(),
-            dataPointObj.decimals ?? DEFAULT_DECIMALS
+            dataPointObj.metadata?.decimals ?? DEFAULT_DECIMALS
```
koonopek, line 30: As @hatskier  suggested earlier we don't have to care about decimals. We can compare strings like:
```js
const valueFromDataPoint = BigNumber.from(base64.decode(dataPointObj.value);
const valueFromContract = contract.getValue(); // it is already BigNumber

// comparison
valueFromDataPoint.sub(valueFromContract)
```

koonopek (changes requested):

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
cehali: Currently we need decimals if we want to be backwards compatible as you suggested

PR #798, opened 2023-07-04
chore: update STETH sources

cehali:


packages/oracle-node/src/fetchers/ccxt/CcxtFetcher.ts
```diff
@@ -80,4 +81,21 @@ export class CcxtFetcher extends BaseFetcher {
     const stableCoinSymbol = pairSymbol.slice(-4);
     return getLastPrice(stableCoinSymbol)?.value;
   }
+
+  async handleRequestsForBybit(ids: string[]) {
```
cehali, line 85: This is required because "old" and "new" tokens are treated differently and handled by different versions of bybit API. Without it, we will only get tokens with `symbol-to-id` values with `USDT:USDT` in the end. If we try to use `STETH/USDT:USDT` we will receive error than this symbol doesn't exist

hatskier: I would like to solve it differently. First of all, I believe our mechanism with symbol to id custom mapping should solve all potential cases and shouldn't require functions like this. But even if for some reason it is not sufficient, I don't think it's a good idea to make async calls for fetching tickers, since we can do it once and save in files (like we do with symbol-to-id)

hatskier (changes requested):

packages/oracle-node/src/fetchers/ccxt/CcxtFetcher.ts
cehali: I don't understand tbh. We always make async call to fetch tickers https://github.com/redstone-finance/redstone-oracles-monorepo/blob/6aaf5fa5c761449e41b20075d16f5e97e4af62ae/packages/oracle-node/src/fetchers/ccxt/CcxtFetcher.ts#L52. 

Maybe you misunderstand what `fetchTickers` do here, it fetches price for ids. 

It is a second attempt to integrate a new version of bybit and the previous solution with a separate fetcher name `bybit-v5` was rejected. Then it wasn't important but now we need to integrate it. 
If you have any idea how to solve this case with custom symbol to id mapping I am gladly to hear that I am buying you a beer if you find it 🍺  😄

hatskier (approved):

PR #801, opened 2023-07-05
feat: add MultiBlockDecorator

koonopek:


packages/oracle-node/src/fetchers/MultiBlockDecorator.ts
```diff
@@ -0,0 +1,95 @@
...
+  intervalLength: number;
+};
+
+export function decorateWithMultiBlock<T extends BaseFetcher>(
+  fetcher: T,
+  provider: providers.Provider,
+  multiblockConfig: MultiBlockConfig
+): T {
+  const oldGetName = fetcher.getName.bind(fetcher);
+  (fetcher as any).getName = () => {
```
rpiszczatowski, line 17: Why is this type assertion needed?

packages/oracle-node/src/fetchers/MultiBlockDecorator.ts
```diff
@@ -0,0 +1,95 @@
...
+    );
+
+    const responsePerBlock = (await Promise.all(
+      blockSequence.map((blockTag) =>
+        oldFetchAll(tokens, {
+          ...opts,
+          blockTag,
+        })
+      )
+    )) as PriceDataFetched[][];
```
rpiszczatowski, line 43: Why this type assertion is needed?

packages/oracle-node/src/fetchers/MultiBlockDecorator.ts
```diff
@@ -0,0 +1,95 @@
...
+) {
+  const sequence = [start];
+
+  if (intervalLength === 1) {
+    return sequence;
+  }
+
+  const stepsCount = Math.ceil(intervalLength / step);
+
+  const scaledSecond = Math.floor((start - 1) / step) * step;
```
rpiszczatowski, line 87: Why is this variable called `scaledSecond`?

packages/oracle-node/src/fetchers/MultiBlockDecorator.ts
```diff
@@ -0,0 +1,95 @@
...
+    return Object.entries(groupedByToken).map(([symbol, values]) => ({
+      symbol,
+      value: SafeNumber.getMedian(values).unsafeToNumber(),
+    }));
+  };
+
+  return fetcher;
+}
+
+export function generateRoundedStepSequence(
```
rpiszczatowski, line 74: Why not making this function an arrow function? Similarly the function above.

packages/oracle-node/src/fetchers/MultiBlockDecorator.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        groupedByToken[token.symbol].push(
+          SafeNumber.createSafeNumber(token.value)
+        );
+      }
+    }
+
+    return Object.entries(groupedByToken).map(([symbol, values]) => ({
+      symbol,
+      value: SafeNumber.getMedian(values).unsafeToNumber(),
```
rpiszczatowski, line 67: Do we want a median here? If let's say we have 10 values and the price is raising all the time, we are 5 values behind (which I assume translates to `5*step` seconds or blocks).

koonopek: Yes, we want a median here. That is intentional.

packages/oracle-node/src/fetchers/MultiBlockDecorator.ts
koonopek: Renamed to secondValue

packages/oracle-node/src/fetchers/MultiBlockDecorator.ts
koonopek: Don't know. Changed.

packages/oracle-node/src/fetchers/MultiBlockDecorator.ts
koonopek: It is no longer needed. Good catch!

rpiszczatowski (approved):

packages/oracle-node/src/fetchers/MultiBlockDecorator.ts
```diff
@@ -0,0 +1,95 @@
...
+  fetcher: T,
+  provider: providers.Provider,
+  multiblockConfig: MultiBlockConfig
+): T {
+  const oldGetName = fetcher.getName.bind(fetcher);
+  fetcher.getName = () => {
+    return `${oldGetName()}-multi-block`;
+  };
+
+  const oldFetchAll = fetcher.fetchAll.bind(fetcher);
```
koonopek, line 21: To cache just add RedstoneCommon.memoize here

PR #810, opened 2023-07-05
feat: adapt Transaction Delivery man to arb network

koonopek:


packages/on-chain-relayer/src/config.ts
```diff
@@ -48,6 +48,7 @@ export const config = Object.freeze({
   expectedTxDeliveryTimeInMS: Number(
     getFromEnv("EXPECTED_TX_DELIVERY_TIME_IN_MS", false)
   ),
+  isArbitrumNetwork: getFromEnv("IS_ARBITRUM_NETWORK", true) === "true",
```
koonopek, line 51: Maybe this name is a little bit stupid, however, my priority was to minimize errors when someone will be configuring on-chain-relayer

packages/rpc-providers/src/TransactionDeliveryMan.ts
```diff
@@ -50,15 +60,18 @@ type TransactionDeliveryManOpts = {
...
 const unsafeBnToNumber = (bn: BigNumber) => Number(bn.toString());
 
-const getEthFeeFromGasOracle: GasOracleFn = async (apiKey: string = "") => {
+const getEthFeeFromGasOracle: GasOracleFn = async (
+  opts: TransactionDeliveryManOpts
+) => {
   const response = // rate limit is 5 seconds
     (
       await fetchWithCache<any>(
-        `https://api.etherscan.io/api?module=gastracker&action=gasoracle&apikey=${apiKey}`,
```
hatskier, line 61: Why do we remove the apiKey?

packages/rpc-providers/src/TransactionDeliveryMan.ts
```diff
@@ -161,8 +176,13 @@ export class TransactionDeliveryMan {
 
   private isUnderpricedError(e: any) {
     return (
+      // RPC errors sucks most of the time, thus we can really relay on them
```
hatskier, line 179: can -> can not?

hatskier: relay -> rely?

packages/rpc-providers/src/TransactionDeliveryMan.ts
```diff
@@ -240,21 +272,42 @@ export class TransactionDeliveryMan {
...
     );
     const maxFeePerGas = Math.round(
-      currentFees.maxFeePerGas * this.opts.priorityFeePerGasMultiplier
-    );
-
-    this.opts.logger(
-      `Scaling fees maxFeePerGas to ${maxFeePerGas} and maxPriorityFeePerGas to ${maxPriorityFeePerGas}`
+      currentFees.maxFeePerGas * this.opts.multiplier
     );
+    const gasLimit = this.opts.twoDimensionFees
```
hatskier, line 299: Do I understand correctly that if `twoDimensionFees` is `true`, then we increase both fee and gasLimit in the same way (using the same multiplier)?

hatskier (dismissed):

packages/rpc-providers/src/TransactionDeliveryMan.ts
koonopek: Correct, if we observe that multiplier for gaszlimit.shoul.be Independent,we can always change it

packages/rpc-providers/src/TransactionDeliveryMan.ts
koonopek: We are using cache.so.we will.never get into api reqursts limits (we werent using it anyway)

packages/rpc-providers/src/TransactionDeliveryMan.ts
koonopek: Changed to seperate

packages/rpc-providers/src/TransactionDeliveryMan.ts
hatskier: Should we have the `&apikey=` then?

hatskier (approved):

packages/rpc-providers/src/TransactionDeliveryMan.ts
koonopek: Will fix in next PR :)

PR #811, opened 2023-07-05
feat: add fallback-deviation trigger function for on-chain-relayer

Lukasz2891:


Lukasz2891:
Waiting for [robal/relayer-manifests](https://github.com/redstone-finance/redstone-oracles-monorepo/tree/robal/relayer-manifests) to be rebased/merged with main.

packages/on-chain-relayer/relayer-manifests/fallback.json
```diff
@@ -0,0 +1,16 @@
+{
```
jakub-wojciechowski, line 1: how about naming it goerli-eth-fallback.json

packages/on-chain-relayer/src/core/update-conditions/should-update.ts
```diff
@@ -27,11 +34,28 @@ export const shouldUpdate = (context: Context, config: RelayerConfig): Condition
...
+const checkConditionByName = (
+  name: ConditionCheckNames,
+  context: Context,
+  config: RelayerConfig
+) => {
+  switch (name) {
+    case "time":
+      return timeUpdateCondition(context.lastUpdateTimestamp, config);
+
+    case "value-deviation":
```
rpiszczatowski, line 46: Having both "value-deviation" and "fallback-deviation" conditions enabled is probably a configuration error. Could you add a check for this situation?

Lukasz2891: why :) there is A || (A && B) so that means and A && B is just not needed in some cases but doesn't lead to an error.

packages/on-chain-relayer/src/FilesystemConfigProvider.ts
```diff
@@ -31,15 +31,21 @@ function readJSON<T>(path: string): T {
...
   const manifest = readJSON<OnChainRelayerManifest>(mainfestPath);
-  const { timeSinceLastUpdateInMilliseconds, deviationPercentage } =
-    manifest.updateTriggers;
-  let updateConditions = [] as ConditionChecksNames[];
+  const {
+    timeSinceLastUpdateInMilliseconds,
+    deviationPercentage,
+    fallbackDeviationCheckOffsetInMinutes,
+  } = manifest.updateTriggers;
+  let updateConditions = [] as ConditionCheckNames[];
```
hatskier, line 39: In my opinion, `offsetInMinutesForFallback` should not be a part of manifest. Manifest is common for primary and fallback relayers, but the fallback relayer can just have this param in its env variables (in case of our relayer) or in js config (in case of gelato relayer)

hatskier: It automatically requires support of the fallback mode for both `time` and `value-deviation` updateConditions

Lukasz2891: but that's not only the fallback-relayer, but just a trigger-function implementing the functionality which needs to be configured. So I suppose it'd be better to have it configured in the same way as the others.

packages/on-chain-relayer/src/core/contract-interactions/update-prices.ts
```diff
@@ -29,38 +21,11 @@ const getDeliveryMan = () => {
...
-    });
-    console.log(`Update prices tx sent: ${updateTx.hash}`);
-    await updateTx.wait();
-    console.log(`Successfully updated prices: ${updateTx.hash}`);
-  }
+export const updatePrices = async (updatePricesArgs: UpdatePricesArgs) => {
+  const updateTx = await updatePriceInAdapterContract(updatePricesArgs);
+  console.log(`Update prices tx sent: ${updateTx.hash}`);
+  await updateTx.wait();
+  console.log(`Successfully updated prices: ${updateTx.hash}`);
```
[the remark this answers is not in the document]
Lukasz2891: Just copied from the previous code :) maybe we should ;)

packages/on-chain-relayer/src/FilesystemConfigProvider.ts
Lukasz2891: Done

packages/on-chain-relayer/src/core/update-conditions/should-update.ts
rpiszczatowski: The problem is not in the predicate but in the fact that if somebody sets both conditions in the gelato fallback relayer then the consequences are that the fallback relayer runs too often which means the cost for us. Also:
- it is simple to check and
- it is hard to figure out in runtime because the relayer basically works as expected and
- setting both conditions doesn't simply make much sense and if actually done probably somebody just did some kind of copy-paste error

Lukasz2891: will be changed in #837

rpiszczatowski (approved):

PR #814, opened 2023-07-06
feat: implement gelato-based fallback relayer

Lukasz2891:


packages/fallback-relayer/contracts/GelatoPriceFeedsAdapter.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity ^0.8.4;
+
+import "redstone-on-chain-relayer/contracts/price-feeds/without-rounds/PriceFeedsAdapterWithoutRounds.sol";
+
+contract GelatoPriceFeedsAdapter is PriceFeedsAdapterWithoutRounds {
```
jakub-wojciechowski, line 7: Wouldn't it better to call it EthMainDemoPriceFeedsAdapter, it's not specific to Geltato but to our feeds config

packages/fallback-relayer/hardhat.config.ts
```diff
@@ -0,0 +1,133 @@
...
+  verify: {
+    etherscan: {
+      apiKey: ETHERSCAN_KEY ? ETHERSCAN_KEY : "",
+    },
+  },
+
+  solidity: {
+    compilers: [
+      {
+        version: "0.8.17",
```
jakub-wojciechowski, line 119: Wouldn't 0.8.4 be enough, or does it depend on Gelato?

packages/fallback-relayer/web3-functions/redstone/userArgs.json
```diff
@@ -0,0 +1,13 @@
+{
+  "adapterContractAddress": "0x89b6137B9c09677BC77dc53B9dc564E2B3eA2d43",
+  "dataServiceId": "redstone-main-demo",
+  "uniqueSignersCount": 1,
+  "dataFeeds": ["ETH"],
+  "updateConditions": ["fallback-deviation"],
+  "updatePriceInterval": 140000,
+  "minDeviationPermillage": 5,
```
jakub-wojciechowski, line 8: Why cannot we keep is as percentage "0.5". Are we restricted to ints only?

packages/fallback-relayer/contracts/GelatoPriceFeedsAdapter.sol
```diff
@@ -0,0 +1,30 @@
...
+  override
+  returns (uint8)
+  {
+    if (signerAddress == 0x0C39486f770B26F5527BBBf942726537986Cd7eb) {
+      return 0;
+    } else {
+      revert SignerNotAuthorised(signerAddress);
+    }
+  }
+}
```
jakub-wojciechowski, line 30: Don't we need to extend any Gelato base contracts?
Would it be possible to get an address of a Gelato account?

packages/fallback-relayer/hardhat.config.ts
hatskier: Our adapter contracts use 0.8.14

hatskier (commented):
Can we call it `gelato-relayer`, not `fallback-relayer`?

packages/fallback-relayer/contracts/GelatoPriceFeedsAdapter.sol
Lukasz2891: I'll change the name to "Sample" or sth

Lukasz2891:
> 

Why 'gelato' :)? It's a fallback relayer implemented on gelato ;p

packages/fallback-relayer/contracts/GelatoPriceFeedsAdapter.sol
Lukasz2891: It's just a sample contract

packages/fallback-relayer/web3-functions/redstone/userArgs.json
Lukasz2891: And also I don't think we should pass it as a string... but maybe

packages/fallback-relayer/hardhat.config.ts
Lukasz2891: gelato doesn't compile our contracts - just their ones, that was configured as above and I didn't change it. Probably that doesn't matter for gelato.

packages/fallback-relayer/contracts/GelatoPriceFeedsAdapter.sol
Lukasz2891: We don't need to extend gelato contracts, it's wrapped in another way - we return the whole transaction data to be sent to our contract by the web3f.

hatskier (approved):

PR #815, opened 2023-07-06
chore: update WSTETH sources

cehali:


koonopek (approved):
How is solved problem with maverick?

cehali:
@koonopek Depending if the data feed is token0 or token1 we calculate `ratio * price of underlying token` or `1 / ratio * price underlying token`

PR #820, opened 2023-07-07
feat: remove dry run from data service wrapper

cehali:


packages/evm-connector/test/data-service-wrapper/data-service-wrapper.test.ts
```diff
@@ -80,14 +80,6 @@ describe("DataServiceWrapper", () => {
       );
     });
 
-    it("Should properly execute with one invalid and one slower cache", async () => {
```
cehali, line 83: We need to delete it because without a dry-run we always take the first or fastest data packages

cehali:
Removing it due to Alex's request: https://github.com/orgs/redstone-finance/projects/5/views/9?pane=issue&itemId=32516477

hatskier (approved):

PR #823, opened 2023-07-07
refactor: switch non usd based fetcher to multi request fetcher

cehali:


packages/oracle-node/src/fetchers/non-usd-based/NonUsdBasedFetcher.ts
```diff
@@ -1,34 +1,30 @@
...
-      prices[id] = { mainTokenPrice, pairedTokenPrice };
+    const mainTokenPrice = getLastPrice(mainToken)?.value;
+    const pairedTokenPrice = getLastPrice(pairedToken)?.value;
+    if (!mainTokenPrice || !pairedTokenPrice) {
+      throw new Error(
+        `Cannot get last price from cache for: ${mainToken} or ${pairedToken}`
+      );
     }
-    return Promise.resolve(prices);
+    return Promise.resolve({ mainTokenPrice, pairedTokenPrice });
```
hatskier, line 27: I would prefer having `async` function and avoiding `Promise.resolve`, but it's just a matter of tase

hatskier (approved):

packages/oracle-node/src/fetchers/non-usd-based/NonUsdBasedFetcher.ts
cehali: Changed

hatskier (approved):

PR #831, opened 2023-07-07
chore: allow dry run to skip sources and tokens

rpiszczatowski:


.github/workflows/oracle-main-dry-run-test.yml
```diff
@@ -27,3 +27,5 @@ jobs:
           TWELVE_DATA_API_KEY: ${{ secrets.TWELVE_DATA_API_KEY }}
           KAIKO_API_KEY: ${{ secrets.KAIKO_API_KEY }}
           STLOUISFED_API_KEY: ${{ secrets.STLOUISFED_API_KEY }}
+          SKIPPED_SOURCES: '["binance", "binancecoinm", "binanceusdm"]'
```
hatskier, line 30: What is the reason of keeping these params in the env variables, not just in the dry-run tests code?

rpiszczatowski: The difference is that if one runs tests locally, Binance will be taken into account and this is good because Binance at the moment doesn't work if run from the US only which is the case of GitHub runners and not local runs.

PR #835, opened 2023-07-10
feat: add fee to Curve fetcher

cehali:


cehali:
https://discord.com/channels/786251205008949258/1055853648242016397/1126894885392355441

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -63,13 +63,13 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
   calculateSpotPrice(assetId: string, response: CurveFetcherResponse): number {
     const pairedTokenPrice = new Decimal(this.getPairedTokenPrice(assetId));
     const ratio = response.ratio;
-
-    return ratio.mul(pairedTokenPrice).toNumber();
+    const fee = this.poolsConfig[assetId].fee;
+    const ratioWithFee = ratio.add(fee);
```
rpiszczatowski, line 67: The analysis below assumes no slippage.
let dx - amount in, dy - amount out, f - fee, p - price (the real one), r - ratio from response (price that takes fee into account)

dy/dx=r - this is how ratio was calculated in makeRequest
dy=p*(1-f)*dx - this is how fee works: (1-f)*dx - amount that is actually swapped f*dx is a fee.
Dividing second equation by dx and substituting dy/dx=r in it we get:
r=p*(1-f) and so p = r / (1-f) which is the operation you should perform (instead of simple addition r + f)

rpiszczatowski (approved):

PR #838, opened 2023-07-11
feat: added time-fallback condition to on-chain-relayer

Lukasz2891:


packages/on-chain-relayer/.env.example
```diff
@@ -11,3 +11,6 @@ HISTORICAL_PACKAGES_DATA_SERVICE_ID=redstone-primary-prod
 
 # runs the 'fallback-deviation' trigger function when >0
 FALLBACK_DEVIATION_CHECK_OFFSET_IN_MINUTES=0
```
hatskier, line 13: I think it would be enough to have only one parameter `FALLBACK_OFFSET_IN_MINUTES`

hatskier (changes requested):

hatskier (approved):

PR #845, opened 2023-07-12
chore: refactoring of manual payload

hatskier:


cehali (approved):
LGTM

packages/evm-connector/src/wrappers/BaseWrapper.ts
```diff
@@ -17,38 +11,34 @@ interface OverwriteFunctionArgs {
...
-    // will be padded right and contract will revert with `CalldataMustHaveValidPayload`
-    if (shouldBePreparedForManualUsage) {
-      const originalPayload = RedstonePayload.prepare(
-        signedDataPackages,
-        unsignedMetadata
-      );
+    const originalPayload = RedstonePayload.prepare(
+      signedDataPackages,
+      unsignedMetadata
+    );
```
cehali, line 39: Why it is above `if (shouldBeMultipleOf32)`? We run  `prepare` function two times always

packages/evm-connector/src/wrappers/BaseWrapper.ts
```diff
@@ -17,38 +11,34 @@ interface OverwriteFunctionArgs {
...
-  async getRedstonePayloadForManualUsage(
-    params?: ParamsForDryRunVerification
-  ): Promise<string> {
-    const shouldBePreparedForManualUsage = true;
-    const payloadWithoutZeroExPrefix = await this.getBytesDataForAppending(
-      params,
-      shouldBePreparedForManualUsage
-    );
+  async getBytesDataForAppending(): Promise<string> {
+    const shouldBeMultipleOf32 = false;
```
cehali, line 19: I am not really a fan of it but I couldn't develop a better solution. The only thing I came up with is to pass it as an object to `prepareRedstonePayload` which looks then more like config than some random boolean argument

hatskier: Frankly, I don't see a value of passing an object with only one prop

packages/evm-connector/src/wrappers/BaseWrapper.ts
hatskier: Pushed a fix

cehali (approved):

cehali (approved):

PR #846, opened 2023-07-12
feat: add manual payload to data service wrapper

hatskier:


packages/evm-connector/test/data-service-wrapper/data-service-wrapper.test.ts
```diff
@@ -131,12 +158,26 @@ describe("DataServiceWrapper", () => {
...
     });
+
+    it("Should work with manual payload without passed params", async () => {
+      const wrapper = new DataServiceWrapper({
+        dataFeeds: ["ETH", "BTC"],
+      });
+
+      // We need to call `setContractForFetchingDefaultParams` to fetch default
+      // values for uniqueSignersThreshold, dataServiceId, and gatewayUrls
+      wrapper.setContractForFetchingDefaultParams(contract);
```
hatskier, line 176: Idea for improvement: maybe we can call this function inside `getRedstonePayloadForManualUsage`, but then we would require passing `contract` object as an argument there

cehali: I agree, it would be easier to understand 👍

hatskier: Added this change

cehali (approved):

PR #854, opened 2023-07-13
feat: integration tests in typescript

rpiszczatowski:


packages/integration/integration-test-core.ts
```diff
@@ -0,0 +1,22 @@
...
+  await common.waitForDataAndDisplayIt();
+  await common.buildEvmConnector();
+  await common.runWithLogPrefix("yarn", ["test", "test/monorepo-integration-tests/localhost-mock.test.ts"], "evm-connector");
+
+  process.exit();
+}
+
+common.configureCleanup(stopAll);
+
+main()
```
hatskier, line 22: Missing semicolon

packages/integration/integration-test-common.ts
```diff
@@ -0,0 +1,282 @@
...
+export const updateDotEnvFile = (varName: string, varValue: string) => {
+  copyAndReplace(
+    new RegExp(`^${varName}=.*$`, "gm"),
+    `${varName}=${varValue}`,
+    "./.env",
+    "./.env"
+  );
+};
+
+const configureLogging = (childProcess: ChildProcess, logPrefix: string) => {
```
hatskier, line 101: I would propose to rename this function to `addPrefixToProcessOutput` or `addPrefixToProcessLogs`

packages/integration/integration-test-common.ts
```diff
@@ -0,0 +1,282 @@
...
+
+export const lazilyInstallNPMDeps = async () => {
+  if (fs.existsSync("node_modules")) {
+    console.log("Node modules are already installed. Skipping...");
+  } else {
+    console.log("Installing NPM deps");
+    const result = spawn("yarn", {
+      stdio: ["ignore", "ignore", "pipe"],
+    });
+    configureLogging(result, "yarn install");
```
hatskier, line 35: I would call this vairable `childProcess` instead of `result`, as the name `result` can be confused with the process result

packages/integration/integration-test-common.ts
```diff
@@ -0,0 +1,282 @@
...
+
+export const installAndBuild = async () => {
+  await lazilyInstallNPMDeps();
+  await lazilyBuildTypescript();
+};
+
+export const waitForFile = async (filename: string) => {
+  while (!fs.existsSync(filename)) {
+    console.log(`${filename} is not present, waiting...`);
+    await sleep(5000);
```
hatskier, line 67: Can we please move 5000 to a constant here and in the `waitForUrl` function?

packages/integration/integration-test-common.ts
```diff
@@ -0,0 +1,282 @@
...
+  );
+};
+
+const configureLogging = (childProcess: ChildProcess, logPrefix: string) => {
+  const normalizedPrefix = (logPrefix + " ".repeat(25)).substring(0, 25);
+  const normalizedErrorPrefix = (
+    logPrefix +
+    " (stderr)" +
+    " ".repeat(25)
+  ).substring(0, 25);
```
hatskier, line 107: Can we move 25 to a constant?

packages/integration/integration-test-classic.ts
```diff
@@ -0,0 +1,106 @@
...
+  process.env.ADAPTER_CONTRACT_ADDRESS = adapterContractAddress;
+
+  onChainRelayerProcess = common.runWithLogPrefixInBackground(
+    "yarn",
+    ["start"],
+    "on-chain-relayer"
+  );
+
+  // Verify prices on-chain
+  while (
```
hatskier, line 83: Won't it freeze instead of failing tests sometimes?

hatskier (changes requested):
Great job! Clean, nice, and easily extendable code 💪

packages/integration/integration-test-classic.ts
rpiszczatowski: I'll change it to 5 loop revolutions and then fail.

hatskier: Why do we need to have a loop here at all?

hatskier (approved):

PR #856, opened 2023-07-14
chore: simplify fallback deviation/time update conditions

Lukasz2891:


packages/on-chain-relayer/src/args/get-iteration-args.ts
```diff
@@ -22,49 +18,41 @@ export const getIterationArgs = async (
...
     valuesFromContract = await getValuesForDataFeeds(
       adapterContract,
       dataFeeds
     );
   }
+  const dataPackages = await fetchDataPackages(
+    relayerConfig,
+    valuesFromContract
+  );
+  const olderDataPackagesPromise =
```
hatskier, line 42: Why can't we wait just add await here instead of passing the promise?

packages/on-chain-relayer/src/core/fetch-data-packages.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+  }
+
+  throw (
+    `Historical packages fetcher for fallback deviation check is not properly configured: ` +
+    `offset=${fallbackOffsetInMinutes} min., gateway=${historicalPackagesGateway}`
+  );
+};
+
+const olderPackagesTimestamp = (deviationCheckOffsetInMinutes: number) => {
```
hatskier, line 50: Please rename to `calculateOlderPackagesTimestamp`

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
```diff
@@ -1,117 +1,43 @@
...
+    valuesFromContract,
+    config
+  );
+
+  const isFallback = (config.fallbackOffsetInMinutes ?? 0) > 0;
+  let olderShouldUpdatePrices = true;
+  let olderWarningMessage = "";
+
+  if (shouldUpdatePrices && isFallback) {
+    const olderDataPackages = await olderDataPackagesPromise;
```
hatskier, line 22: AFAIU, you try to fetch older data packages only if it's needed. But the following code doesn't work this way. IN the moment you've created the `olderDataPackagesPromise` it starts fetching the data. Here we only add waiting for it

hatskier (changes requested):

packages/on-chain-relayer/src/args/get-iteration-args.ts
Lukasz2891: The intention was to have lazy fetching, but it failed to do it :( 

I'll change it to a callback.

packages/on-chain-relayer/src/core/update-conditions/value-deviation-condition.ts
```diff
@@ -1,117 +1,67 @@
...
+import { fetchDataPackages } from "../fetch-data-packages";
 
-export const valueDeviationCondition = (
-  dataPackages: DataPackagesResponse,
+export const performValueDeviationConditionChecks = async (
+  latestDataPackages: DataPackagesResponse,
   valuesFromContract: ValuesForDataFeeds,
-  config: RelayerConfig
+  config: RelayerConfig,
+  olderDataPackagesFetchCallback: () => Promise<DataPackagesResponse>
```
hatskier, line 10: Can't we just use this function directly without passing it as an argument?
Also, I would rename it to `getOlderDataPackages` or `fetchOlderDataPackages`

hatskier (approved):

PR #858, opened 2023-07-14
feat: fetch uniqueSignersCount from contract in on-chain-relayer

Lukasz2891:


rpiszczatowski:
looks good.

rpiszczatowski (approved):

packages/on-chain-relayer/src/core/contract-interactions/get-unique-signers-threshold.ts
```diff
@@ -0,0 +1,7 @@
+import { RedstoneAdapterBase } from "../../../typechain-types";
+
+export const getUniqueSignersThresholdFromContract = async (
```
hatskier, line 3: What is the added value of this file?

hatskier (approved):

packages/on-chain-relayer/src/core/contract-interactions/get-unique-signers-threshold.ts
Lukasz2891: Consistency with all other similar files.

PR #864, opened 2023-07-18
feat: add balancer sources to GHO

cehali:


hatskier (dismissed):
LGTM with one minor comment

PR #870, opened 2023-07-19
feat: integration tests framework

rpiszczatowski:


packages/cache-service/test/e2e/data-packages.spec.ts
```diff
@@ -363,7 +363,7 @@ describe("Data packages (e2e)", () => {
     const parsedDataPoints = JSON.parse(allFeedsDataPackages[0]).dataPoints;
     expect(allFeedsDataPackages.length).toBe(4);
     expect(parsedDataPoints.length).toBe(3);
-    for (const [_, dataPackages] of Object.entries<any>(testResponse2.body)) {
+    for (const [, dataPackages] of Object.entries<any>(testResponse2.body)) {
```
cehali, line 366: Can we change `any` to data packages type?

packages/integration/.eslintrc.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
```
cehali, line 1: Have you checked if it is compatible with our prettier settings? We don't want to have conflicts between these frameworks.

packages/integration/integration-test-classic.ts
```diff
@@ -94,14 +58,39 @@ const main = async () => {
...
       ],
       "relayer-contract",
+      { ADAPTER_CONTRACT_ADDRESS: adapterContractAddress },
       false
     );
+
+const main = async () => {
+  cacheLayerInstance = await startAndWaitForCacheLayer("1");
+  oracleNodeInstance = await startAndWaitForOracleNode(
+    "1",
```
cehali, line 68: Please extract `1` to separate variable called `instanceId` or something similar

packages/integration/integration-test-classic.ts
```diff
@@ -94,14 +58,39 @@ const main = async () => {
...
+  );
 
   process.exit();
 };
 
-common.configureCleanup(stopAll);
+configureCleanup(stopAll);
 
-main();
+void main();
```
cehali, line 96: Do we need `void` here?

packages/integration/integration-test-core.ts
```diff
@@ -1,22 +1,46 @@
...
 
   process.exit();
-}
+};
 
-common.configureCleanup(stopAll);
+configureCleanup(stopAll);
 
-main();
+void main();
```
cehali, line 46: Do we need void here?

cehali (dismissed):
Good job, small things in comments

packages/integration/.eslintrc.js
rpiszczatowski: I checked that but I'll additionally add eslint-prettier-plugin.

packages/integration/integration-test-classic.ts
rpiszczatowski: This is a price we pay for having "@typescript-eslint/no-floating-promises" eslint rule enabled. The void cast disables this check and here we simply cannot use await keyword. I'll change this to explicitly disabling the rule for this line.

packages/cache-service/test/e2e/data-packages.spec.ts
rpiszczatowski: My understanding of the code below is that `dataPackages` variable type is `string[]` as we access it as an array and pass this array's elements to `JSON.parse`. I'll then leave it as is.

PR #877, opened 2023-07-24
chore: improve testing framework even more

rpiszczatowski:


packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -0,0 +1,226 @@
...
+  if (direct && instance.directCacheServicePort) {
+    return instance.directCacheServicePort;
+  }
+  if (!direct && instance.publicCacheServicePort) {
+    return instance.publicCacheServicePort;
+  }
+  return ++cacheServicePort;
+};
+
+async function startAndWaitForCacheService(
```
cehali, line 41: Please try to use one of the arrow function s(const <function name> = () => {}) or simple functions in one file

packages/oracle-node/mock-prices.json
```diff
@@ -0,0 +1 @@
+{"BTC":16000,"ETH":1500,"__DEFAULT__":42}
```
cehali, line 1: Please add new line

packages/integration/src/prices-updates-core.ts
```diff
@@ -0,0 +1,54 @@
...
+const oracleNodeInstance: OracleNodeInstance = { instanceId: "1" };
+
+const stopAll = () => {
+  console.log("stopAll called");
+  stopOracleNode(oracleNodeInstance);
+  stopCacheLayer(cacheLayerInstance1);
+  stopCacheLayer(cacheLayerInstance2);
+};
+
+const main = async () => {
```
cehali, line 26: I am wondering if we need it to be a separate test from `simple-price-propagation-core`. It is similar but changes price a few times and wait for it to propagate. Maybe we could merge it into one test?

packages/integration/src/simple-price-propagation-core.ts
```diff
@@ -0,0 +1,38 @@
...
+const oracleNodeInstance: OracleNodeInstance = { instanceId: "1" };
+
+const stopAll = () => {
+  console.log("stopAll called");
+  stopOracleNode(oracleNodeInstance);
+  stopCacheLayer(cacheLayerInstance);
+};
+
+const main = async () => {
+  setMockPrices({ __DEFAULT__: 42 });
```
cehali, line 25: Why did you decide to use only one token here? I think it could be helpful to work on multiple tokens because I can imagine a situation where for one token something could work (populating object or sth like that) but not work with multiple tokens.

packages/integration/src/framework/cache-layer-manager.ts
rpiszczatowski: Of course. This is a result of vscode's extract method refactoring.

packages/oracle-node/mock-prices.json
rpiszczatowski: This file was committed by accident. I'll delete it.

packages/integration/src/prices-updates-core.ts
rpiszczatowski: Creating a proper set of test suits is something I am still working on. I created this particular suit to make sure that the framework is usable. I started with simple cases and now I am trying to add more complex cases as separate suits. My approach here is to leave it and in the final PR where I'll implement most of the use cases of our interest, we'll decide on what to leave and what to get rid of.
For now, we want to make the framework usable.

packages/integration/src/simple-price-propagation-core.ts
rpiszczatowski: The way mock fetcher works is that it returns anything you ask it for but if the price is not explicitly set for the token you'll get __DEFAULT__ value. For this suit I was only interested in price propagation, so 42 was a price good enough for every token :)
Once setMockPrices is called with another PriceSet the mock fetcher immediately starts returning 'fresh' prices, so I can set new set of prices later on if needed.

cehali (approved):

PR #884, opened 2023-07-26
feat: add long dry runs for avalanche and primary

cehali:


.github/workflows/long-avalanche-dry-run-test.yml
```diff
@@ -0,0 +1,33 @@
+name: Long avalanche dry run test
+on:
+  push:
+    branches: "main"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
cehali, line 7: This will cancel previous jobs in the new one is coming

koonopek: What do you think about running all these tests in a single workflow file (sequentially)?
I think that it is the worse solution, however, it has one advantage of minimising API usage (cause of rate limits problem)

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -0,0 +1,95 @@
+import { MathUtils } from "redstone-utils";
+
+const omittedDataFeeds = ["___ALL_FEEDS___", "crvUSDBTCETH"];
```
koonopek, line 3: Why do we omit `crvUSDBTCETH`

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -0,0 +1,95 @@
...
+      dataFeedId: string;
+      value: number;
+    }>;
+    dataServiceId: string;
+    dataFeedId: string;
+    signerAddress: string;
+  }>;
+}
+
+export const compareDataPackagesFromLocalAndProd = (
```
koonopek, line 20: Are we reusing the script written by you or Alex, for comparing responses from DB?
If not, maybe we could do this WDYT?

koonopek: If so, it would be best to separate it to some common function used both here and in scripts

packages/integration/src/framework/run-long-price-propagation-core-test.ts
```diff
@@ -0,0 +1,124 @@
...
+    );
+
+    const iterationsCount =
+      nodeWorkingTimeInMilliseconds / nodeIntervalInMilliseconds;
+    const timestampsRange = [...Array(iterationsCount).keys()];
+    const timestampsCountToAnalyze = timestampsRange.slice(
+      0,
+      iterationsCount - coldStartIterationsCount
+    );
+    for (const timestampDiffNumber of timestampsCountToAnalyze) {
```
koonopek, line 66: What do you think about optimising it with Promise.all (maybe no need, if it is pretty fast already)

koonopek (commented):
GJ! Left some questions / ideas

packages/integration/src/framework/compare-data-packages.ts
cehali: It is a little bit different, in db we aggregate them by the signer and also check if there is the same number of data packages. Here is the simplified version and I think it is so simple that adjusting code from db would be more complex than this code

packages/integration/src/framework/compare-data-packages.ts
cehali: We changed calculation method and currently there is big difference between prod and dev nodes, it could be removed after updating pro nodes

packages/integration/src/framework/run-long-price-propagation-core-test.ts
cehali: Agree, I will change it

.github/workflows/long-avalanche-dry-run-test.yml
cehali: Good idea

packages/integration/src/framework/run-long-price-propagation-core-test.ts
cehali: Done

.github/workflows/long-avalanche-dry-run-test.yml
cehali: As discussed in person, we should leave it as it is. If we hit rate limits with one additional node we should probably consider increasing APIs plans

packages/integration/.eslintrc.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
   ],
   parser: "@typescript-eslint/parser",
   parserOptions: {
-    project: "tsconfig.json",
+    tsconfigRootDir: __dirname,
```
rpiszczatowski, line 9: instead of this change you should add the following to your vscode workspace configuration file:
```
    "eslint.workingDirectories": [
      "./packages/gelato-relayer",
      "./packages/integration",
      "./packages/cache-service",
    ]
```

rpiszczatowski (approved):

packages/integration/src/framework/compare-data-packages.ts
hatskier, line 3: I would see a big value in comparing data from ____ALL_FEEDS___ along with all other feeds

hatskier, line 3: I believe we should not ignore any feeds and we can use local manifest for checking if all required feeds are available in the local node, we can also use the manifest from the URL specified in the remote config for monitoring, as we specify it there. OFC, we should not hardocde the url to remote-config, we should pass it as a secret

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -0,0 +1,95 @@
...
+    const dataPointsFromSecond = secondDataPackage[dataFeedId];
+    const deviations = dataPointsFromSecond.map(({ dataPoints }) =>
+      MathUtils.calculateDeviationPercent({
+        newValue: valueFromFirstDataPackages,
+        prevValue: dataPoints[0].value,
+      })
+    );
+    const maxDeviation = Math.max(...deviations);
+    if (maxDeviation >= maxPercentageValueDifference) {
+      throw new Error(
```
hatskier, line 90: I think we should not throw the error right away. We should collect the final report, print it and if it has any errors then throw an error (in the end of the github action)

packages/integration/src/framework/fetch-data-packages-from-local-and-prod-cache.ts
```diff
@@ -0,0 +1,21 @@
...
+  manifestFileName: string
+) => {
+  const responseFromLocalCache = (
+    await axios.get(
+      `http://localhost:${cacheLayerInstance.directCacheServicePort}/data-packages/historical/mock-data-service/${timestamp}`
+    )
+  ).data;
+  const responseFromProdCache = (
+    await axios.get(
+      `https://oracle-gateway-2.a.redstone.finance/data-packages/historical/redstone-${manifestFileName}-prod/${timestamp}`
```
hatskier, line 16: Can we move the URL prefix to a constant?

packages/integration/src/framework/fetch-latest-timestamp-from-local-cache.ts
```diff
@@ -0,0 +1,15 @@
+import axios from "axios";
```
hatskier, line 1: I would rename the folder integration to e2e (end to end tests), maybe in a separate PR

packages/integration/src/framework/compare-data-packages.ts
hatskier: But we can do it in a separate PR

hatskier: We can do it in a separate PR

packages/integration/src/framework/run-long-price-propagation-core-test.ts
```diff
@@ -0,0 +1,108 @@
...
+const overwriteNodeInterval = (
+  nodeInterval: number,
+  manifestFileName: string
+) => {
+  const newManifestPath = `../oracle-node/manifests/data-services/${manifestFileName}-${oracleNodeInstance.instanceId}.json`;
+  fs.copyFileSync(
+    `../oracle-node/manifests/data-services/${manifestFileName}.json`,
+    newManifestPath
+  );
+  updateManifestFile("interval", String(nodeInterval), newManifestPath);
```
hatskier, line 107: Can't we just call
```js
updateManifestFile("interval", String(nodeInterval), `../oracle-node/manifests/data-services/${manifestFileName}.json`);
```
?

packages/integration/src/framework/run-long-price-propagation-core-test.ts
```diff
@@ -0,0 +1,108 @@
...
+    await sleep(nodeWorkingTimeInMilliseconds);
+    stopOracleNode(oracleNodeInstance);
+
+    const latestTimestamp = await fetchLatestTimestampFromLocal(
+      cacheLayerInstance
+    );
+
+    const iterationsCount =
+      nodeWorkingTimeInMilliseconds / nodeIntervalInMilliseconds;
+    const timestampsRange = [...Array(iterationsCount).keys()];
```
hatskier, line 59: Let's add a simple check if `timestampsRange.length` is not too small (e.g. at least 12)

packages/integration/src/framework/run-long-price-propagation-core-test.ts
```diff
@@ -0,0 +1,108 @@
...
+    const timestampsCountToAnalyze = timestampsRange.slice(
+      0,
+      iterationsCount - coldStartIterationsCount
```
hatskier, line 62: Since we need to do a cold start to calculate dependent values, our change of the interval (from 10s to 60s) can influence the result values. Maybe we can run first `coldStartIterationsCount` iterations with 10s and then switch to `60s`. Let's discuss it and maybe implement in a separate PR

packages/integration/src/framework/run-long-price-propagation-core-test.ts
```diff
@@ -0,0 +1,108 @@
...
+    const nodeWorkingTimeInMilliseconds =
+      MINUTE_IN_MILLISECONDS * nodeWorkingTimeInMinutes;
+    await sleep(nodeWorkingTimeInMilliseconds);
+    stopOracleNode(oracleNodeInstance);
+
+    const latestTimestamp = await fetchLatestTimestampFromLocal(
+      cacheLayerInstance
+    );
+
+    const iterationsCount =
```
hatskier, line 57: WDYT about requesting the latest timestamp and then just deducting `nodeIntervalInMilliseconds`?

hatskier (changes requested):

packages/integration/src/framework/compare-data-packages.ts
cehali: After discussion, print all deviations also

packages/integration/src/framework/run-long-price-propagation-core-test.ts
cehali: Change to default 10 sec, delete overwrite manifest function

PR #885, opened 2023-07-26
feat: fix stats endpoint in gateway to effective one

koonopek:
The current implementation was very very slow, it never finished on example queries from single point in time (not range).

New query is more effective. It works


packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -304,48 +304,37 @@ export class DataPackagesService {
...
+        "Too big search period. to-timestamp - from-timestamp has to be bigger than 3600_000"
+      );
+    }
+
+    const oraclesState = await getOracleState();
+    const nodes = Object.values(oraclesState.nodes);
+
+    const countsPerNode = await Promise.all(
+      nodes.map(async (node) => {
+        const count = await DataPackage.countDocuments({
```
koonopek, line 318: Court documents is much more effective then aggregate - https://stackoverflow.com/questions/33181878/mongodb-count-vs-aggregation

packages/cache-service/src/data-packages/data-packages.interface.ts
```diff
@@ -86,9 +86,7 @@ export interface DataPackagesResponse {
 
 export interface DataPackagesStatsResponse {
   [signerAddress: string]: {
-    dataPackagesCount: number;
     verifiedDataPackagesCount: number;
-    verifiedDataPackagesPercentage: number;
```
koonopek, line 91: It would require more queries, no value added I think

packages/cache-service/test/e2e/data-packages.spec.ts
```diff
@@ -541,18 +541,9 @@ describe("Data packages (e2e)", () => {
...
         "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266": {
-          dataPackagesCount: 32,
-          verifiedDataPackagesCount: 32,
-          verifiedDataPackagesPercentage: 100,
+          dataServiceId: "mock-data-service-1",
+          verifiedDataPackagesCount: 8,
           nodeName: "Mock node 1",
-          dataServiceId: MOCK_DATA_SERVICE_ID,
-        },
-        "0x2": {
```
koonopek, line 550: We now return only packages signed by nodes from oracleState

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -304,48 +304,37 @@ export class DataPackagesService {
   ): Promise<DataPackagesStatsResponse> {
     const { fromTimestamp, toTimestamp } = statsRequestParams;
 
-    // Fetching stats form DB
-    const signersStats = await DataPackage.aggregate([
-      {
-        $match: {
+    if (toTimestamp - fromTimestamp > 7200_000) {
```
koonopek, line 307: to prevent from accidental DDoS of our DB

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -304,48 +304,37 @@ export class DataPackagesService {
...
+      throw new BadRequestException(
+        "Too big search period. to-timestamp - from-timestamp has to be bigger than 7200_000"
```
hatskier, line 309: has to be bigger than 7200_000 -> can not be bigger than 7_200_000?

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -304,48 +304,37 @@ export class DataPackagesService {
...
+      );
+    }
+
+    const oraclesState = await getOracleState();
+    const nodes = Object.values(oraclesState.nodes);
+
+    const countsPerNode = await Promise.all(
```
hatskier, line 316: Have you verified if making 5 separate queries is better then one aggregation call? Have you checked if there is a way to send a batch with the count queries instead of using Promise.all and sending separate requests to the mongo

koonopek: Yes as I mentioned in the description and in the comment https://github.com/redstone-finance/redstone-oracles-monorepo/pull/885#discussion_r1274906967 aggregation is at least x1000+ slower (of course depending on the number of documents, but we have a lot of them).
There is something like https://stackoverflow.com/questions/52913008/mongodb-execute-multiple-queries-in-one-round-trip but it is no native and has some limitations, I would go with the most straightforward solution here.

hatskier (approved):

PR #891, opened 2023-07-27
fix: manifest path in long dry runs

cehali:


.github/workflows/long-avalanche-dry-run-test.yml
```diff
@@ -1,7 +1,7 @@
 name: Long avalanche dry run test
 on:
   push:
-    branches: "main"
+    branches: "*"
```
rpiszczatowski, line 4: Are you sure about this? All the code review branches?

packages/integration/src/framework/run-long-price-propagation-core-test.ts
```diff
@@ -89,6 +86,7 @@ export const runLongPricePropagationCoreTest = async (
       );
     }
     process.exit();
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
rpiszczatowski, line 89: The correct way is to use the `unknown` as a type for the `error` variable instead of disabling linter.

rpiszczatowski (changes requested):
.

packages/integration/src/framework/run-long-price-propagation-core-test.ts
cehali: image
I am not sure, I cannot use `error` then

rpiszczatowski: This is unfortunate but technically any type of object can be thrown from JS and this is why typescript does not allow arbitrary types for catch variables but only `any` and `unknown` are allowed. If you want to make use of the object as a particular type you have to add proper type assertion: (error as Error).stack will do.

To summarize IMO the 'proper' way would be to use `unknown` as a type and then explicitly cast in the catch statement body.

rpiszczatowski (approved):

packages/integration/src/framework/run-long-price-propagation-core-test.ts
rpiszczatowski: BTW if you looked at the `configureCleanup` method you would see that it adds handling of unhandled exceptions, so this whole try catch is probably not very valuable. If you think it is good to have some better handling maybe you could add it in `configureCleanup` so all the suits make use of it.

cehali: Ok, will remove it in next PR which will improve long time tests

PR #893, opened 2023-07-28
feat: add obfuscation to node code

cehali:


packages/oracle-node/Dockerfile
```diff
@@ -10,6 +10,14 @@ WORKDIR /app/packages/oracle-node
 
 RUN yarn
 RUN yarn build
+RUN apt-get update && apt-get install cpio
+RUN yarn copy-jsons
+RUN yarn obfuscate
+RUN cp -R oracle-node-obfuscated ..
+RUN cp Dockerfile ../oracle-node-obfuscated
```
koonopek, line 17: Why do we need Dockerfile in Docker already :)

packages/oracle-node/Dockerfile
```diff
@@ -10,6 +10,14 @@ WORKDIR /app/packages/oracle-node
 
 RUN yarn
 RUN yarn build
+RUN apt-get update && apt-get install cpio
```
koonopek, line 13: I think this should be run on line 10 for better caching

koonopek (dismissed):

packages/oracle-node/Dockerfile
```diff
@@ -8,8 +8,15 @@ FROM redstone-monorepo:${VERSION}
 
 WORKDIR /app/packages/oracle-node
 
+RUN apt-get update && apt-get install cpio
 RUN yarn
 RUN yarn build
+RUN yarn copy-jsons
+RUN yarn obfuscate
+RUN cp -R oracle-node-obfuscated ..
```
rpiszczatowski, line 16: Wouldn't it be cleaner to output the obfuscated code directly to `../oracle-node-obfuscated`? You wouldn't need to copy around the result.

cehali: It is obfuscated to `../oracle-node-obfuscated` but we want to have it in the main folder so it could replace the original code

rpiszczatowski (changes requested):
Assuming the goal of this PR is to hide oracle-node code from the public, it doesn't do it as oracle-node code can be still seen:
- in other images: relayer, cache service
- in the intermediate layers

I recommend switching to multistage builds and being explicit about what we have in our images.

packages/oracle-node/Dockerfile
rpiszczatowski: It is obfuscated to `./oracle-node-obfuscated` from what I see. I think the better way would be to do as you wrote to obfuscate it to `../oracle-node-obfuscated` and then just rename the directory, exactly as you are doing now. This would allow for getting rid of `cp -R ...`

cehali: It won't be as then it will rename each obfuscated file to e.g. `index-obfuscated.js` and all things depending on filenames would stop working

cehali: It will change in new version and I will remove `cp -R ...`

Dockerfile
```diff
@@ -1,10 +1,94 @@
...
+# ---- Base node ----
+FROM node:16 AS base
 WORKDIR /app
 
-# Bundle app source
-COPY . .
+# ---- Build protocol code ----
+FROM base AS protocol
+WORKDIR /app/protocol
+COPY packages/protocol/package.json ./
```
rpiszczatowski, line 8: What about top-level package.json and what about yarn.lock? Shouldn't those be present to use the same packages versions as those used for testing?

packages/cache-service/test/common/mock-values.ts
```diff
@@ -1,7 +1,7 @@
 import { ethers } from "ethers";
 import { RedstoneOraclesState } from "redstone-oracles-smartweave-contracts/src/contracts/redstone-oracle-registry/types";
 import { DataPackage, DataPoint, NumericDataPoint } from "redstone-protocol";
-import { convertNumberToBytes } from "redstone-protocol/src/common/utils";
+import { utils } from "redstone-protocol";
```
rpiszczatowski, line 4: Why those changes are required? Test code shouldn't be even touched during docker image build so it should be irrelevant to changes done in Dockerfiles.

cehali: It is much relevant for build stage and importing from local dependencies

rpiszczatowski (commented):
What about cache-service and relayer Dockerfiles. The are not changed by this PR (yet)

packages/rpc-providers/package.json
```diff
@@ -17,7 +17,7 @@
   "devDependencies": {
     "@nomicfoundation/hardhat-chai-matchers": "^1.0.0",
     "@nomicfoundation/hardhat-network-helpers": "^1.0.8",
-    "@nomicfoundation/hardhat-toolbox": "^2.0.2",
+    "@nomicfoundation/hardhat-toolbox": "2.0.2",
```
rpiszczatowski, line 20: Why this change?

packages/cache-service/Dockerfile
```diff
@@ -1,25 +1,34 @@
-# Docker container built using this Dockerfile
-# can be published to the public docker registry
-# But before running the container it must be configured to provide
-# the node config through environment variables
+# ---- Build root ----
+FROM node:16 AS build
+WORKDIR /app
+COPY . .
+RUN yarn install --frozen-lockfile --non-interactive --production=false
```
rpiszczatowski, line 5: What is the idea here? Why do we first run `yarn install --production=false` here and then we run the same command with `--production=true`?

packages/cache-service/Dockerfile
```diff
@@ -1,25 +1,34 @@
...
+RUN yarn workspaces run build
```
rpiszczatowski, line 6: Wouldn't it be enough to simply run `yarn build` in the cache-service directory. This should build all the dependencies automatically.

packages/oracle-node/Dockerfile
```diff
@@ -1,27 +1,45 @@
...
+FROM node:16 AS build
+WORKDIR /app
+COPY . .
+RUN yarn install --frozen-lockfile --non-interactive --production=false
+RUN yarn workspaces run build
 
-ARG VERSION
-FROM redstone-monorepo:${VERSION}
+# ---- Obfuscate node code ----
+FROM build AS oracle-node
```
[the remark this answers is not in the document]
cehali: It is better to have more stages as they can be calculated parallelly

packages/cache-service/package.json
```diff
@@ -41,6 +41,10 @@
     "mongoose": "^6.5.2",
     "nestjs-pino": "^3.1.1",
     "pino-http": "^8.2.1",
+    "redstone-protocol": "*",
```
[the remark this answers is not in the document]
cehali: I don't think I understand the question, you want to get rid of the version or get rid of the whole `redstone-protocol` etc? The empty string is equivalent to `"*"`. After reading about how dependencies work I think we should have all dependencies defined in the `package.json`, also local ones.

packages/oracle-node/Dockerfile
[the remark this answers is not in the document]
rpiszczatowski: But they aren't if one stage uses another as a base.

packages/cache-service/package.json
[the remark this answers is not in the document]
rpiszczatowski: The fact that we use workspaces does the job of making each workspace available to all the others (via symlinks in top-level node_modules directory created during `yarn install` command run).

packages/cache-service/Dockerfile
cehali: Discussed during call

packages/cache-service/package.json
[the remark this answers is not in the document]
cehali: Discussed during call

packages/cache-service/Dockerfile
cehali: Agree, implemented

packages/cache-service/Dockerfile
rpiszczatowski: Basically there is no way to compile the code without dev dependencies.

packages/on-chain-relayer/tsconfig.json
```diff
@@ -14,7 +14,8 @@
     "declaration": true,
     "declarationMap": true,
     "incremental": true,
-    "composite": true
+    "composite": true,
+    "baseUrl": "."
```
rpiszczatowski, line 18: Do we need `baseUrl`. It looks like a leftover of something. Similarly in other places.

packages/cache-service/Dockerfile
```diff
@@ -1,25 +1,61 @@
...
+# ---- Install and copy necessary files ----
+FROM node:16 AS install
+WORKDIR /app
+COPY package.json .
+COPY yarn.lock .
+COPY .yarnrc .
 
-ARG VERSION
-FROM redstone-monorepo:${VERSION}
+COPY packages/cache-service/package.json ./packages/cache-service/
```
rpiszczatowski, line 8: Now, with explicitly defined workspaces, we end up having workspaces for which we don't copy package.json before performing the build. I am not sure what implications this have.

rpiszczatowski (approved):

packages/cache-service/Dockerfile
cehali: We don't use the command `yarn workspaces` which will iterate through all defined workspaces, I tested it locally and we are able to build image and then run container from it

packages/on-chain-relayer/tsconfig.json
cehali: yes, I will remove it

rpiszczatowski (approved):

packages/cache-service/Dockerfile
```diff
@@ -1,25 +1,61 @@
...
+
+COPY --from=build /app/packages/utils/dist ./packages/utils/dist
+COPY --from=build /app/packages/utils/package.json ./packages/utils
+
+COPY --from=build /app/packages/oracles-smartweave-contracts/dist ./packages/oracles-smartweave-contracts/dist
+COPY --from=build /app/packages/oracles-smartweave-contracts/package.json ./packages/oracles-smartweave-contracts
+
+RUN yarn install --frozen-lockfile --non-interactive --production=true
+
 ENV ENABLE_JSON_LOGS=true
```
koonopek, line 55: We don't create own user as on OWASP suggetion

koonopek: Our recent report highlighted that 58% of images are running the container entrypoint as root (UID 0). However, it is a Dockerfile best practice to avoid doing that. There are very few use cases where the container needs to execute as root, so don’t forget to include the USER instruction to change the default effective UID to a non-root user.

packages/on-chain-relayer/Dockerfile
```diff
@@ -1,13 +1,63 @@
...
-FROM redstone-monorepo:${VERSION}
+# ---- Install and copy necessary files ----
+FROM node:16 AS install
+WORKDIR /app
+COPY package.json .
+COPY yarn.lock .
+COPY .yarnrc .
 
-WORKDIR /app/packages/on-chain-relayer
+COPY packages/on-chain-relayer/package.json ./packages/on-chain-relayer/
```
koonopek, line 8: Same suggestion as with cache-service DockerFile. Also as I see we could extract common image from all docker files because first stage of multibuild is the same

packages/oracle-node/Dockerfile
```diff
@@ -1,27 +1,72 @@
-# Docker container built using this Dockerfile
-# can be published to the public docker registry
-# But before running the container it must be configured to provide
-# the node config through environment variables
+# ---- Install and copy necessary files ----
+FROM node:16 AS install
+WORKDIR /app
+COPY package.json .
+COPY yarn.lock .
+COPY .yarnrc .
```
koonopek, line 7: Same as with other Dockerfiles

packages/cache-service/Dockerfile
```diff
@@ -1,25 +1,61 @@
...
 RUN yarn build
 
-# Setting production env variables
+# ---- Release ----
+FROM node:16 AS release
+WORKDIR /app
+COPY package.json .
+COPY yarn.lock .
+
+COPY --from=build /app/packages/cache-service/dist ./packages/cache-service
```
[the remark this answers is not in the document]
cehali: Because we don't want to copy all `package.json` files, we just want part of it. Yes, and this is correct that you add new packages and you consciously add it docker. It increase security as only things you know should be in container are here and not whole monorepo

package.json
```diff
@@ -18,7 +18,17 @@
   "private": true,
   "workspaces": {
     "packages": [
-      "packages/*"
```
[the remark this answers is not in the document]
cehali: Because not all things should be in workspaces, especially packages which are completely separate and not connected with others. It simplifies dependencies and add well defined structure

packages/cache-service/Dockerfile
```diff
@@ -1,25 +1,61 @@
...
+COPY --from=build /app/packages/streamr-proxy/dist ./packages/streamr-proxy/dist
+COPY --from=build /app/packages/streamr-proxy/package.json ./packages/streamr-proxy
+
+COPY --from=build /app/packages/utils/dist ./packages/utils/dist
+COPY --from=build /app/packages/utils/package.json ./packages/utils
+
+COPY --from=build /app/packages/oracles-smartweave-contracts/dist ./packages/oracles-smartweave-contracts/dist
+COPY --from=build /app/packages/oracles-smartweave-contracts/package.json ./packages/oracles-smartweave-contracts
+
+RUN yarn install --frozen-lockfile --non-interactive --production=true
```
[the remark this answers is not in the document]
cehali: Because previous `node_modules` won't be the same as this one as we use production mode. Also, I don't think copying `node_modules` in good practice.

packages/cache-service/Dockerfile
```diff
@@ -1,25 +1,61 @@
...
+FROM node:16 AS install
+WORKDIR /app
+COPY package.json .
+COPY yarn.lock .
+COPY .yarnrc .
 
-ARG VERSION
-FROM redstone-monorepo:${VERSION}
+COPY packages/cache-service/package.json ./packages/cache-service/
+COPY packages/protocol/package.json ./packages/protocol/
```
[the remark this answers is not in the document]
cehali: Do you know the command which copies well-defined part of the monorepo to the docker? Also, it is better to define what packages and not copy the whole monorepo as it was before. It reduces docker image size, increases security and makes it really straightforward what it is used. If you want to add a new package used by another it would require additional 4 lines of change.

packages/on-chain-relayer/Dockerfile
cehali: It is not the same as we don't want to copy the whole monorepo. On-chain relayer or cache service don't need oracle-node or evm-connector. This could provide additional problems and additional work like obfuscating node code. In this way, we have well-defined structure what is needed to make the package work.

packages/oracle-node/Dockerfile
cehali: Same anserw

cehali: I think the best is to explain it on Monday during the call

packages/cache-service/Dockerfile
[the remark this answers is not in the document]
koonopek: Can't we use docker-ignore for that?

packages/cache-service/Dockerfile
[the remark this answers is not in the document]
koonopek: Why? It is a common pattern I think in docker files. You don't have yarn installed on the final image which is an advantage.

packages/cache-service/Dockerfile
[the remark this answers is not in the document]
koonopek: Per Dockerfile, in my opinion it matters as long as it is more then 1 line.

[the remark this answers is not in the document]
koonopek: `.dockerignore` + `COPY src_dir dst_dir`

packages/on-chain-relayer/Dockerfile
koonopek: Okey, I got your point now, I think we can discuss it after daily WDYT.
I see the advantages and disadvantages of both solutions now, thanks.

packages/cache-service/Dockerfile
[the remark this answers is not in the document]
rpiszczatowski: The main point is that we don't want devDependencies in the final image. The approach with calling yarn install twice is described e.g. here: https://sachithsiriwardana.medium.com/dockerizing-nodejs-application-with-multi-stage-build-e30477ca572

packages/cache-service/Dockerfile
rpiszczatowski: Let's create a ticket for this.

packages/cache-service/Dockerfile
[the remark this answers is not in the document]
rpiszczatowski: We build multiple docker images from the top directory of our monorepo. Having common .dockerignore maybe problematic. We may only put there things that are not required by any of images.

.github/workflows/oracle-node-deploy-to-ecr-public.yml
```diff
@@ -4,7 +4,7 @@
```
rpiszczatowski, line 4: Why changing the file name. Now it is not adequate as it builds also cache-service.

BTW do we have any particular reason to have two workflows for three images? Maybe we could have a single workflow?

cehali: You are right, this name shouldn't be changed. Two workflows for three images is pointless, either three workflows or one 😄 I would opt for three

packages/cache-service/Dockerfile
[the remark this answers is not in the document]
cehali: I thought about docker-ignore also but I belive it wouldn't be simpler. The biggest issue I think would be with readability, you had to remember that most of configuration how would docker container would look like is in docker-ignore

.github/workflows/oracle-node-deploy-to-ecr-public.yml
cehali: Updated

rpiszczatowski (approved):

PR #896, opened 2023-07-29
feat: make /stats endpoint cache friendly

koonopek:


packages/cache-service/src/data-packages/data-packages.controller.ts
```diff
@@ -137,9 +137,10 @@ export class DataPackagesController {
     this.sendSerializableResponse(res, payload, query.format);
   }
 
-  @Get("stats")
+  @Get("stats/:PERIOD")
```
koonopek, line 140: I don't see any advantage of providing here hours instead of MS.
MS has an advantage that can be more granular.

PR #897, opened 2023-07-29
integration tests improvements and fixes

rpiszczatowski:
- unify logging
- allow to configure the time period from which data returned from cache service is taken. This allows to shrink the time after a single node failure after which customers of cache service API starts getting data from remaining nodes
- fix stopping applications on ubuntu (use node directly instead of yarn). When yarn was used intermediate processes were created and it was not possible to kill actual node or cache service.


packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -38,15 +41,25 @@ const portNumberForInstance = (
   return ++cacheServicePort;
 };
 
+const cacheServiceDir = "../cache-service";
+const evmConnectorDir = "../evm-connector";
+
+export const buildCacheLayer = async () =>
```
hatskier, line 47: IMO we can remove async and await here, and it will work in exactly the same way

packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -38,15 +41,25 @@ const portNumberForInstance = (
...
+  await installAndBuild(cacheServiceDir, false);
+
+export const buildEvmConnector = async () =>
```
hatskier, line 50: And here :)

packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -98,41 +111,36 @@ export const startAndWaitForCacheLayer = async (
...
+  instance.mongoDbProcess = undefined;
 };
 
 export const stopDirectCacheService = (instance: CacheLayerInstance) => {
+  debug(`stopping direct-${logPrefix(instance)}`);
   stopChild(
-    instance?.directCacheServiceProcess,
-    `direct cache service-${instance?.instanceId}`
+    instance.directCacheServiceProcess,
+    `direct-${logPrefix(instance)}`
```
hatskier, line 134: logPrexif -> getLogPrefix?

packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -230,3 +240,20 @@ export const verifyPricesInCacheService = async (
     "prices are not what expected"
   );
 };
+
+export const verifyPricesNotInCacheService = async (
+  cacheLayerInstances: CacheLayerInstance[],
+  expectedPrices: PriceSet
+) => {
+  let exceptionOccurred = false;
```
hatskier, line 248: What do you think about rewriting it in the following way?:
```js
try {
  await verifyPricesInCacheService(cacheLayerInstances, expectedPrices);
  throw new Error("ShouldNotThrow");
} catch (e: any) {
  if (e.message === "ShouldNotThrow") {
    throw new Error(
      "IMPOSSIBLE: prices were updated even though there was no active node. Most probably there is a bug in testing code."
    );
  }
}
```

packages/integration/src/framework/integration-test-utils.ts
```diff
@@ -31,21 +29,29 @@ export const printDotenv = (label: string, filePath: string) => {
...
     await sleep(sleepTimeMilliseconds);
   }
+  debug(`${filePath} bacame available`);
 };
 
-export const waitForUrl = async (url: string) => {
-  // eslint-disable-next-line no-constant-condition
+export const waitForUrl = async (url: string, tries = 5) => {
+  // eslint-disable-next-line no-constant-condition,@typescript-eslint/no-unnecessary-condition
   while (true) {
```
hatskier, line 45: Can't we use for loop with `break` instead of `while true` loop?

packages/integration/src/framework/oracle-node-manager.ts
```diff
@@ -14,38 +15,52 @@ export const hardhatMockPrivateKey =
...
 export type OracleNodeInstance = {
-  oracleNodeProcess?: ChildProcess;
   instanceId: string;
+  oracleNodeProcess?: ChildProcess;
 };
 
-const mockPricesPath = "./mock-prices.json";
+const oracleNodeDir = "../oracle-node";
+
+export const buildOracleNode = async () =>
```
hatskier, line 24: I would propose to remove async and await here

packages/integration/src/framework/relayer-manager.ts
```diff
@@ -25,18 +21,20 @@ export type RelayerInstance = {
   instanceId: string;
 };
 
+const relayerDir = "../on-chain-relayer";
+
+const logPrefix = (instance: RelayerInstance) =>
+  `relayer-${instance.instanceId}`;
+
+export const buildRelayer = async () => installAndBuild(relayerDir, true);
```
hatskier, line 29: I would remove async here

packages/integration/src/framework/relayer-manager.ts
```diff
@@ -62,15 +60,16 @@ export const startRelayer = async (
...
+    ["dist/src/run-relayer"],
+    logPrefix(instance),
+    relayerDir,
     { DOTENV_CONFIG_PATH: dotenvPath }
   );
 };
 
-export const stopRelayer = (relayer: RelayerInstance | undefined) => {
-  stopChild(relayer?.relayerProcess, `on chain relayer-${relayer?.instanceId}`);
+export const stopRelayer = (instance: RelayerInstance) => {
```
hatskier, line 71: Do we need curly braces here?

packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -23,8 +24,10 @@ export type CacheLayerInstance = {
   dotenvPath?: string;
 };
 
-let cacheServicePort = 3000;
+const logPrefix = (instance: CacheLayerInstance) =>
```
cehali, line 27: get or build `logPrefix` ?

packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -98,41 +111,36 @@ export const startAndWaitForCacheLayer = async (
...
-    cacheLayer?.publicCacheServiceProcess,
-    `public cache service-${cacheLayer.instanceId}`
-  );
-  cacheLayer.publicCacheServiceProcess = undefined;
-  stopChild(cacheLayer.mongoDbProcess, `mongo-${cacheLayer?.instanceId}`);
-  cacheLayer.mongoDbProcess = undefined;
+export const stopCacheLayer = (instance: CacheLayerInstance) => {
+  debug(`stopping ${logPrefix(instance)}`);
+  stopDirectAndPublicCacheServices(instance);
+  stopChild(instance.mongoDbProcess, `mongo-${instance.instanceId}`);
```
cehali, line 126: If we have a separate function for stopping cache services maybe we want to have a separate function for stopping mongo?

packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -98,41 +111,36 @@ export const startAndWaitForCacheLayer = async (
...
-    instance?.directCacheServiceProcess,
-    `direct cache service-${instance?.instanceId}`
+    instance.directCacheServiceProcess,
+    `direct-${logPrefix(instance)}`
   );
   instance.directCacheServiceProcess = undefined;
 };
 
 export const stopPublicCacheService = (instance: CacheLayerInstance) => {
+  debug(`stopping public-${logPrefix(instance)}`);
```
cehali, line 140: Why don't we use `stopPublicCacheService`?

packages/integration/src/framework/cache-layer-manager.ts
rpiszczatowski: The difference will be that in the case of an exception thrown from the `installAndBuild` method, it won't appear in the stack trace and one more tick will be consumed to calculate the result. I think it is a good practice to write `return await f()` instead of `return f()` if `f` returns promise even if the difference in behavior is as slight as I mentioned.

packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -98,41 +111,36 @@ export const startAndWaitForCacheLayer = async (
...
+  stopChild(instance.mongoDbProcess, `mongo-${instance.instanceId}`);
+  instance.mongoDbProcess = undefined;
 };
 
 export const stopDirectCacheService = (instance: CacheLayerInstance) => {
+  debug(`stopping direct-${logPrefix(instance)}`);
   stopChild(
-    instance?.directCacheServiceProcess,
-    `direct cache service-${instance?.instanceId}`
+    instance.directCacheServiceProcess,
```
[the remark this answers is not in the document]
rpiszczatowski: Because we are inside this function?

cehali (approved):
LGTM

packages/integration/src/framework/relayer-manager.ts
rpiszczatowski: I'll setup proper eslint rules that will guide us in the future. I promise. For now closing.

PR #901, opened 2023-07-31
feat: add balancer pools for GHO

cehali:


packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
```diff
@@ -25,9 +25,8 @@ type BalancerPoolsConfigs = Record<
     swapAmountForSwaps: BigNumber;
     swaps: SwapConfig[];
     tokenAddresses: string[];
-    tokenInForSwaps: string;
-    tokenOutFromSwaps: string;
     tokenToFetch?: string;
+    priceSerializer?: number;
```
rpiszczatowski, line 29: Why this is named `priceSerializer`?

cehali: Because I know that in some cases we will need to multiple and in other divide by this number

rpiszczatowski (approved):

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
cehali: Ok, renamed

rpiszczatowski (approved):

PR #905, opened 2023-07-31
chore: remove curve as source

cehali:


rpiszczatowski (approved):

packages/oracle-node/manifests/data-services/primary.json
```diff
@@ -139,24 +139,14 @@
         "uniswap-v3-on-chain-weth-500"
       ]
     },
-    "FRXETH": {
-      "source": ["curve-eth", "uniswap-v3-on-chain-weth-500"]
```
rpiszczatowski, line 143: Why remove also the uni source?

packages/oracle-node/manifests/data-services/primary.json
```diff
@@ -44,6 +44,8 @@
         "coinbaseprime",
         "coinbasepro",
         "coingecko",
+        "curve-usdc",
```
rpiszczatowski, line 47: Why not remove this?

rpiszczatowski (approved):

packages/oracle-node/manifests/data-services/primary.json
cehali: This token was used only in `curve-frxeth` for ALETH which is not used anymore so `FRXETH` is not necessary

packages/oracle-node/manifests/data-services/primary.json
cehali: Those pools still have a lot of liquidity, we will also monitor them. This is why I decided to leave them

PR #912, opened 2023-08-03
feat: curve fetcher liqudity and slippage

koonopek:
Sllipage calcuation:
```
low_buy = get_dy(0,1, 1 token)
low_sell = 1/low_buy

// assume 5 tokens = 10k USD
high_buy = get_dy(0,1, 5 tokens)
high_sell = get_dy(1,0,5tokens)

slippage_buy = abs(high_buy-low_buy) / low_buy
slippage_sel = abs(high_sell-low_sell) / low_sell
```
Values are sensible however there are not the same as we can see in https://curve.fi


packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,35 +38,150 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+      ratioBigNumber,
+      baseTokenSupply,
+      pairedTokenSupply,
+      bigBuy,
+      bigSell,
     };
   }
 
+  private static convertUsdToTokenAmount(assetId: string): string {
+    return new Decimal(CURVE_SLIPPAGE_PARAMS.amountInUsd)
```
koonopek, line 136: will extract to same shared code in next PR

packages/oracle-node/src/fetchers/curve/curve-fetchers-config.ts
```diff
@@ -18,6 +18,10 @@ export interface PoolsConfig {
   };
 }
 
+export const CURVE_SLIPPAGE_PARAMS = {
```
koonopek, line 21: I assumed that we will be using only one value for slippage

packages/oracle-node/test/fetchers/curve-multi-block.spec.ts
```diff
@@ -149,7 +256,11 @@ describe("Curve Multi Block", () => {
 });
 
 const mockGetDyCalls = async (contract: MockContract, returns: any[]) => {
+  // this is the mock for slippage requests
+  await contract.mock.get_dy.returns(BigNumber.from("1000130962107597656"));
```
koonopek, line 260: ethereum-waffle sucks so much, it fails to queue calls properly and is ultra slow.

packages/utils/src/common/multicall.ts
```diff
@@ -0,0 +1,62 @@
+import { ContractCallContext, Multicall } from "ethereum-multicall";
+import { Contract } from "ethers";
+import { FormatTypes } from "ethers/lib/utils";
+
+let multicallAddress: undefined | string = undefined;
+export function overrideMulticallAddress(address: string) {
+  multicallAddress = address;
+}
+
+export async function multiCallOneContract(
```
koonopek, line 10: Simpler abstraction over etherum-mutlicall.
etherum-mutlicall is better then multicall-service, because from out perspective it is multichain agnostic

packages/oracle-node/index.ts
```diff
@@ -2,7 +2,9 @@ import { Consola } from "consola";
 import { config } from "./src/config";
 import NodeRunner from "./src/NodeRunner";
 import { closeLocalLevelDB, setupLocalDb } from "./src/db/local-db";
+import Decimal from "decimal.js";
 
+Decimal.set({ toExpPos: 9e15 });
```
rpiszczatowski, line 7: We want to be veeeeeeeery precise :)

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -1,9 +1,23 @@
 import { Decimal } from "decimal.js";
-import { BigNumber, Contract } from "ethers";
-import { getRawPrice } from "../../db/local-db";
+import { Contract } from "ethers";
+import { RedstoneCommon } from "redstone-utils";
+import { SlippageData } from "redstone-utils/src/types";
```
rpiszczatowski, line 4: Shouldn't we import `RedstoneTypes` instead and use `RedstoneTypes.SlippageData` instead of directly accessing the internals of Redstone utils? I know we do this all the time, but today we had a discussion with @cehali about it and we concluded that the suggested way is considered a cleaner one.

packages/oracle-node/index.ts
koonopek: :) This is not precision: https://mikemcl.github.io/decimal.js/#toExpPos

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: Yes of course, just auto-import will fix

packages/oracle-node/src/fetchers/dex-on-chain/DexOnChainFetcher.ts
```diff
@@ -82,6 +83,9 @@ export abstract class DexOnChainFetcher<T> extends MultiRequestFetcher {
   ): PricesObjWithMetadata[string] {
     const response = getDefinedResponse(responses, dataFeedId);
     const metadata = this.getLiquidityAndSlippageMetadata(dataFeedId, response);
+    if (this.getName().includes("curve")) {
+      appendFileSync("./dupa", JSON.stringify({ metadata, dataFeedId }) + "\n");
```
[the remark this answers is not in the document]
koonopek: this pattern is named: dupa debugging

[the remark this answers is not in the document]
rpiszczatowski: I know and appreciate this pattern but I thought it doesn't leave dev computer. Will you leave this debugging?

packages/oracle-node/index.ts
rpiszczatowski: I've read this before. This means that numbers having less than 9e15 digits will be displayed in logs with standard notation (i.e. exact values) and only bigger numbers will be displayed with scientific notation. But scientific notation does not show the PRECISE value of the number. This is why I used the word PRECISE.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +38,220 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
-    const lastPriceFromCache = getRawPrice(pairedToken);
-    if (!lastPriceFromCache) {
-      throw new Error(`Cannot get last price from cache for: ${pairedToken}`);
-    }
-
+    const lastPriceFromCache = getRawPriceOrFail(pairedToken);
     return lastPriceFromCache.value;
   }
+
+  private normalizeRatio(assetId: string, ratioBigNumber: any) {
```
rpiszczatowski, line 249: I don't like the signature of this function. Can we say what exactly we take here as a parameter?

packages/utils/src/common/multicall.ts
```diff
@@ -0,0 +1,62 @@
...
+  const resultsInOrder = new Array(calls.length);
+  for (const result of results.results["this"].callsReturnContext) {
+    if (!result.success) {
+      throw new Error(
+        `Failed to execute method of multicall contract: ${contract.address} method name: ${result.methodName}`
+      );
+    }
+    const index = functionToIndex[result.reference!];
+    if (result.decoded) {
+      resultsInOrder[index] = result.returnValues[0];
```
rpiszczatowski, line 53: What if there is more than one return value? In uni v3 we use the second return value sometimes.

packages/oracle-node/src/fetchers/dex-on-chain/DexOnChainFetcher.ts
```diff
@@ -24,7 +25,7 @@ export abstract class DexOnChainFetcher<T> extends MultiRequestFetcher {
       `slippage calculation not implemented for ${this.getName()}`
     );
   }
-  calculateSpotPrice(_assetId: string, _response: T): number {
```
rpiszczatowski, line 27: Do we really need to change this? How is it possible that the spot price is not a number in a single fetcher when it can be in other fetchers? A similar question about liquidity.

koonopek: It can be number, I think we discussed it already in the end game we want every number to be a string and operation done via `SafeNumber`. So if I'm touching any fetcher I try to refactor already.

packages/utils/src/common/multicall.ts
koonopek: Hmm I have to check that because in my case result.decoded always returns false

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +38,220 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+        new Decimal(ratio.toString())
+      );
+      return CurveFetcher.convertUsdToTokenAmount(
+        assetId,
+        this.normalizeRatio(assetId, ratioBigNumber).toString()
+      );
+    }
+  }
+
+  private static convertUsdToTokenAmount(
```
rpiszczatowski, line 222: I don't understand the meaning of the return value of this function. Sometimes we pass in the paired token as an `assetId` and this paired token can be USDC which has 6 decimals. But we multiply the tokens number by 10^18 always.

As a side note, I don't really understand the idea of CURVE_PRECISION_DECIMALS. In the description, it says it is a safe amount to exchange without both loss of precision and slippage. But such an amount depends on decimals of the swapped token. Think what would happen if we had a pool USDC to USDT both tokens having 6 decimals...

To me, the conceptually simpler model is the one we use in uni fetchers where in the config each token has its decimals defined and the code uses these values when needed.

koonopek: We have such tokens where one token has 18 decimals and the other has 6 decimals and it works.
We use then ratioMultiplier (which is configured per token in PoolsConfig) to make these values precision proper (this is how curveFetcher was designed previously and I did not change it).

I think I can refactor it to multiply by the proper number of decimals already in convertUsdToTokenAmount, instead of later.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +38,220 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+      ratioMultiplier,
       pairedTokenIndex,
+      pairedToken,
     } = this.poolsConfig[assetId];
 
-    const curveFactory = new Contract(address, abi, provider);
+    const curvePool = new Contract(address, abi, provider);
 
-    const ratioBigNumber = (await curveFactory[functionName](
+    const amountInBaseToken = await this.getAmountInBaseToken(
```
rpiszczatowski, line 125: It looks like amountInBaseToken is a number of subtokens (e.g. wei) and amountInPairedToken represents something I don't understand. See my other comment.

It causes us to divide by ratioMultiplier in seemingly random places below.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +38,220 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
 
     return {
-      ratio: ratioFloat,
-      assetId,
+      ratio: this.normalizeRatio(assetId, ratioBigNumber),
+      baseTokenSupply: new Decimal(baseTokenSupply.toString()).div(
+        CURVE_PRECISION_DECIMAL
+      ),
+      pairedTokenSupply: new Decimal(pairedTokenSupply.toString()).div(
+        CURVE_PRECISION_DECIMAL
```
rpiszczatowski, line 186: Is it correct if paired token is USDC (6 decimals)?

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: Yeah you are right this brings a lot of confusion, I will refactor this ratioMultiplier this to decimals per token. Let's keep consistent.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: done

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: done

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: nope, GOOD catch. Done

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: done

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +38,220 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+  private async getAmountInBaseToken(
+    assetId: string,
+    curvePool: Contract,
+    functionName: string,
+    tokenIndex: number,
+    pairedTokenIndex: number
+  ) {
+    try {
+      return CurveFetcher.convertUsdToTokenAmount(assetId);
+      // if this is first iteration of fetcher we won't have assetId price in DB yet, thus we have to fetch it additionally
```
cehali, line 204: But I think that in the first iteration, we also won't have asset price in USD which is used in `doCalculateSpotPrice` -> `const pairedTokenPrice = new Decimal(this.getPairedTokenPrice(assetId))`

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +38,220 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+      {
+        slippageAsPercent: sellSlippage.toString(),
+        direction: "sell",
+        simulationValueInUsd: CURVE_SLIPPAGE_PARAMS.amountInUsd.toString(),
+      },
+    ];
+  }
+
+  calculateSpotPrice(assetId: string, response: CurveFetcherResponse): number {
+    return this.doCalculateSpotPrice(assetId, response.ratio);
```
cehali, line 106: What is the purpose of extraction functionality to `doCalculateSpotPrice`?

packages/oracle-node/test/fetchers/curve-multi-block.spec.ts
```diff
@@ -53,10 +60,19 @@ describe("Curve Multi Block", () => {
...
+    await saveMockPriceInLocalDb(1000, "STETH");
     await clearPricesSublevel();
     provider = new MockProvider();
     const [wallet] = provider.getWallets();
+    const multicall = await deployContract(wallet, multicall3Json as any);
+    RedstoneCommon.overrideMulticallAddress(multicall.address);
     mockContract = await deployMockContract(wallet, abi);
+    await mockContract.mock.balances
+      .withArgs(0)
+      .returns(BigNumber.from(toCurvePrecision("2")));
```
cehali, line 72: Why do you put big numbers in those mocks, you can just pass string here and it should work the same way

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: Yes, but if we have one source like for VST. We will never have VST event after 100 iterations :)

packages/utils/src/common/multicall.ts
koonopek: I think that it only has length more than 1 when result.decoded =false. This is my assumption after reading evm-multicall code. I will add assert, so we will be alarmed quickly if I'm wrong

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: Because we are using doCalculateSpotPrice in diffrent context were we dont have whole `response` object

packages/oracle-node/test/fetchers/curve-multi-block.spec.ts
koonopek: I will try

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
cehali: But what is the difference between getting price of `assetId` in `CurveFetcher.convertUsdToTokenAmount(assetId);` and getting the price of `assetId` in `new Decimal(this.getPairedTokenPrice(assetId))`. The first one is used in try and the second one is used in catch, if I understand correctly both will fail in the same conditions

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +38,209 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+          function: "balances",
+          params: [tokenIndex],
+        },
+        {
+          function: "balances",
+          params: [pairedTokenIndex],
+        },
+        {
+          function: functionName,
+          params: [tokenIndex, pairedTokenIndex, amountInBaseToken],
```
cehali, line 166: Is there a possibility that the big buy or big sell amount would be lower than the standard `CURVE_PRECISION_DECIMAL`?

koonopek: yes

packages/oracle-node/test/fetchers/curve-multi-block.spec.ts
koonopek: It works, nice, done!

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: The problem is that this fetcher relies on its own prices. There is no problem if fetcher depends of prices of other fetchers like USD/ETH etc.

Let's assume that we are in VST case, were we have one fetcher `curve-frax-multi-block` 
This fetcher to calculate liqudiity slippage and spot price, requires (frax and vst) price.
It will get frax from db after N iteration. But will be still missing vst and he can get it in n-iteration, because there is no other fetcher which serves VST dataFeed. Thus we have to fetch this in the same fetcher (the try catch is for this). Because we are situation where some fetcher requires VST and also produces VST.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: discussed online

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +38,218 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+    _assetId: string,
+    response: CurveFetcherResponse
+  ): RedstoneTypes.SlippageData[] {
+    const smallBuy = response.ratio;
+    const smallSell = new Decimal(1).div(response.ratio);
+
+    const buySlippage = response.bigBuy
+      .sub(smallBuy)
+      .abs()
+      .div(response.bigBuy)
```
rpiszczatowski, line 83: We should divide by `smallBuy` here instead because the price impact should be expressed in terms of the price we started with.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +38,218 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+      bigSell,
+    ] = await RedstoneCommon.multiCallOneContract(
+      curvePool,
+      [
+        {
+          function: functionName,
+          params: [
+            tokenIndex,
+            pairedTokenIndex,
+            CURVE_PRECISION_DECIMAL.toString(),
```
rpiszczatowski, line 153: Once we have `tokenDecimals` wouldn't it be more appropriate to use it here instead of `CURVE_PRECISION_DECIMAL`?

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +38,218 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+        {
+          function: "balances",
+          params: [tokenIndex],
+        },
+        {
+          function: "balances",
+          params: [pairedTokenIndex],
+        },
+        {
+          function: functionName,
```
rpiszczatowski, line 165: Is this `buy` or `sell`? From the result name (`bigBuy`) it looks like `buy` but we actually give the base token here and get the paired token so we rather `sell`, right?

koonopek: Everything is okey it is buy, same as first call:
```
          function: functionName,
          params: [
            tokenIndex,
            pairedTokenIndex,
            CURVE_PRECISION_DECIMAL.toString(),
          ],
          ```

File: packages/oracle-node/src/fetchers/curve/CurveFetcher.ts [hunk 21]
```diff
@@ -32,50 +38,218 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+      ],
+      blockTag?.toString()
     );
 
+    const decimalsRatio = tokenDecimals / pairedTokenDecimals;
+
     return {
-      ratio: ratioFloat,
-      assetId,
+      ratio: this.normalizeRatio(assetId, ratioBigNumber),
```
rpiszczatowski, line 179: I believe that here it would be more readable if we simply use our `decimalsRatio` and use it directly to 'normalize' the result as we do below. Especially because `normalizeRatio` works only in one direction. Actually, I'd get rid of `normalizeRatio` altogether.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: True

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
rpiszczatowski: I'd argue that you need the VST price only to calculate slippage and liquidity, not the spot price. The approach I took in Uni v3 is that if the price was not yet available and so I cannot even prepare params that will allow me to fetch data required for slippage calculation I simply don't perform this part of multicall.

The result of my approach is that the very first dataPoint for VST may not have slippage metadata. But this would make the code cleaner, I think. just something to consider.
Despite your decision, it would probably be better to stay consistent among fetchers.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: True!

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: Rename, i will leave it cause we are using it twice

koonopek: parseGetDyResult

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +29,207 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+      ],
+      blockTag?.toString()
     );
 
+    const decimalsRatio = tokenDecimals / pairedTokenDecimals;
+
     return {
-      ratio: ratioFloat,
-      assetId,
+      ratio: new Decimal(ratioBigNumber.toString()).div(pairedTokenDecimals),
```
rpiszczatowski, line 166: We should divide by `decimalsRatio` - the same case below.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: Why do you think that this would be cleaner?

koonopek: Oh i know why it can't be done like this, because in future we will require liquidity and slippage to be fetched or fail https://github.com/redstone-finance/redstone-oracles-monorepo/pull/919/files#r1284543518

rpiszczatowski: Because this function returns two different things depending on the context. It takes the price from the latest round as a source of computation or the price from current multicall result as a source. To me, this is logically different thing.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: Nope then we should do `.div(tokenDecimals).mul(decimalsRatio)`

rpiszczatowski: My bad. I forgot that we swap `baseTokenDecimals` of the first token here.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
rpiszczatowski: Now it looks good.

rpiszczatowski (dismissed):

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
rpiszczatowski: I accept. Once the other PR is merged we have to have uniswap v3 already adjusted.

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -32,50 +29,207 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
+    priceFloat?: number | string
+  ): string {
+    return new Decimal(CURVE_SLIPPAGE_PARAMS.amountInUsd)
+      .div(priceFloat ?? getRawPriceOrFail(assetId).value)
+      .mul(decimals)
+      .round()
+      .toString();
+  }
+
+  private doCalculateSpotPrice(
```
hatskier, line 220: I would rename to  smth like `calculateSpotPriceUsingRatio`, adding `do` doesn't make the code more readable IMHO

packages/oracle-node/src/fetchers/curve/curve-fetchers-config.ts
```diff
@@ -118,7 +129,8 @@ export const curveFetchersConfig: Record<string, PoolsConfig> = {
       pairedTokenIndex: 0,
       pairedToken: "USDT",
       provider: ethereumProvider,
-      ratioMultiplier: 10 ** 12,
+      tokenDecimals: 1e18,
```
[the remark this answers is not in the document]
koonopek: Nope, it isn't javascript can handle very big numbers if we don't have decimals points and a small number in it:
`1e100` is okay?

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: Agree

packages/oracle-node/src/fetchers/curve/curve-fetchers-config.ts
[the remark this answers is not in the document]
rpiszczatowski: The only caveat I can see here is inconsistency with other fetchers.

packages/rpc-providers/src/TransactionDeliveryMan.ts
```diff
@@ -267,6 +268,21 @@ export class TransactionDeliveryMan {
     return Math.max(...rewardsPerBlockForPercentile);
   }
 
+  /**
```
rpiszczatowski, line 271: Isn't this already on main?

rpiszczatowski (approved):

packages/oracle-node/src/fetchers/curve/curve-fetchers-config.ts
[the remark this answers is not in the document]
hatskier: Then I would just suggest renaming, because it's not token decimals, it's a multuplier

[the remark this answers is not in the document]
koonopek: Will do in next PR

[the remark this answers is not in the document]
koonopek: done in this one

rpiszczatowski (approved):
I am wondering if it wouldn't be more natural to rename all '...Ratio' as '...Price'.

PR #914, opened 2023-08-03
feat: classic model in integration tests

rpiszczatowski:
- basic on chain relayer tests
- verification of prices in both adapter and price feed contracts
- verification of prices in a core model way
- enable more eslint rules: be more strict in types in string interpolation, be more restricted in allowed usage of `any`, enforce some rules on async functions, etc.
- remove random isNaN checks


packages/integration/src/prices-updates-node-failures.ts
```diff
@@ -0,0 +1,161 @@
...
+    priceFeedContractAddress,
+    expectedPrices
+  );
+
+  process.exit();
+};
+
+configureCleanup(stopAll);
+
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
```
cehali, line 160: if you don't want to have this ignore you can not use main function and write `(async () => {})`

packages/integration/src/framework/relayer-manager.ts
```diff
@@ -70,52 +93,104 @@ export const startRelayer = async (
...
+  (priceFeedContractAddress: string, expectedPrices: PriceSet) =>
+  async (): Promise<boolean> =>
+    await runHardhatScript(
+      "test/monorepo-integration-tests/scripts/verify-mock-prices-in-price-feed.ts",
+      {
+        PRICE_FEED_CONTRACT_ADDRESS: priceFeedContractAddress,
+        PRICES_TO_CHECK: `${JSON.stringify(expectedPrices)}`,
+      },
+      "price-feed-contract"
+    );
```
cehali, line 132: Do we want to check all prices? It seems that this test only checks if a single price exists

packages/integration/src/prices-updates-node-failures.ts
```diff
@@ -0,0 +1,161 @@
...
+    expectedPrices
+  );
+
+  // stop all nodes and verify that prices are not being updated
+  stopOracleNode(oracleNodeInstance2);
+  stopOracleNode(oracleNodeInstance3);
+  setMockPrices({ __DEFAULT__: 44 });
+  expectedPrices = { BTC: 44 };
+  await verifyPricesNotInCacheService(allCacheLayers, expectedPrices);
+  await verifyPricesNotOnChain(
```
cehali, line 116: I am not sure if I understand this verification. If you have already pushed the price on the chain it would be there, it could be stale but you should be able to get it.

rpiszczatowski: This code assets that price is not equal to 44. The price in cache service is 43 at this point.

packages/integration/src/prices-updates-node-failures.ts
rpiszczatowski: You mean something like?
```
(async () => {
  await main();
})();
```
Eslint complains here too. But the current code at least clearly states its intentions. I'll leave it as is.

packages/on-chain-relayer/test/monorepo-integration-tests/scripts/deploy-mock-price-feed.ts
```diff
@@ -0,0 +1,15 @@
...
+  const priceFeedContractFactory = await ethers.getContractFactory(
+    "PriceFeedWithoutRoundsMock"
+  );
+  const priceFeedContract = await priceFeedContractFactory.deploy();
+  const adapterContractAddress = process.env.ADAPTER_CONTRACT_ADDRESS;
+  if (!adapterContractAddress) {
+    throw new Error("deploying price feed mock contract requires ADAPTER_CONTRACT_ADDRESS");
+  }
+  priceFeedContract.setAdapterAddress(adapterContractAddress);
+  fs.writeFileSync("price-feed-contract-address.txt", priceFeedContract.address);
```
hatskier, line 14: Can we move `price-feed-contract-address.txt` to a constant and reuse it?

packages/integration/src/prices-updates-node-failures.ts
```diff
@@ -0,0 +1,161 @@
...
+    adapterContractAddress,
+    priceFeedContractAddress,
+    expectedPrices
+  );
+
+  // start stopped cache service, stop another one and verify everything works
```
hatskier, line 122: Can you add info that we also start oracle nodes again?

hatskier (commented):
Great job!

packages/integration/.eslintrc.js
```diff
@@ -10,7 +10,9 @@ module.exports = {
   },
   plugins: ["@typescript-eslint"],
   rules: {
-    "@typescript-eslint/no-floating-promises": ["warn"],
-    "@typescript-eslint/no-unnecessary-condition": "error"
+    "@typescript-eslint/no-floating-promises": "warn",
+    "@typescript-eslint/no-unnecessary-condition": "error",
+    "no-return-await": "off",
```
cehali, line 15: I see that you talke with Alex about it 😄

cehali (approved):

PR #917, opened 2023-08-03
feat: improve long dry runs

cehali:


packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -61,35 +63,63 @@ const getMissingDataFeedsInDataPackages = (
...
+      const dataFeedValueFromLocal =
+        dataPackagesFromLocal[dataFeedId][0].dataPoints[0].value;
+      const deviation = MathUtils.calculateDeviationPercent({
+        newValue: dataFeedValueFromLocal,
+        prevValue: dataPoint.value,
+      });
+      const currentDeviationPerDataFeed = deviationsPerDataFeed[dataFeedId];
+      if (
+        deviation > currentDeviationPerDataFeed ||
+        !currentDeviationPerDataFeed
```
rpiszczatowski, line 118: It would be more natural to swap alternatives and compare `current...` only if it exists.

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -1,26 +1,29 @@
 import { MathUtils } from "redstone-utils";
+import {
+  ALL_FEEDS_DATA_FEED_ID,
+  DeviationsPerDataFeed,
+} from "./run-long-price-propagation-core-test";
 
-const omittedDataFeeds = ["___ALL_FEEDS___", "crvUSDBTCETH"];
+interface DataPackages {
+  [dataFeedId: string]: Array<DataPackage>;
```
rpiszczatowski, line 8: From what I see in our code we do not use `Array` notation for arrays, but 99.999% of the time we use `[]` syntax. Is there any specific reason not to use it here and below?

packages/integration/src/framework/check-deviations.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+    const deviation = deviationsPerDataFeed[dataFeedId] as number;
+    if (deviation > maxPercentageValueDifference) {
+      console.log(
+        `Value deviation for ${dataFeedId} is bigger than maximum (${maxPercentageValueDifference}%) - ${deviation}%`
+      );
+      deviationsBiggerThanAllowed += 1;
+    }
+
+    if (deviationsBiggerThanAllowed > 0) {
```
rpiszczatowski, line 29: Wasn't the idea to throw an error only after logging ALL the discrepancies? If so then this `if` statement should be placed outside of the loop. If not you don't actually need the `deviationsBiggerThanAllowed` variable, you can throw straight away in the `if` statement above.

rpiszczatowski (dismissed):

packages/integration/src/framework/compare-data-packages.ts
cehali: But it won't be `[dataFeedId: string]: number | undefined`. Value is a number of whole key-value field doesn't exist in the object. In practice it is the same, when you want to get the value of `dataFeedId` you will receive undefined, but it could be misleading when reading

packages/integration/src/framework/compare-data-packages.ts
cehali: I started to using it as I think it is more readable because `[]` could be easily missed

packages/integration/src/framework/compare-data-packages.ts
cehali: I don't think I understand what `It would be more natural to swap alternatives and compare current... only if it exists.` means, could you provide an example? We have the empty object and in the first iteration we cannot check if the value from this object is bigger than calculated, this is why if the current value is undefined we just put the first value in the object.

packages/integration/src/framework/check-deviations.ts
cehali: Logging is in another function called `log-deviations.ts`. First, we log all deviations, then we log all deviations bigger than allowed and if any deviations are bigger we throw an error.

packages/integration/src/framework/compare-data-packages.ts
koonopek: I don't have opinion on that, but being conistent is a plus

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -1,26 +1,29 @@
 import { MathUtils } from "redstone-utils";
+import {
+  ALL_FEEDS_DATA_FEED_ID,
+  DeviationsPerDataFeed,
+} from "./run-long-price-propagation-core-test";
 
-const omittedDataFeeds = ["___ALL_FEEDS___", "crvUSDBTCETH"];
+interface DataPackages {
```
koonopek, line 7: Could we move it to redstone utils types plis, i think it will be used in other scripts too

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -1,26 +1,29 @@
 import { MathUtils } from "redstone-utils";
+import {
+  ALL_FEEDS_DATA_FEED_ID,
```
koonopek, line 3: This one could be move to new index.ts in utils packages like RedstoneCommons

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -61,35 +63,63 @@ const getMissingDataFeedsInDataPackages = (
...
+      );
+      deviationsPerDataFeed[ALL_FEEDS_DATA_FEED_ID] = deviationsFromBigPackage;
       continue;
     }
-    const valueFromFirstDataPackages = dataFeedObject[0].dataPoints[0].value;
-    const dataPointsFromSecond = secondDataPackage[dataFeedId];
-    const deviations = dataPointsFromSecond.map(({ dataPoints }) =>
+    const dataPointValueFromLocal =
+      dataPackagesFromLocal[dataFeedId][0].dataPoints[0].value;
+    const deviations = dataFeedObject.map(({ dataPoints }) =>
```
koonopek, line 89: I think we should also compare here metadata.source and compare every source price. Then we would know if some source is corrupted. Now we are loosing this info.

In next PR we could also compare liquditiy and slippage info :)

hatskier: That's a great idea! We can even compare slippage and liquidity just in case

cehali: Moved to another task

packages/integration/src/framework/compare-data-packages.ts
cehali: Moved to another task

packages/integration/src/framework/compare-data-packages.ts
cehali: Moved to another task

rpiszczatowski (approved):

PR #919, opened 2023-08-04
feat: add slippage based excluding for oracle nodes

hatskier:


packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -287,6 +289,50 @@ export default class PricesService {
...
+    sourceName: string
+  ) {
+    const slippages = price.sourceMetadata[sourceName]?.slippage;
+    const amountToCheck = config.simulationValueInUsdForSlippageCheck;
+    const checkedDirections: Record<string, boolean> = {};
+    const errors: string[] = [];
+
+    // We always check slippage if it's available, but if it's not - we don't
+    // throw here. We need to throw on the fetcher level if slippage is not
+    // fetched properly
```
rpiszczatowski, line 303: I have a comment to this comment :)
Even though the feature is young we already have different handling of this.

Basically, to calculate slippage, we need the previous price not only of the quote token but also of the base token. The question is what do we do if for some reason we have no price for the base token? One approach is to divide our communication in two parts: first, fetch the base token price and then, based on the response, fetch slippage (Curve fetcher approach - still not merged). The other one is to admit that in some rare situations, we may not be able to fetch slippage (uni v3 fetcher approach - on main already).

For the lack of a quote token price, I proposed some time ago an additional fetching precondition check for fetchers to avoid making requests when we know that we cannot make use of the response.

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -287,6 +289,50 @@ export default class PricesService {
...
+    if (!slippages) {
+      return;
+    }
+
+    for (const slippage of slippages) {
+      if (slippage.simulationValueInUsd === amountToCheck) {
```
rpiszczatowski, line 309: At the moment each fetcher has its own methodology to calculate slippage and each of them makes independent decisions on the simulation amount. Maybe, instead of this check (or except it), we should implement using the common config value in our fetchers, supposedly as a separate task.

rpiszczatowski (approved):
Good job.

packages/oracle-node/src/fetchers/PricesService.ts
koonopek: Are you suggesting that we use a shared value for slippage simulation across all fetchers? If that's the case, I completely agree. It would streamline the configuration process and reduce the risk of errors.

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -251,9 +251,11 @@ export default class PricesService {
           recentPrices: recentPricesInLocalDBForSymbol,
         });
 
+        this.assertAcceptableSlippageForSource(price, sourceName);
+
```
koonopek, line 255: General comment, name of this function is quite unfortunate `calculateAggregatedValues` it is doing a lot of more stuff then it suggests

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -287,6 +289,50 @@ export default class PricesService {
...
+    // throw here. We need to throw on the fetcher level if slippage is not
+    // fetched properly
+    if (!slippages) {
+      return;
+    }
+
+    for (const slippage of slippages) {
+      if (slippage.simulationValueInUsd === amountToCheck) {
+        const { direction, slippageAsPercent } = slippage;
+        if (Number(slippageAsPercent) > config.maxAllowedSlippagePercent) {
```
koonopek, line 311: Maybe we could use `SafeNumber` here, it has some advantages over simply casting string to `Number`

packages/oracle-node/src/fetchers/PricesService.ts
rpiszczatowski: Alex's change actually enforces common value for simulating slippage if I understand it correctly.

hatskier: I would leave the check as is, but we can also add using common value for slippage calculation in fetchers. I will create a separate task for it

packages/oracle-node/src/fetchers/PricesService.ts
hatskier: We discussed it on the call

packages/oracle-node/src/fetchers/PricesService.ts
hatskier: I would say it's quite stratightforward that aggregation can have a custom logic (including sources filtering and outliers removing). Do you have a proposal for another better name for this function?

koonopek: For me agrregation is for example median. Also we use this convention in code https://github.com/redstone-finance/redstone-oracles-monorepo/blob/0e44f37c36305048151d6d217529f5cb919ea04f/packages/oracle-node/src/aggregators/median-aggregator.ts.
`computeValidAggregates`

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -319,15 +323,15 @@ export default class PricesService {
 
     // We need to check both trade directions
     for (const direction of ["buy", "sell"]) {
```
koonopek, line 325: This would be more idiomatic 
```
type B = "a" | "b";
let b: B;
for (const a of ["a", "b"] as const){
    b = a;
}
```
using as const to defined static array instead of casting. Or in this case even `Object.keys(checkedDirections)`

koonopek (approved):

rpiszczatowski (approved):

packages/oracle-node/src/fetchers/PricesService.ts
hatskier: I agree, but I would propose to make this change in a separate PR with creating a single type for "buy" | "sell" and reusing it everywhere across fetchers and PricesService

PR #931, opened 2023-08-07
fix: adjust symbol to id for TRX in bybit

cehali:


packages/oracle-node/src/fetchers/ccxt/symbol-to-id/bybit.json
```diff
@@ -6,7 +6,7 @@
   "CRV": "CRV/USDT:USDT",
   "AAVE": "AAVE/USDT:USDT",
   "STETH": "STETH/USDT",
-  "TRX": "TRX/USDT",
+  "TRX": "TRX/USDT:USDT",
```
rpiszczatowski, line 9: Is it correct that SHIB and PEPE do not have the :USDT suffix?

rpiszczatowski (approved):

PR #939, opened 2023-08-08
fix: add dependencies required to build images

cehali:


packages/utils/package.json
```diff
@@ -13,6 +13,7 @@
   },
   "dependencies": {
     "decimal.js": "^10.4.3",
+    "ethereum-multicall": "^2.19.0",
```
koonopek, line 16: I think that this could be peer dependency, will fix in another PR

koonopek (approved):

PR #940, opened 2023-08-08
feat: Add slipapge to uniswap v2

koonopek:


packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -28,20 +33,17 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
-    } = this.poolsConfig[assetId];
+    const { address, provider, pairedToken, pairedTokenDecimalsMultiplier } =
+      this.poolsConfig[assetId];
 
     const curvePool = new Contract(address, abi, provider);
 
     const amountInBaseToken = await this.getAmountInBaseToken(assetId);
 
-    const amountInPairedToken = this.convertUsdToTokenAmount(
+    const amountInPairedToken = await convertUsdToTokenAmount(
```
rpiszczatowski, line 43: await?

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
```diff
@@ -44,19 +55,19 @@ export class UniswapV2LikeFetcher extends DexOnChainFetcher<Reserves> {
...
     const {
       isSymbol0CurrentAsset,
       reserve0Normalized,
       reserve1Normalized,
       pairedTokenPrice,
-    } = this.getParamsFromResponse(assetId, response);
+    } = response;
 
     const balanceRatioBigNumber = isSymbol0CurrentAsset
       ? reserve1Normalized.mul(decimalsMultiplier).div(reserve0Normalized)
```
rpiszczatowski, line 73: All this normalizing stuff was relevant when we had whole numbers only (i.e. BigNumber) but now, once you introduced Decimals here, we can simplify those calculations.

I just have spent a couple of minutes to understand what is going on here. In particular, the innocently looking `bigNumberToFloat` method that divides by 10**18 under the hood was quite a surprise to me.

packages/oracle-node/test/fetchers/uniswap-v2-like.spec.ts
```diff
@@ -71,3 +84,4 @@ describe("UniswapV2Like", () => {
     ]);
   });
 });
+// X * symbolDecimals * (18 - symbolDecimals)
```
rpiszczatowski, line 87: What is this?

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
koonopek: :)

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -113,25 +113,21 @@ export class CurveFetcher extends DexOnChainFetcher<CurveFetcherResponse> {
...
     const { tokenDecimalsMultiplier: tokenDecimals } =
       this.poolsConfig[assetId];
     try {
-      return this.convertUsdToTokenAmount(assetId, tokenDecimals);
+      return convertUsdToTokenAmount(
+        assetId,
+        tokenDecimals,
+        DEFAULT_AMOUNT_IN_USD_FOR_SLIPPAGE
+      );
       // if this is first iteration of fetcher we won't have assetId price in DB yet, thus we have to fetch it additionally
```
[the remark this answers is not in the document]
koonopek: True

packages/oracle-node/test/fetchers/uniswap-v2-like.spec.ts
koonopek: some crap

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
```diff
@@ -27,8 +22,8 @@ export interface PoolsConfig {
 }
 
 interface UniswapV2LikeResponse {
-  reserve0Normalized: BigNumber;
-  reserve1Normalized: BigNumber;
+  reserve0Float: Decimal;
```
rpiszczatowski, line 25: We don't use the suffix `Float` normally.

rpiszczatowski: I see now, that the value stored here is actually what you called 'normalizedReserve'. I'd stick to this name as reserve0 to me is strictly related to the value specified in uniswap protocol which is subtokens amount, not tokens amount.

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
```diff
@@ -44,58 +50,132 @@ export class UniswapV2LikeFetcher extends DexOnChainFetcher<Reserves> {
...
+    const priceSellWithoutSlippage = this.calculateSpotPrice(assetId, response);
+    const priceBuyWithoutSlippage = this.calculateSpotPrice(assetId, {
+      ...response,
+      isSymbol0CurrentAsset: !response.isSymbol0CurrentAsset,
+    });
+
+    // sell slippage
+    const reserve0SupplyIncrease = convertUsdToTokenAmount(
+      assetId,
+      10 ** symbol0Decimals,
```
rpiszczatowski, line 110: Seeing this, I am wondering whether PrecisionScaler isn't overkill. Theoretically, if we want such a class, we should use it here, either by changing the semantics of the `convertUsdToTokenAmount` method to return tokens instead of subtokens or passing the scaler to that method instead of the 'raw' multiplication factor.

rpiszczatowski, line 110: Here, you don't need `symbol0Decimals` but the decimals of the token that is your `assetId`. It can be any of 0 or 1 depending on the pool.

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
```diff
@@ -44,58 +50,132 @@ export class UniswapV2LikeFetcher extends DexOnChainFetcher<Reserves> {
...
+      DEFAULT_AMOUNT_IN_USD_FOR_SLIPPAGE
+    );
+    const priceWithSellSlippage = this.calculateSpotPrice(assetId, {
+      ...response,
+      reserve0Float: response.reserve0Float.add(
```
rpiszczatowski, line 115: This doesn't make sense. You have to simulate the trade because trade changes BOTH reserves. (Alternatively, you can use the CMM formula to calculate the price after such a simulated trade yourself)

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
koonopek: I don't agree we dont have to use it now everywhere it is possible.

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
koonopek: Nope it doesnt change both reservers, only one. We discussed it here: https://discord.com/channels/786251205008949258/1138211326716555394/1138412793029214218 .
What is CMM? I'm just simulating what will happen if I increase supply of one of the tokens in pool.

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
koonopek: I saw that we used float prefix in some fetchers.
Subtokens is your naming convention not propagated in our repository. So I would avoid introducing it.

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
```diff
@@ -44,58 +50,132 @@ export class UniswapV2LikeFetcher extends DexOnChainFetcher<Reserves> {
...
       isSymbol0CurrentAsset,
-      reserve0Normalized,
-      reserve1Normalized,
+      reserve0Float: reserve0Normalized,
+      reserve1Float: reserve1Normalized,
       pairedTokenPrice,
-    } = this.getParamsFromResponse(assetId, response);
+    } = response;
 
     const reserve = isSymbol0CurrentAsset
```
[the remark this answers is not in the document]
koonopek: Don't get it

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
koonopek: And I think it is not overkill, it helps me. And make code more readable. What are disadvantages?

koonopek: Nice catch!

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
rpiszczatowski: I don't impose the name `subtokens`. What I say is that the `reserve0Float` contains the number of tokens in the pool whereas in uniswap v2 protocol the value `reserve0` is the number of subtokens. This may be confusing in the context of uniswap v2 fetcher.

My suggestion is in this situation to stick to the suffix `Normalized` that suggests that some adjustments were done on the original value. Suffix `Float` suggests (not correctly in my opinion) that we have a float number here and says noting about what is the value stored.

This sounds a bit like a Hungarian notation to me.

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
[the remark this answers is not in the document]
rpiszczatowski: The variable `reserve` contains the value that is better described by the name `pairedTokenReserve`.

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
rpiszczatowski: No disadvantages. This was rather a comment than a change suggestion. Although I'd make use of `PrecisionScaler` here if I were you :)

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
rpiszczatowski: Typo. I meant CFMM (constant function market maker)

koonopek: But in my opinion I have used CFMM

rpiszczatowski: discussed on the call

packages/oracle-node/src/fetchers/uniswap-v2-like/UniswapV2LikeFetcher.ts
```diff
@@ -44,58 +50,152 @@ export class UniswapV2LikeFetcher extends DexOnChainFetcher<Reserves> {
...
+        simulationValueInUsd: DEFAULT_AMOUNT_IN_USD_FOR_SLIPPAGE.toString(),
+      },
+    ];
   }
 
-  getParamsFromResponse(assetId: string, response: Reserves) {
-    const { reserve0, reserve1 } = response;
+  // token1Amount_received = reserve_1 -(reserve_0 * reserve_1)/(reserve_0 + token0Amount)
+  // token0Amount/ token1Amount_received
+  private getToken0SwapPrice(
```
rpiszczatowski, line 153: if you renamed params in the following way:
- reserve0Scaled -> reserveIn
- reserve1Scaled -> reserveOut
- token0Amount   -> amountIn
then you can have a single function as trade is symmetric in terms of token0 and token1. On the call side you decide which is tokenIn and which is tokenOut then.

rpiszczatowski (approved):

PR #944, opened 2023-08-09
feat: add slippage calculation to TraderJoeV2OnChainFetcher

rpiszczatowski:


packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
+                .returnValues[1]
+            ),
+            pairedTokenConfig.decimals,
+            baseTokenConfig.decimals,
+            basePriceInPaired!
+          )
+        : undefined,
+      sellSlippage: basePrice
+        ? this.getSlippageInPercent(
+            swapAmountBase!,
```
cehali, line 124: Why you use here `!` when you `swapAmountBase` can be undefined based on:
```js
const swapAmountBase = basePrice
      ? this.convertDollarsToSubTokens(
          SLIPPAGE_SIMULATION_AMOUT,
          basePrice,
          baseTokenConfig.decimals
        )
      : undefined;
```
I feel like we are losing typing here, we are responsible for calculating this so we can decide what to do with an undefined value but now we just assume that it won't be undefined.

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
+            this.decimal(
+              routerResults.find((x) => x.reference === "slippage_sell")!
+                .returnValues[1]
+            ),
+            baseTokenConfig.decimals,
+            pairedTokenConfig.decimals,
+            pairedPriceInBase!
```
cehali, line 131: The same situation as with `swapAmountBase` from https://github.com/redstone-finance/redstone-oracles-monorepo/pull/944/files#r1289821323

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
+      buySlippage: basePrice
+        ? this.getSlippageInPercent(
+            swapAmountPaired,
+            this.decimal(
+              routerResults.find((x) => x.reference === "slippage_buy")!
+                .returnValues[1]
+            ),
+            pairedTokenConfig.decimals,
+            baseTokenConfig.decimals,
+            basePriceInPaired!
```
cehali, line 119: The same situation as with swapAmountBase from https://github.com/redstone-finance/redstone-oracles-monorepo/pull/944/files#r1289821323

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
 
-    const binId = await lbPair.getActiveId();
-    const binStep = await lbPair.getBinStep();
+    const multicallResult: ContractCallResults =
+      await this.createMulticallInstance().call(callContexts);
+    const poolResults = multicallResult.results.poolContract.callsReturnContext;
+    const routerResults =
+      multicallResult.results.routerContract?.callsReturnContext;
 
     return {
```
cehali, line 103: Can we move calculations from the return statement?

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
   ): number {
     const { binId, binStep } = response;
     const binStepDivided = binStep.div(BIN_STEP_DIVIDER);
     const binStepPlusOne = ONE_AS_DECIMAL.add(binStepDivided);
     const binIdSerialized = binId.minus(BIN_ID_DEFAULT_SUBTRACTION);
 
-    return binStepPlusOne.pow(binIdSerialized).toNumber();
+    const baseInPairedPrice = binStepPlusOne.pow(binIdSerialized).toNumber();
+    return (
+      baseInPairedPrice *
```
cehali, line 148: Maybe calculate it outside the return statement and use `decimal.js` function `.mul`

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
+    const tokensSent = this.convertSubTokensToTokens(
+      subtokensSent,
+      tokensSentDecimals
+    );
+    const tokensGot = this.convertSubTokensToTokens(
+      subtokensGot,
+      tokensGotDecimals
     );
+    const actualPrice = tokensSent.dividedBy(tokensGot);
+    return actualPrice.minus(spotPrice).abs().dividedBy(actualPrice).times(100);
```
cehali, line 200: Please move `100` to a separate variable, preferred to the top of the file with capital letters name

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
+    const baseTokenConfig = this.getBaseTokenConfig(assetId);
+    const pairedTokenConfig = this.getPairedTokenConfig(assetId);
+    const basePrice = getLastPrice(assetId)?.value;
+    const pairedPrice = getLastPrice(pairedTokenConfig.symbol)?.value;
+    if (!pairedPrice) {
+      throw new Error(
+        `missing price for quote token ${pairedTokenConfig.symbol}`
+      );
+    }
+    const basePriceInPaired = basePrice ? basePrice / pairedPrice : undefined;
```
koonopek, line 77: What do you think about using abstraction like https://github.com/redstone-finance/redstone-oracles-monorepo/blob/80f8540b234247c237e396dd10dc684c61d97472/packages/oracle-node/src/fetchers/curve/CurveFetcher.ts#L152
Iny my opinion it simplify core logic. Allow to avoid this constant ifs => `?:``

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
+            this.decimal(
+              routerResults.find((x) => x.reference === "slippage_buy")!
+                .returnValues[1]
+            ),
+            pairedTokenConfig.decimals,
+            baseTokenConfig.decimals,
+            basePriceInPaired!
+          )
+        : undefined,
+      sellSlippage: basePrice
```
koonopek, line 122: Yeah I think we should create some helper class here, it is beacming a little bit unreadable

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
     const { binId, binStep } = response;
     const binStepDivided = binStep.div(BIN_STEP_DIVIDER);
     const binStepPlusOne = ONE_AS_DECIMAL.add(binStepDivided);
     const binIdSerialized = binId.minus(BIN_ID_DEFAULT_SUBTRACTION);
 
-    return binStepPlusOne.pow(binIdSerialized).toNumber();
+    const baseInPairedPrice = binStepPlusOne.pow(binIdSerialized).toNumber();
+    return (
+      baseInPairedPrice *
+      getLastPrice(this.getPairedTokenConfig(assetId).symbol)!.value
```
koonopek, line 149: It will return some weird shit if getLastPrice returns undefined. Use getRawPriceOrFail to have better error message in node. Then just "couldn't cast undefined to number" or smth like this.
Also to keep precision, you could do: `binStepPlusOne.pow(binIdSerialized).mul(getRawPriceOrfail(this.getPairedTokenConfig(assetId).symbol))`

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
+    );
+  }
+
+  override calculateLiquidity(
+    _assetId: string,
+    _response: MulticallResult
+  ): string {
+    return undefined!;
```
koonopek, line 157: Why you just don't change signature of calcualteLiquidity. This way we are avoiding typescript checks. Thus making our software less secure.

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
+          {
+            direction: "sell",
+            simulationValueInUsd: String(SLIPPAGE_SIMULATION_AMOUT),
+            slippageAsPercent: response.sellSlippage.toString(),
+          },
+        ];
+  }
+
+  private convertSubTokensToTokens(amount: Decimal, decimals: number): Decimal {
+    return amount.dividedBy(this.decimal(10).toPower(decimals));
```
koonopek, line 181: I have create little helper class for this in RedstoneUtils: https://github.com/redstone-finance/redstone-oracles-monorepo/blob/3bb852eb33d7a60a6468d5dae48751827862a1f5/packages/utils/src/math/index.ts#L37 
(still not merged)

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
koonopek: Can we be consistent with this Decimal names sometimes I see `mul` and `div` sometimes `dividedBy` and `times`

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
+    const tokensGot = this.convertSubTokensToTokens(
+      subtokensGot,
+      tokensGotDecimals
     );
+    const actualPrice = tokensSent.dividedBy(tokensGot);
+    return actualPrice.minus(spotPrice).abs().dividedBy(actualPrice).times(100);
+  }
+
+  private decimal(bigNumberLike: {}): Decimal {
+    return new Decimal(BigNumber.from(bigNumberLike).toString());
```
koonopek, line 204: We have a lot of function like that maybe we could export it to common packages

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,311 @@
...
+  private getPairedTokenConfig(assetId: string) {
+    return this.getTokenConfig(assetId, !this.isXBase(assetId));
+  }
+
+  private convertDollarsToSubTokens(
+    amountInDolars: number,
+    tokenPrice: number,
+    tokenDecimals: number
+  ): Decimal {
+    return this.decimal(
```
koonopek, line 236: I think I create already helper for that, could you re use it:
 https://github.com/redstone-finance/redstone-oracles-monorepo/blob/3bb852eb33d7a60a6468d5dae48751827862a1f5/packages/oracle-node/src/fetchers/SlippageAndLiquidityCommons.ts#L6

koonopek (changes requested):

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
rpiszczatowski: Because it won't. This condition can be deduced by static code analysis, TS compiler simply doesn't see this. It requires help. BTW. this is one of the reasons why ! exists, i.e. helping the compiler.

So why it CANNOT be `undefined` here:
As you can see `swapAmountBase` is undefined iff (if and only if) `basePrice` is undefined and I use `swapAmountBase!` only in code paths where `basePrice !== undefined`.

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
rpiszczatowski: Sure, they were a little simpler when I put them directly in the return statement :)

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
rpiszczatowski: We could, but it would require remembering what we agreed to and even before that to agree on something. @koonopek feel free to impose a convention and I'll stick to it.

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
rpiszczatowski: I wanted to have all the variables const, but I can change those couple of ternary operators to a single if statement that will initialize everything that should be initialized when the `basePrice` is available and then use another if statement while handling response. This would make the code more similar to the one you pointed. I'll do that.
The !s will disappear as a consequence. @cehali :)

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
rpiszczatowski: `getLastPrice` cannot return `undefined` as we called it above and it returned something that is not undefined. But you are right, I'll reuse the originally fetched value to avoid confusion. Also, I've changed already this code to return string and so it makes sense to use Decimals all to the very end of calculations.

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
rpiszczatowski: Because when this code was written it was not allowed. Will change now. Thanks.

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
rpiszczatowski: Because when this code was written it was not allowed. Will change now. Thanks.

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
rpiszczatowski: This code is not merged yet.

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
rpiszczatowski: I now. Once your code is merged, we can change this.

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
rpiszczatowski: I thought about it, but on the other hand, this is an obvious oneliner. Maybe next time? :)

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
koonopek: It merged

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
koonopek: Please add it to common like all other functions, there will be no next time probably

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
koonopek: Merged

koonopek (commented):
Some minior style comments, you have to resolve conflicts anyway

packages/oracle-node/src/fetchers/dex-on-chain/DexOnChainFetcher.ts
```diff
@@ -45,10 +45,13 @@ export abstract class DexOnChainFetcher<T> extends MultiRequestFetcher {
...
     try {
-      metadata.liquidity = this.calculateLiquidity(
+      const liquidity = this.calculateLiquidity(
         dataFeedId,
         response
-      ).toString();
+      )?.toString();
+      if (liquidity) {
+        metadata.liquidity = liquidity;
+      }
```
cehali, line 54: WDYT 
```js
      const liquidity = this.calculateLiquidity(
        dataFeedId,
        response
      );
      if (liquidity) {
        metadata.liquidity = liquidity.toString();
      }
```
?

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,319 @@
...
-      binId: new Decimal(binId),
-    };
+    const binStep = this.decimal(
+      poolResults.find((r) => r.reference === "getBinStep")!.returnValues[0]
+    );
+    const binId = this.decimal(
+      poolResults.find((r) => r.reference === "getActiveId")!.returnValues[0]
+    );
+    let buySlippage;
+    let sellSlippage;
```
cehali, line 115: Maybe add `Decimal | undefined` as type?

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,319 @@
...
+    if (baseTokenPrice) {
+      buySlippage = this.getSlippageInPercent(
+        swapAmountPaired,
+        this.decimal(
+          routerResults.find((x) => x.reference === "slippage_buy")!
```
cehali, line 120: Maybe change `x` to `result`?

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,319 @@
...
+        {
+          reference: "slippage_sell",
+          methodName: "getSwapOut",
+          methodParameters: [
+            poolAddress,
+            swapAmountBase.toString(),
+            this.isXBase(assetId),
+          ],
+        },
+      ];
```
cehali, line 308: Could we extract a function called something like `buildContractCalls` which will return this array?

packages/oracle-node/src/fetchers/evm-chain/avalanche/trader-joe-on-chain/trader-joe-v2/TraderJoeV2OnChainFetcher.ts
```diff
@@ -1,76 +1,319 @@
...
+            methodParameters: [],
+          },
+          {
+            reference: "getBinStep",
+            methodName: "getBinStep",
+            methodParameters: [],
+          },
+        ],
+      },
+    ];
```
cehali, line 287: Could we extract a function called something like `buildContractCalls` which will return this array?

cehali (approved):
LGTM

PR #951, opened 2023-08-10
feat: improve long dry runs, add sources comparing

cehali:


.github/workflows/monorepo-integration-tests.yml
```diff
@@ -17,6 +17,8 @@ jobs:
           node-version: 18.15
       - name: Install deps
         run: yarn
+      - name: Build
+        run: yarn build
```
rpiszczatowski, line 21: At the moment this isn't required, I think unless we start using the compiled version to run the tests.

cehali: It doesn't work without so I would say it is required 😄

packages/integration/tsconfig.json
```diff
@@ -12,6 +12,15 @@
   "references": [
     {
       "path": "../utils"
+    },
+    {
+      "path": "../protocol"
+    },
+    {
+      "path": "../cache-service"
```
rpiszczatowski, line 20: Why this is needed? Do we refer in our integration tests code to cache-service code or oracle-node code?

packages/integration/src/framework/print-all-deviations.ts
```diff
@@ -10,9 +10,10 @@ export const printAllDeviations = (
 ) => {
   const dataFeeds = Object.keys(deviationsPerDataFeed);
   for (const dataFeedId of dataFeeds) {
-    if (dataFeedId === ALL_FEEDS_DATA_FEED_ID) {
+    const ALL_FEEDS_KEY = consts.ALL_FEEDS_KEY as string;
```
rpiszczatowski, line 13: Is this cast required?

packages/integration/tsconfig.json
cehali: We use it but we don't rebuild it, I change slightly code in `cache-service` (imports) and couldn't run tests because `cache-service` wasn't up-to-date

packages/integration/src/framework/print-all-deviations.ts
cehali: Ohh, after rebuilding `protocol` package it isn't required, will fix it

cehali: Sorry, it is required
image

.github/workflows/monorepo-integration-tests.yml
cehali: Without it I get 
image

packages/integration/tsconfig.json
cehali: Probably it is my local problem as cache etc is already built and no rebuild is initialized

.github/workflows/monorepo-integration-tests.yml
cehali: I changed .yml file to use built version

rpiszczatowski: Now I see, we import a constant from protocol...

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -65,52 +68,116 @@ const getMissingDataFeedsInDataPackages = (
...
+      for (const [source, { value }] of Object.entries(
+        sourceMetadataFromProd
+      )) {
+        if (sourcesRemovedFromManifest.includes(source)) {
+          console.log(`Source ${source} is removed from manifest, skipping`);
+          continue;
+        }
+        const valueFromLocal = Number(
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+          sourceMetadataFromLocal?.[source]?.value ?? 0
```
rpiszczatowski, line 155: `sourceMetadataFromLocal` cannot be undefined here.

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -65,52 +68,116 @@ const getMissingDataFeedsInDataPackages = (
...
+        const valueFromProd = Number(value);
+        if (valueFromLocal && valueFromProd) {
+          const deviation = MathUtils.calculateDeviationPercent({
+            newValue: valueFromLocal,
+            prevValue: valueFromProd,
+          });
+          deviationsPerSource[source] = Math.max(
+            deviation,
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+            deviationsPerSource?.[source] ?? 0
```
rpiszczatowski, line 166: `deviationsPerSource` cannot be undefined here.

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -122,8 +189,8 @@ const compareValuesFromBigPackageAndLocalCache = (
       const dataFeedValueFromLocal =
         dataPackagesFromLocal[dataFeedId][0].dataPoints[0].value;
       const deviation = MathUtils.calculateDeviationPercent({
-        newValue: dataFeedValueFromLocal,
-        prevValue: dataPoint.value,
+        newValue: Number(dataFeedValueFromLocal),
+        prevValue: Number(dataPoint.value),
```
rpiszczatowski, line 193: This change is not required.

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -65,52 +68,116 @@ const getMissingDataFeedsInDataPackages = (
...
+  return Math.max(...deviations);
+};
+
+const compareSourcesValuesFromProdAndLocal = (
+  dataPackagesFromLocal: DataPackages,
+  dataFeedId: string,
+  allFeedObjectsFromProd: DataPackagePlainObj[]
+) => {
+  const deviationsPerSource: DeviationsPerSource = {};
+  const dataPoints = dataPackagesFromLocal[dataFeedId][0].dataPoints;
```
rpiszczatowski, line 136: We reuse this name in for loop. Maybe you can inline it here not to have two variables with the same name in the single method because you use it only once.

rpiszczatowski: This is just a note, I am ok with that. Just wanted to put you attention on it?

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -65,52 +68,116 @@ const getMissingDataFeedsInDataPackages = (
...
+      )) {
+        if (sourcesRemovedFromManifest.includes(source)) {
+          console.log(`Source ${source} is removed from manifest, skipping`);
+          continue;
+        }
+        const valueFromLocal = Number(
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+          sourceMetadataFromLocal?.[source]?.value ?? 0
+        );
+        const valueFromProd = Number(value);
```
rpiszczatowski, line 157: No need for wrapping in Number here and above

rpiszczatowski (dismissed):

packages/integration/src/framework/compare-data-packages.ts
cehali: It would be undefined in the first iteration

packages/integration/src/framework/compare-data-packages.ts
cehali: source metadata values are strings
image

packages/integration/src/framework/compare-data-packages.ts
cehali: Fixed

packages/integration/src/framework/compare-data-packages.ts
cehali: Fixed

packages/integration/src/framework/compare-data-packages.ts
cehali: Fixed

packages/integration/src/framework/compare-data-packages.ts
cehali: Fixed

packages/integration/src/framework/compare-data-packages.ts
rpiszczatowski: I know. But `calculateDeviationPercent` accepts strings as well as numbers.

rpiszczatowski (dismissed):

packages/integration/src/framework/compare-data-packages.ts
cehali: Yes, you are right, fixed

PR #961, opened 2023-08-11
feat: implemented cron scheduling in on-chain-relayer

hatskier:


packages/on-chain-relayer/src/core/update-conditions/cron-condition.ts
```diff
@@ -0,0 +1,41 @@
+import cronParser from "cron-parser";
+import { RelayerConfig } from "../../types";
+
+// TODO: move to a common place, redstone-utils
+const MS_IN_ONE_MINUTE = 60000;
```
jakub-wojciechowski, line 5: Not a strong opinion, but why reinventing the wheel. Cannot we simply use moment.js to translate time units.

cehali (approved):
LGTM

packages/on-chain-relayer/src/types.ts
```diff
@@ -47,7 +49,8 @@ export interface RelayerConfig {
   adapterContractType: string;
   expectedTxDeliveryTimeInMS: number;
   isArbitrumNetwork: boolean;
-  fallbackOffsetInMinutes?: number;
+  fallbackOffsetInMinutes: number;
+  fallbackOffsetInMS: number;
```
Lukasz2891, line 53: do we need both of these values in the config?

packages/on-chain-relayer/src/types.ts
```diff
@@ -47,7 +49,8 @@ export interface RelayerConfig {
   adapterContractType: string;
   expectedTxDeliveryTimeInMS: number;
   isArbitrumNetwork: boolean;
-  fallbackOffsetInMinutes?: number;
```
Lukasz2891, line 50: is the gelato-relayer still able to build :)?

packages/on-chain-relayer/src/core/update-conditions/cron-condition.ts
```diff
@@ -0,0 +1,39 @@
...
+  const lastExpectedUpdateTime = cronParser
+    .parseExpression(config.cronExpression || "", {
+      // We move current time a bit back for the case with fallback
+      currentDate: new Date(currentTimestamp - config.fallbackOffsetInMS),
+    })
+    .prev();
+
+  const timeElapsedSinceLatestUpdate = currentTimestamp - lastUpdateTimestamp;
+
+  // We want to update price only if there are no update txs after the previous
```
koonopek, line 18: So if we had cron on 8:00 everyday and had update because of deviation at 4:00, we want update prices?

koonopek (approved):

packages/on-chain-relayer/src/core/update-conditions/cron-condition.ts
hatskier: Yes

packages/on-chain-relayer/src/types.ts
hatskier: Yep :)

packages/on-chain-relayer/src/types.ts
hatskier: Will check as a separate task

PR #965, opened 2023-08-14
feat: slippage calculation in balancer multi fetcher

rpiszczatowski:


packages/oracle-node/test/fetchers/balancer-multi.spec.ts
```diff
@@ -1,41 +1,66 @@
...
 
   afterAll(async () => {
     await closeLocalLevelDB();
   });
 
-  test("should properly fetch data", async () => {
+  // despite of the supposedly synchronous interface waffle mock setup has to be awaited, quiet the compiler
+  const asAwaitable = <T = void>(awaitableObject: any): Promise<T> => {
+    return awaitableObject as unknown as Promise<T>;
+  };
```
cehali, line 53: I think I have seen it already, maybe move it to some common file?

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
```diff
@@ -52,58 +66,317 @@ export class BalancerMultiFetcher extends DexOnChainFetcher<DeltaResponse> {
...
-      BALANCER_VAULT_ADDRESS,
-      BalancerVaultAbi,
-      this.provider
+  override async makeRequest(assetId: string): Promise<BalancerResponse> {
+    const { baseTokenDecimals, pairedTokenDecimals, swaps, tokenAddresses } =
+      this.configs[assetId];
+    const pairedToken =
+      this.configs[assetId].tokenToFetch ?? this.underlyingToken;
+    const pairedTokenPrice = Number(getRawPriceOrFail(pairedToken).value);
+    const baseTokenPrice = getLastPrice(assetId)?.value;
```
cehali, line 80: Don't we want to use here `getRawPriceOrFail` also and not have `if (baseTokenPrice)` below?

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
```diff
@@ -52,58 +66,317 @@ export class BalancerMultiFetcher extends DexOnChainFetcher<DeltaResponse> {
...
-    const deltas = await vaultContract.callStatic.queryBatchSwap(
-      SwapType.SwapExactIn,
+    const callContexts = this.buildContractCallContext(
+      swapAmountBase,
+      swapAmountPaired,
       swaps,
+      tokenAddresses
+    );
+
+    const multicallResult = await this.createMulticallInstance().call(
```
cehali, line 102: I think extracting should be in `calculate...` function not in `makeRequest`. In most of the fetchers we use `makeRequest` function only to fetch data and extraction is done in separate functions

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
```diff
@@ -52,58 +66,317 @@ export class BalancerMultiFetcher extends DexOnChainFetcher<DeltaResponse> {
...
-    if (!tokenPrice) {
-      throw new Error(`Cannot get last price from cache for: ${tokenToFetch}`);
-    }
-    const tokenPriceAsDecimal = new Decimal(tokenPrice.value);
+  private convertSubTokensToTokens(amount: Decimal, decimals: number): Decimal {
+    return amount.div(this.decimal(10).toPower(decimals));
+  }
 
-    return tokenPriceAsDecimal.mul(ratioSerialized).abs().toNumber();
+  private decimal(bigNumberLike: {}): Decimal {
```
cehali, line 264: If returned value from the multicall is big number in hex we can directly pass it to `new Decimal`

koonopek: Also can we use some better type then `{}`?

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
```diff
@@ -52,58 +66,317 @@ export class BalancerMultiFetcher extends DexOnChainFetcher<DeltaResponse> {
...
-    _response: DeltaResponse
-  ): number {
-    throw new Error(
-      `calculateLiquidity is not implemented for ${this.getName()}`
+  private convertDollarsToSubTokens(
+    amountInDolars: number,
+    tokenPrice: number,
+    tokenDecimals: number
+  ): Decimal {
+    return this.decimal(
```
koonopek, line 273: We already have common function like this, is there specific reason you don't want to reause it?

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
```diff
@@ -52,58 +66,317 @@ export class BalancerMultiFetcher extends DexOnChainFetcher<DeltaResponse> {
...
+    swapsForSpot: SwapConfig[],
+    newInitialAmount: string
+  ): SwapConfig[] {
+    // we only want to change initial trade amount
+    const swapsCopy = swapsForSpot.slice();
+    swapsCopy[0] = { ...swapsCopy[0], amount: newInitialAmount };
+    return swapsCopy;
+  }
+
+  private reverseTokensOrder(swap: SwapConfig): SwapConfig {
```
koonopek, line 293: Shouldnt be static?

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
```diff
@@ -52,58 +66,317 @@ export class BalancerMultiFetcher extends DexOnChainFetcher<DeltaResponse> {
...
+      );
+      callContexts.push(
+        this.createBatchSwapCall(SLIPPAGE_SELL_LABEL, slippageSellParams),
+        this.createBatchSwapCall(SLIPPAGE_BUY_LABEL, slippageBuyParams)
+      );
+    }
+    return [
+      {
+        reference: VAULT_CALLS,
+        contractAddress: BALANCER_VAULT_ADDRESS,
```
koonopek, line 376: Why you don't want to use https://github.com/redstone-finance/redstone-oracles-monorepo/blob/3c52e558d51345c4bacaed343ee4f25d47fb8d9e/packages/utils/src/common/multicall.ts#L10?
You could then remove this `labels` logic and resolving output. There would be less => less bugs :P

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
```diff
@@ -52,58 +66,317 @@ export class BalancerMultiFetcher extends DexOnChainFetcher<DeltaResponse> {
...
-      ratioSerialized = ratioSerialized.mul(priceMultiplier);
-    }
-    const tokenToFetch =
-      this.configs[dataFeedId].tokenToFetch ?? this.underlyingToken;
-    const tokenPrice = getLastPrice(tokenToFetch);
-    if (!tokenPrice) {
-      throw new Error(`Cannot get last price from cache for: ${tokenToFetch}`);
-    }
-    const tokenPriceAsDecimal = new Decimal(tokenPrice.value);
+  private convertSubTokensToTokens(amount: Decimal, decimals: number): Decimal {
```
koonopek, line 260: Again you could use `PriceScaler` here it is doing exactly the same thing. However, has one advaantage that you could rember what is is doing and don't have to read code like this every time.

koonopek: Then if you would have to touch many fetcher and the would use common Classes as helper it will be easier to introduce changes.

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
```diff
@@ -52,58 +66,317 @@ export class BalancerMultiFetcher extends DexOnChainFetcher<DeltaResponse> {
...
+    const tokensSent = this.convertSubTokensToTokens(
+      subtokensIn,
+      tokensInDecimals
+    );
+    const tokensGot = this.convertSubTokensToTokens(
+      subtokensOut,
+      tokensOutDecimals
+    );
+    const actualPrice = tokensSent.div(tokensGot);
+    return actualPrice
```
koonopek, line 241: We already have common function for this: https://github.com/redstone-finance/redstone-oracles-monorepo/blob/3c52e558d51345c4bacaed343ee4f25d47fb8d9e/packages/oracle-node/src/fetchers/SlippageAndLiquidityCommons.ts#L18.
Some advantages as in next comment.

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
```diff
@@ -1,13 +1,25 @@
...
+import {
+  CallReturnContext,
+  ContractCallContext,
+  Multicall,
+} from "ethereum-multicall";
+import { RedstoneTypes } from "redstone-utils";
 
 export const BALANCER_VAULT_ADDRESS =
   "0xBA12222222228d8Ba445958a75a0704d566BF2C8";
+const SLIPPAGE_SIMULATION_AMOUT = 10_000;
```
koonopek, line 17: We already have const for this, lets use it. It will be easier then to change :)
https://github.com/redstone-finance/redstone-oracles-monorepo/blob/3c52e558d51345c4bacaed343ee4f25d47fb8d9e/packages/oracle-node/src/fetchers/SlippageAndLiquidityCommons.ts#L4

koonopek (changes requested):

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
rpiszczatowski: Honestly, I don't understand the idea here. Let's look at some code that is not mine, MaverickFetcher. This code parses protocol response to 5 values, namely:

- price
- smallSellAmountOutScaled
- smallBuyAmountOutScaled
- bigSellAmountOutScaled
- bigBuyAmountOutScaled

All of them are already parsed and adjusted somehow, e.g. Scaled suffix means that the value is already converted from subtokens to tokens and price is the value returned from protocol squared and multiplied by constant to reflect base to paired ratio.
In the calculateXXX methods, only the 'final' transformations are done like multiplying price by paired token price to get the price in dollars or calculation of slippage from spot price and parsed values from swap simulations. Could you explain to me what is the reasoning behind such a division of responsibilities? Also, what exactly should be done with the response from protocol in makeRequest and what in calculateXXX methods?

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
rpiszczatowski: It is rather a BigNumber object, not a string. This function is supposed to work also with strings and numbers. As I have it defined in a couple of places now, I can extract it or maybe get rid of it once I adjust the code according to other comments in this PR.

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
rpiszczatowski: Honestly, I see not much value in making methods static if they are not part of the class interface. Also, we do not have this After marking all of the methods in Uniswap fetcher as static the annoying thing was that each time I needed to call the static method I had to prefix it with class name. The latest refactor of this class (change of the class name) generated lots of changes (in terms of diff size). I just today spent quite a lot of time resolving those conflicts on my branch with changing slippage implementation for Uniswap...

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
rpiszczatowski: Because multicall.ts:
- does not support returning multiple values from a single call
- shifts labels logic to indexing logic. Imagine what happens if you change calls order in multicall. Just from reading the code it is not obvious that such a change requires change in parsing values from response. The interface where I label the call first and then get the response by that label cannot be confused that way.

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
cehali: Yes, I see. Previously we used `fetchData` in non-DEX fetchers as a function which only calls e.g. `axios.get()`. Then we introduce DEXes fetchers with `makeRequest` and in my understanding, they should only fetch on-chain and all extractions, and calculations should be done in separate functions like `calculateSpotPrice` and `calculateLiquidity`. I don't know when it was mixed and I remember I wrote a similar thing in some other PR but probably it wasn't taken under consideration. I assume it is all right then

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
koonopek: Maybe, I'm not a fun of static too. But this is a rule that we established some time ago, that if we don't touch class fields, the method should be static, something like a `pure` modifier in solidity. I wouldn't change this rule based on a single PR. Let's follow conventions and if we want to change them let's discuss it separately from PR to avoid being biased.

koonopek (dismissed):

packages/oracle-node/src/fetchers/balancer-multi/BalancerMultiFetcher.ts
rpiszczatowski: fixed

koonopek (approved):
LGTM

PR #969, opened 2023-08-16
feat: adjust long dry runs with error values

cehali:


packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -129,11 +129,19 @@ const compareValuesFromSmallPackagesAndLocalCache = (
 ) => {
   const dataPointValueFromLocal =
     dataPackagesFromLocal[dataFeedId][0].dataPoints[0].value;
-  const deviations = allFeedObjectsFromProd.map(({ dataPoints }) =>
-    MathUtils.calculateDeviationPercent({
-      newValue: Number(dataPointValueFromLocal),
-      prevValue: Number(dataPoints[0].value),
-    })
+  const deviations = allFeedObjectsFromProd.reduce(
```
koonopek, line 132: Should we report missing / error values?? I think they should fail test too for example, when value is defined on prod but not on local.

cehali: This test is not for reporting missing values in data packages, monitoring is responsible for it. Here we only want to know if our new changes will not break node functionalities. It will throw an error if the value is defined on prod but not on local, we have an error `Missing data packages data package from local cache service`. It is in different checks, we first check missing packages and then compare values by calculating deviations. Changes in this PR only touch deviations calculation in which we sometimes miss values as a single node from five could have some issues.

koonopek: Thanks for the explanation!

koonopek (approved):

PR #971, opened 2023-08-16
feat: use-data-timestamp-instead-of-block-timestamp

koonopek:
Merged into #970


Lukasz2891 (approved):

packages/on-chain-relayer/test/update-conditions/should-update.test.ts
```diff
@@ -50,7 +53,11 @@ describe("should-update", () => {
       {
         dataPackages,
         valuesFromContract: biggerValueDiff,
-        lastUpdateTimestamp,
+        lastUpdateTimestamps: {
+          lastBlockTimestampMS: lastUpdateTimestamp,
+          // this timestamp should not be taken into account here
```
Lukasz2891, line 58: what does that mean ;)?

Lukasz2891 (approved):

packages/on-chain-relayer/test/update-conditions/should-update.test.ts
koonopek: That this timestamp shouldn't trigger this test (as it used to, that is why i set it explicitly to number which would used to trigger it)
trigger = fail

PR #972, opened 2023-08-16
feat: add wstETH from contract

cehali:


koonopek (commented):
Could you add unit test to LidoTokensRequestHandlers

koonopek (approved):

PR #973, opened 2023-08-16
feat: add sAVAX-bb-a-WAVAX-BPT from balancer

cehali:


packages/oracle-node/src/fetchers/evm-chain/avalanche/evm-fetcher/sources/balancer/BalancerRequestHandlers.ts
```diff
@@ -0,0 +1,92 @@
...
+    );
+
+    const firstTokenPrice = getRawPriceOrFail(tokensToFetch[0]);
+    const secondTokenPrice = getRawPriceOrFail(tokensToFetch[1]);
+
+    const secondPoolPrice = secondPoolReserve
+      .mul(secondTokenPrice.value)
+      .div(10 ** tokenDecimals);
+
+    const underlyingTokenPrice = Decimal.min(
```
koonopek, line 60: Why are we selecting min of two here?
Could you elaborate on this algorithm, it is possible that we will multplie reserve from one pool with price from second pool

koonopek (changes requested):
Missing unit tests, for new request handler

packages/oracle-node/src/fetchers/evm-chain/avalanche/evm-fetcher/sources/balancer/BalancerRequestHandlers.ts
cehali: It is a method consulted with Delta Prime and based on https://dev.gearbox.fi/oracle/curve-pricefeed. It will be also discussed with Balancer

PR #979, opened 2023-08-20
Add u token and stamp token to permaswap fetcher

ethever:
Hi, I am adding the U token and the STAMP token and relative tests to the permaswap fetcher. Thx for review advance, more details are willing to communicate with.


cehali (approved):
LGTM

PR #984, opened 2023-08-21
feat: use redstone memoize instead of nestjs Cache

koonopek:


packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -70,45 +70,18 @@ export class DataPackagesService {
...
-  async getMostRecentDataPackagesWithCache(
-    dataServiceId: string,
-    cacheManager: Cache
-  ): Promise<DataPackagesResponse> {
-    const cacheKey = `data-packages/latest-not-aligned-by-time/${dataServiceId}`;
-    const dataPackagesFromCache = await cacheManager.get<DataPackagesResponse>(
-      cacheKey
-    );
+  getMostRecentDataPackagesWithCache = RedstoneCommon.memoize(
+    (dataServiceId: string) =>
```
koonopek, line 80: It has to be  wrapped with arrow function, otherwise jest.spy doesnt work.
`bind(this)` also doesnt work

packages/utils/src/common/memoize.ts
```diff
@@ -1,13 +1,24 @@
 type Cache<T> = { promise?: Promise<T>; lastSet: number };
 
-export function memoize<T>(fn: () => Promise<T>, ttl: number) {
-  const cache: Cache<T> = { promise: undefined, lastSet: 0 };
+/**
+ * It also cache by arguments by calling toString on each one and then concat them = cache key
+ * Important if you want to cache also on objects you should implement toString method on this objects
+ * otherwise every object by default is converted to `[Object object]` string
+ */
+export function memoize<T, A extends unknown[]>(
```
cehali, line 8: I think maybe we could add more tests for the `memoize` function. It seems that currently we don't have any and we could check e.g. if it works correctly with hundreds of requests per minute as it could case in gateway

koonopek: Yeah, that is good point we should properly reset cache after TTL, I will add this and tests

koonopek: @cehali  done

hatskier:
Newer version has been published

packages/utils/test/common/memoize.spec.ts
```diff
@@ -0,0 +1,129 @@
+import { memoize } from "../../src/common/memoize";
+
+describe("memoize", () => {
+  jest.useFakeTimers();
+
+  afterEach(() => {
```
hatskier, line 6: Can you please add a test case with "more real" promise (e.g. sleep 500ms then resolve)?
Can you please also create a test case with different memoized functions `(e.g. const memoized1 = memoize(mockFn1, 1000); const memoize(mockFn2, 1000)` and test if they return different and correct results?

hatskier (changes requested):

packages/utils/src/common/memoize.ts
```diff
@@ -1,13 +1,43 @@
...
+  fn: (...args: A) => Promise<T>,
+  ttl: number,
+  cleanCacheInterval = 10_000
+): (...args: A) => Promise<T> {
+  const cache: Record<string, MemoizeCache<T>> = {};
```
hatskier, line 13: Nice!

packages/utils/src/common/memoize.ts
```diff
@@ -1,13 +1,43 @@
...
 
-  return async () => {
-    if (cache.promise === undefined || Date.now() - cache.lastSet > ttl) {
-      cache.lastSet = Date.now();
-      cache.promise = fn();
+  setInterval(() => cleanStaleCacheEntries(cache, ttl), cleanCacheInterval);
```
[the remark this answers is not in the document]
koonopek: I was thinking about it, however it has one big drawback, that if we will call like in cache service memoized function very many times in the short period of time. We will have to iterate over whole cache and check for stale ttls, this is O(cache_size) thus  is some kind of overhead. Running it every 10 seconds removes that overhead, and we are using native mechanics of nodeJS, so we are not moving anything new :)

packages/utils/src/common/memoize.ts
```diff
@@ -1,13 +1,43 @@
...
+
+  return async (...args: A) => {
+    const cacheKey = args.map((arg) => arg!.toString()).join();
```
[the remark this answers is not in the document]
koonopek: It will handle better but still not perfectly cause it won't handle sorting keys in Object, which can lead to hard to catch bugs. I would rather treat it explicit and if we need support for Object just add it. Serializing every object to json can be quite inefficient especially for large objects :)

packages/utils/src/common/memoize.ts
```diff
@@ -1,13 +1,43 @@
...
+    if (!cache[cacheKey] || Date.now() - cache[cacheKey].lastSet > ttl) {
+      cache[cacheKey] = {
+        lastSet: Date.now(),
+        promise: fn(...args).catch((err) => {
+          // don't propagate cache when promise resolves to error
+          cache[cacheKey].lastSet = -1;
```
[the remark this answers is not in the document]
koonopek: I like delete :)
1 or -1 doesnt have difference

packages/utils/test/common/memoize.spec.ts
koonopek: Good idea

packages/utils/src/common/memoize.ts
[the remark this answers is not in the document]
hatskier: Agreed to use JSON.stringify by defualt with assertion to strings be shorter than 10K symbols. We also want to have an option to set a custom cacheKeyBuilder, e.g.
```js
const memoized = memoize({
	fn: (args: any) => Promise<any>;
	cacheKeyBuilder?: (args: any) => string;
});

```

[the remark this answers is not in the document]
koonopek: Done

packages/utils/src/common/memoize.ts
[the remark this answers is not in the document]
koonopek: Done :)

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -70,45 +70,18 @@ export class DataPackagesService {
...
-  async getMostRecentDataPackagesWithCache(
-    dataServiceId: string,
-    cacheManager: Cache
-  ): Promise<DataPackagesResponse> {
-    const cacheKey = `data-packages/latest-not-aligned-by-time/${dataServiceId}`;
-    const dataPackagesFromCache = await cacheManager.get<DataPackagesResponse>(
-      cacheKey
-    );
+  getMostRecentDataPackagesWithCache = RedstoneCommon.memoize({
+    functionToMemoize: (dataServiceId: string) =>
```
hatskier, line 80: Mabe call it asyncFunctionToMemoize? Just an idea, we can also do it later

packages/utils/src/common/memoize.ts
```diff
@@ -1,13 +1,59 @@
...
+
+const cleanStaleCacheEntries = <T>(
+  cache: Record<string, MemoizeCache<T>>,
+  ttl: number
+) => {
+  const now = Date.now();
+  const cacheKeys = Object.keys(cache);
+
+  // we want to avoid slowing down
+  assertWithLog(
```
hatskier, line 49: Shouldn't we assert it after creating a new cache key? under the line `const cacheKey = cacheKeyBuilder(args);`?

hatskier (dismissed):

packages/utils/src/common/memoize.ts
koonopek: What is a difference?

packages/cache-service/src/data-packages/data-packages.service.ts
koonopek: I dont see any pros of adding async this is encoded in signature anyway, what is your view on that?

hatskier: I though about adding async just to make it more clear that memoization works only with functions that return promises. But you are right that it will also be suggested by typescript

packages/utils/src/common/memoize.ts
hatskier: The difference is in the potential number is errors in logs. It would work esp. in a different way if we throw an error. If we assert it only after creating the cache key we will run the assertion one time per each cache key, but if we assert it here, it can be called several times for the same cache key

hatskier (approved):

PR #986, opened 2023-08-21
chore: added GitHub action to trigger gelato function building

Lukasz2891:


.github/workflows/gelato-relayer-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+      run:
+        working-directory: packages/gelato-relayer
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install modules
+        run: yarn
+      - name: Build
+        run: yarn build
+        env:
+          ALCHEMY_ID: ${{ secrets.ALCHEMY_API_KEY }}
```
hatskier, line 22: Can't we build without ALCHEMY_API_KEY?

Lukasz2891: Yes, we can't :p It'd need to go deeper into the harthat env & plugin to verify it. For now, there's an assert in the harthat.config to prevent passing no Alchemy_id.

koonopek (dismissed):

.github/workflows/gelato-relayer-tests.yml
Lukasz2891: I'll remove the `assert` but there will be no check for running the function in local env.

koonopek (approved):

hatskier (approved):

PR #988, opened 2023-08-22
feat: add method to fetch prices from gateways

koonopek:
I haven't included arg: useDataPackagesWithAllFeeds
Discussed in discord: https://discord.com/channels/786251205008949258/1055853648242016397/1143457477988466808


packages/sdk/src/data-feed-values.ts
```diff
@@ -0,0 +1,84 @@
...
+const DEFAULT_DATA_SERVICE_ID = "redstone-main-demo";
+const DEFAULT_AGGREGATION_ALGORITHM = "median";
+
+export const getDataFeedValues = async (args: GetDataFeedValuesInput = {}) => {
+  const dataServiceId = args.dataServiceId ?? DEFAULT_DATA_SERVICE_ID;
+  const aggregationAlgorithm =
+    args.aggregationAlgorithm ?? DEFAULT_AGGREGATION_ALGORITHM;
+  const gatewayUrls = args.gatewayUrls ?? resolveDataServiceUrls(dataServiceId);
+
+  const dataPackagesPerFeed = await Promise.any<GatewayResponse>(
```
hatskier, line 24: I would propose to use requestDataPackages function in this function inistead. It will help to avoid logic duplication and will also force us to support different types of encoding (e.g. strings) for values

packages/sdk/src/data-feed-values.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export type GetDataFeedValuesOutput = Record<string, number>;
+type GatewayResponse = RedstoneTypes.DataPackageFromGatewayResponse;
+
+const DEFAULT_DATA_SERVICE_ID = "redstone-main-demo";
+const DEFAULT_AGGREGATION_ALGORITHM = "median";
+
+export const getDataFeedValues = async (args: GetDataFeedValuesInput = {}) => {
+  const dataServiceId = args.dataServiceId ?? DEFAULT_DATA_SERVICE_ID;
+  const aggregationAlgorithm =
```
hatskier, line 20: I would propose to unpack all args in the beginning of this function, e..g `const {dataServiceId, aggregationAlorithm, ...} = args;` to slightly reduce amount of the code

koonopek: It would would if I didn't have to use default values, which sometimes has to be computed. Also i could not simply merged object cause not all default values all static: `const gatewayUrls = args.gatewayUrls ?? resolveDataServiceUrls(dataServiceId);`

packages/sdk/src/data-feed-values.ts
koonopek: It could be, but `requestDataPackages` works differently for example it requires uniqueSignersCount, and if trim results according to that:

```
    parsedResponse[dataFeedId] = dataFeedPackagesSorted
      .slice(0, reqParams.uniqueSignersCount)
      .map((dataPackage: SignedDataPackagePlainObj) =>
        SignedDataPackage.fromObj(dataPackage)
      );
  ```
 I don't want to specify uniqueSignersCount in this endpoint also there is no good default for that (should be fetched from contract, but we are not in the context of contract when using this method)
Also, it seems that requestDataPackages is constructed for another purpose, and using it here seems a little bit drawn-out. Thus, the will change from other requirements. In GPT words:
"Within the realm of software development, it's not uncommon to encounter code structures that, at first glance, appear repetitious or seemingly redundant. However, it's crucial to understand that while two segments of code might appear identical today, they might have been crafted with distinct, future requirements in mind. As projects evolve, each segment might diverge to address specific nuances, functionalities, or constraints. By keeping them separate, we preemptively avoid entangling disparate functionalities and ensure that changes in one segment don't inadvertently impact another. In essence, what might seem like an inefficiency in the present is a strategic decision to accommodate anticipated differences in requirements, ensuring the maintainability and flexibility of the codebase in the long run."

cehali (approved):
LGTM

packages/sdk/src/data-feed-values.ts
hatskier: Sounds very convincing 😂

hatskier (approved):

PR #1003, opened 2023-08-24
fix: remove duplicated log in on-chain-relayer

koonopek:


packages/on-chain-relayer/src/run-relayer.ts
```diff
@@ -16,12 +16,16 @@ console.log(
 const runIteration = async () => {
   const adapterContract = getAdapterContract();
   const iterationArgs = await getIterationArgs(adapterContract);
+
   sendHealthcheckPing();
+  console.log(
+    `Update condition ${
+      iterationArgs.shouldUpdatePrices ? "" : "NOT "
+    }satisfied: ${iterationArgs.message}`
+  );
```
Lukasz2891, line 25: But now the log is removed from gelato. Gelato don't use the runner (run-relayer.ts) but their own.

Lukasz2891: but maybe that's not needed because it's passed to the returning value `{ canExec: false, message: ... }` :)

Lukasz2891 (approved):

packages/on-chain-relayer/src/run-relayer.ts
koonopek: so Maybe(YES) or Maybe(No)?

PR #1006, opened 2023-08-25
fix: wrong wsteth limit check id

cehali:


packages/oracle-node/manifests/data-services/arbitrum.json
```diff
@@ -283,9 +283,6 @@
         "maverick",
         "uniswap-v3-ethereum-on-chain-weth-500"
       ]
-    },
-    "WSTETH_LIMIT_CHECK": {
```
hatskier, line 287: Can we call it WSTETH_FUNDAMENTAL instead?

cehali: Sure

cehali: Done

hatskier (approved):

PR #1016, opened 2023-08-26
chore: decreased min valid sources percentage for main manifest

hatskier:


cehali:
I made PR with adding new sources for CANTO, removing this check could be bad for us, we already have only crucial tokens in the main manifest: https://github.com/redstone-finance/redstone-oracles-monorepo/pull/1017

hatskier:
Closing because of https://github.com/redstone-finance/redstone-oracles-monorepo/pull/1017/files

PR #1018, opened 2023-08-28
Add fetch with fallback method

koonopek:


koonopek:
Published on npm

packages/utils/src/common/fetch-with-fallback.ts
```diff
@@ -0,0 +1,14 @@
...
+  urls: string[];
+  axiosConfig?: AxiosRequestConfig<any>;
+};
+
+export const fetchWithFallbacks = async <T>(args: AxiosGetRequest) => {
+  const requests = args.urls.map((url) => axios.get<T>(url, args.axiosConfig));
+
+  const response = await Promise.any(requests);
+
+  return response;
```
cehali, line 13:
```js
return Promise.any(requests);
```

hatskier (approved):

packages/utils/src/common/fetch-with-fallback.ts
hatskier: I agree, we can also add a return type here (e.g. smth like `Promise<AxiosResponse<T>>` and remove async and await in the function

hatskier (approved):
Good job!

packages/utils/src/common/fetch-with-fallback.ts
koonopek: done

PR #1019, opened 2023-08-29
feat: run dry runs after Docker image deployment

cehali:


.github/workflows/run-dry-runs-after-images-deploy.yml
```diff
@@ -0,0 +1,71 @@
...
+
+  dry-runs-tests:
+    if: github.event.workflow_run.conclusion == 'success'
+
+    needs: get-short-sha
+    runs-on: LargeRunner
+    container:
+      image: public.ecr.aws/y7v2w8b2/redstone-oracle-node:${{ needs.get-short-sha.outputs.sha }}
+    steps:
+      - uses: actions/checkout@v3
```
hatskier, line 28: Let's please test if the code we are running is actually from docker container. I also think we should not do check out here

Lukasz2891 (approved):

PR #1032, opened 2023-09-01
feat: separate auction model and eip1559

koonopek:


packages/rpc-providers/src/TransactionDeliveryMan.ts
```diff
@@ -61,16 +67,18 @@ type TransactionDeliveryManOpts = {
    */
   percentileOfPriorityFee?: number;
 
+  isAuctionModel?: boolean;
+
   logger?: (text: string) => void;
 };
 
-type FeeStructure = {
+export type Eip1559Fee = {
```
Lukasz2891, line 75: We have Eip1559Fee defined here, but AuctionModelFee is defined in AuctionModelGasEstimator file ;)

packages/rpc-providers/src/TransactionDeliveryMan.ts
```diff
@@ -61,16 +67,18 @@ type TransactionDeliveryManOpts = {
    */
   percentileOfPriorityFee?: number;
 
+  isAuctionModel?: boolean;
```
Lukasz2891, line 70: Maybe an enum?

packages/rpc-providers/src/TransactionDeliveryMan.ts
```diff
@@ -2,24 +2,30 @@ import { ErrorCode } from "@ethersproject/logger";
...
   nonce: number;
   maxFeePerGas: number;
   result?: TransactionResponse;
 };
 
-type FeeHistoryResponse = { reward: string[] };
+export type FeeHistoryResponse = { reward: string[] };
 
-type GasOracleFn = (opts: TransactionDeliveryManOpts) => Promise<FeeStructure>;
+type GasOracleFn = (opts: TransactionDeliveryManOpts) => Promise<Eip1559Fee>;
```
Lukasz2891, line 26: Maybe to restore (renamed) `FeeStructure` and define it as `Eip1559Fee | AuctionModelFee`

packages/rpc-providers/src/GasEstimator.ts
```diff
@@ -0,0 +1,6 @@
+import { providers } from "ethers";
+
+export interface GasEstimator<T> {
```
Lukasz2891, line 3: Don't we expect here to have `T` extending `FeeStructure`?

packages/rpc-providers/src/TransactionDeliveryMan.ts
```diff
@@ -201,17 +216,17 @@ export class TransactionDeliveryMan {
 
   private async getFees(
     provider: providers.JsonRpcProvider
-  ): Promise<FeeStructure> {
+  ): Promise<Eip1559Fee> {
```
Lukasz2891, line 219: Is it important for `TransactionDeliveryMan` to specify the type here or can be used any `FeeStructure` (after having it restored ;p)?

Lukasz2891:
What do u think about creating unit tests to show the differences :)?

packages/rpc-providers/src/TransactionDeliveryMan.ts
koonopek: Nit pick: why?

packages/rpc-providers/src/TransactionDeliveryMan.ts
koonopek: Good catch

packages/rpc-providers/src/GasEstimator.ts
koonopek: good point!

packages/rpc-providers/src/TransactionDeliveryMan.ts
koonopek: fixed

cehali (approved):

packages/rpc-providers/src/TransactionDeliveryMan.ts
Lukasz2891: extendability ;) 

maybe - not so important for me ;p

Lukasz2891 (approved):

PR #1033, opened 2023-09-04
enable eslint

rpiszczatowski:


.eslintrc.js
```diff
@@ -0,0 +1,29 @@
+module.exports = {
+  extends: [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended-type-checked",
+    "plugin:prettier/recommended",
+    "plugin:workspaces/recommended",
+  ],
+  parser: "@typescript-eslint/parser",
+  plugins: ["@typescript-eslint", "workspaces"],
+  ignorePatterns: ["dist", "tools", "test", "jest.config*", ".eslintrc.js"],
```
koonopek, line 10: Why do you want ignore "test" I think that test should follow same rules as source code. (cause they are source code :))

packages/sdk/.eslintrc.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
```
koonopek, line 1: Can we have one file shared across whole monorepo (at least)?

koonopek:
Great!!!!

.eslintrc.js
rpiszczatowski: this was taken from some of existing config files. I absolutely agree that we shouldn't ignore tests.

packages/sdk/.eslintrc.js
rpiszczatowski: Unfortunately, I failed to set up parser options in the top level .eslintrc. As you see all the configs: rules, plugins, and basic ignores are defined in the top-level file.

A side benefit of this approach is that vscode doesn't check directories without parser options set and we can enable linging one directory at a time.

packages/utils/src/common/multicall.ts
```diff
@@ -43,14 +43,14 @@ export async function multiCallOneContract(
 
   const results = await multicall.call(multiCallContext, { blockNumber });
 
-  const resultsInOrder = new Array(calls.length);
+  const resultsInOrder: unknown[] = new Array(calls.length);
```
Lukasz2891, line 46: I don't know if the rule was really needed when we change `any` for `unknown` ;p I think the global goal of the rule is to specify types ;p 

But that's just a free comment ;)

packages/integration/package.json
```diff
@@ -8,17 +8,19 @@
   "license": "MIT",
   "scripts": {
     "run-ts": "ts-node",
-    "lint": "eslint",
+    "lint": "eslint . --ext .ts",
```
Lukasz2891, line 11: In the other places there's `lint` without `--ext .ts`, isn't it? is it ok? What's the difference ;)?

rpiszczatowski: I'll remove this. This was copied from another project. Basically, it only looks at .ts files but this is not something we want as eslint is also run from IDE where this option is not passed. It is better to shrink the scope to ts files in .eslintrc.js file if needed so that IDE works the same way as cmdline.

packages/utils/src/common/multicall.ts
rpiszczatowski: The difference between `unknown` and `any` in typescript is crucial. `any` skips type checking and `unknown` forces the user of the code to cast the variable to the appropriate type before using making it impossible to simply read any property of it by accident.

Of course, it is always better to use proper types than `unknown` but sometimes it is hard or impossible. Here, in particular, this is the caller's responsibility to know what the types of returned values are. Variable `resultsInOrder` can be an array of anything. It is a proper usage of the `unknown` type in TS.

packages/sdk/.eslintrc.js
```diff
@@ -0,0 +1,23 @@
...
+  ],
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    project: "tsconfig.json",
+    tsconfigRootDir : __dirname,
+    sourceType: 'module',
+  },
+  plugins: ["@typescript-eslint", "workspaces"],
+  ignorePatterns: ['.eslintrc.js', 'dist', 'jest.config*'],
+  rules: {
```
koonopek, line 16: This rules are duplicated from top level config?

packages/sdk/.eslintrc.js
koonopek: I think we have also rules set here no only parserOptions, why is so?

packages/oracle-node/src/aggregated-price-handlers/DataPackageBroadcastPerformer.ts
```diff
@@ -146,7 +147,7 @@ export class DataPackageBroadcastPerformer
         this.manifestDataProvider.latestManifest!,
         price.symbol
       ),
-      metadata: createMetadataForRedstonePrice(price),
+      metadata: createMetadataForRedstonePrice(price) as unknown as Metadata,
```
koonopek, line 150: Why? This function is typed why are you casting it?

packages/sdk/.eslintrc.js
rpiszczatowski: eh, this must have happened during rebasing. I'll clean it up...

packages/sdk/.eslintrc.js
rpiszczatowski: As above

packages/oracle-node/src/aggregated-price-handlers/DataPackageBroadcastPerformer.ts
rpiszczatowski: Well spotted! The pattern `as unknown as X` means someone tries to cast between two unrelated types. We have incompatible types here. I simply don't know how to fix this as I don't fully understand our `Metadata` vs `SourceMetadata` vs `Metadata with value property`. Before my change one of the types was `any` and the compiler didn't warn us. Apparently, this code works as is so I just added the cast but at some point, we should probably unify our approach to the metadata concept and then we can get rid of this cast.

koonopek: Yeah so it is unified:
* Metadata is very general type Record<string,any> it literally can be anything
* RedstoneTypes.MetadataForRedstonePrice is concrete "implementation" of Metadata this types are compatible and any casting should not be required

packages/protocol/src/data-point/DataPoint.ts
```diff
@@ -11,7 +10,7 @@ export interface IStandardDataPoint {
 }
 export type DataPointPlainObj = IStandardDataPoint | INumericDataPoint;
 
-export type Metadata = Record<string, any>;
+export type Metadata = Record<string, unknown>;
```
koonopek, line 13: Root of error

packages/oracle-node/src/aggregated-price-handlers/DataPackageBroadcastPerformer.ts
rpiszczatowski: I agree that it is unified.

But typescript simply works differently. You cannot simply assign an interface with two properties to the Record type. Leaving `any` in the type signature for `Metadata` simply disables type checking which is not we want. See this article: https://bobbyhadz.com/blog/typescript-index-signature-for-type-is-missing-in-type

There are a couple of solutions here (in order of my preference):
- leave type assertion I added
- change interface to type as for type such assignment is acceptable
- use spread operator on assignment {...create()} (not sure why exactly but it works)
- leave `any` in type signature and disable linter for that line


Why (philosophically) such an assignment is not accepted by TS:
`Metadata` is a type on which you can access any property you want. `MetadataForRedstonePrice` is an interface with two properties. Accessing anything else on it is something that should be rejected by the type system as not type safe operation.

koonopek (approved):

PR #1035, opened 2023-09-05
chore: fix running unit tests from vscode

rpiszczatowski:
axios 1.x.x requires newer jest


packages/oracle-node/package.json
```diff
@@ -40,7 +40,7 @@
     "@bundlr-network/client": "^0.7.4",
     "@metamask/eth-sig-util": "^4.0.1",
     "arweave": "^1.10.23",
-    "axios": "^0.21.1",
+    "axios": "^1.5.0",
```
rpiszczatowski, line 43: rpc-providers uses now axios 1.4.x and for the reason that is not that clear to me having here older axios caused jest (in vscode) fail to start tests. This was connected to the fact that axios 1.x by default uses EcmaScript modules system instead of commonJS.

cehali (approved):

PR #1036, opened 2023-09-05
chore: fix maverick abi and make multicall more roboust

rpiszczatowski:
- allow throwing to the multicall itself instead of us verifying if all the calls were successful by setting tryAggregate to false.

- ABIs in the form of strings instead of objects cause multicall to fail to decode responses. Having this changed we can drop the code branch where we handle not decoded responses as the only way for a response not to get decoded is when one of the calls failure which we disallow.

- Once responses are being parsed by multicall we have to fix parsing responses from multicall in our fetchers as they are not BigNumber objects but plain JS objects with two properties set: type and hex. This kind of objects are properly handled by BigNumber.from static method.


koonopek:
tests are failing

packages/oracle-node/src/fetchers/curve/CurveFetcher.ts
```diff
@@ -202,13 +202,13 @@ class MultiCallHandler {
     ] = multicallResponse;
 
     const commonParsedResult = {
-      sellRatio: new Decimal(ratioBigNumber.toString()).div(
+      sellRatio: new Decimal(BigNumber.from(ratioBigNumber).toHexString()).div(
```
Lukasz2891, line 205: What do you think about introducing a function like `convert = x => Decimal(BigNumber.from(x).toHexString())` :)?

rpiszczatowski: We could introduce such a method into our utils and use it in more places but I'd prefer not to change much right now as I have eslint PR pending and I'd prefer to avoid conflicts with that PR

koonopek (approved):

PR #1048, opened 2023-09-07
chore: create simple relayer function for non-evm chains

Lukasz2891:


packages/starknet-connector/yarn.lock
rpiszczatowski, line 1: You've just added starknet-connector as a workspace. Shouldn't starknet-connector share global yarn.lock?

Lukasz2891: Hmm, I've added it before - now just modified. So what should I do?

rpiszczatowski (dismissed):
I looked at sdk changes and building related stuff.

packages/eth-contracts/yarn.lock
```diff
@@ -0,0 +1,10767 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
cehali, line 1: I don't think `yarn.lock` is required here

cehali (dismissed):
1 minor comment

packages/eth-contracts/yarn.lock
Lukasz2891: needed to do some small fixes, so it's removed now ;)

Lukasz2891:
Re-check it pls after rebase&lint fixes :)

packages/starknet-relayer/src/starknet/cairo0/PriceFeedContractAdapter.ts
```diff
@@ -0,0 +1,14 @@
+import { Contract } from "starknet";
+import { PriceFeedRoundData } from "redstone-sdk";
+
+export class PriceFeedContractAdapter {
+  constructor(private readonly contract: Contract) {}
+
+  async readLatestRoundData(): Promise<PriceFeedRoundData> {
+    return (
+      (await this.contract.call("latest_round_data")) as {
+        [key: string]: any;
```
rpiszczatowski, line 10: Let's avoid using `any` type as this will become a rule in the very near future.

rpiszczatowski (approved):

packages/starknet-relayer/src/starknet/cairo0/PriceFeedContractAdapter.ts
Lukasz2891: I'll try to do it next time, thanks :)

PR #1050, opened 2023-09-07
feat: wombat fetcher

rpiszczatowski:


packages/utils/src/common/conversions.ts
```diff
@@ -0,0 +1,5 @@
+import Decimal from "decimal.js";
+import { BigNumber, BigNumberish } from "ethers";
+
+export const bignumberishToDecimal = (value: BigNumberish) =>
```
rpiszczatowski, line 4: I'd appreciate a shorter name for this method.

packages/utils/src/common/multicall.ts
```diff
@@ -55,3 +62,76 @@ export async function multiCallOneContract(
...
+
+  getResult<T = unknown>(
+    contractLabel: string,
+    callLabel: string,
+    returnValueIndex: number
+  ): T {
+    return this.getResults<T>(contractLabel, callLabel)[returnValueIndex];
+  }
+
+  getResults<T = unknown>(contractLabel: string, callLabel: string): T[] {
```
hatskier, line 79: Is it always true that all the returned values have the same type?

packages/oracle-node/src/fetchers/wombat/WombatFetcher.ts
```diff
@@ -0,0 +1,233 @@
...
+      DEFAULT_AMOUNT_IN_USD_FOR_SLIPPAGE
+    );
+
+    const amountInQuoteToken = convertUsdToTokenAmount(
+      pairedToken ?? quoteToken.symbol,
+      quoteTokenScaler.tokenDecimalsScaler.toNumber(),
+      DEFAULT_AMOUNT_IN_USD_FOR_SLIPPAGE
+    );
+
+    const multicallHandler = new MultiCallHandler(
```
hatskier, line 70: [Up to you] WDYT about renaming MultiCallHandler to WombatMultiCallHandler?

hatskier (approved):

packages/utils/src/common/multicall.ts
rpiszczatowski: No, it is not necesarily true. But it is ofter the case that all of them are BigNumberish. If this is not the case there are two options:
- do not use this method, use `getResult` to get a single value
- use it with `T=unknown` and put type assertions on the caller side on a per-element basis

packages/oracle-node/src/fetchers/wombat/WombatFetcher.ts
rpiszczatowski: In both Maverick and Curve where this pattern is already used this 'internal' class is called simply `MultiCallHandler`. I personally have no preference.

PR #1066, opened 2023-09-12
chore: github actions - use 2-core instead of 8-core

wrybakiewicz:
I tested that our workflows running on 8-core cpu doesn't influence workflow time. 

The only workflow I found takes more time is "Oracle Node CI" - from 7-9 minutes, takes 10-11 minutes (but still changed it to 2-core, is it ok to you ?).

I also added timeouts (it needs to be per workflow) with some reserve.

I changed 8-core (which we are 100% paiying for)to 2-core (which we have included in enterprise plan). We were paying additional 1k$ for it.


cehali (commented):
Looks good for me

cehali (approved):

PR #1073, opened 2023-09-13
publish: redstone-sdk 1.8.2

koonopek:
We were using constant ___ALL_FEEDS___ which is not avaiable in redstone-protocol published package.
I wanted to avoid publishing new redstone-protocol just to add simple constant, so I have hardcoded it.


Lukasz2891 (approved):
Nice, I've been wanting (what a form;p) to do it yesterday for 1.8.1 :D

PR #1083, opened 2023-09-13
chore: update configs and code for other connectors

Lukasz2891:


.gitignore
```diff
@@ -114,3 +114,5 @@ typechain-types
 worktree-*
 oracle-node-obfuscated
 packages/on-chain-relayer/price-feed-contract-address.txt
+
+packages/**/yarn*
```
rpiszczatowski, line 118: What is this?

packages/fuel-connector/src/prices/FuelPricesContractAdapter.ts
```diff
@@ -9,14 +9,19 @@ export class FuelPricesContractAdapter implements IPricesContractAdapter {
     private gasLimit: number
   ) {}
 
+  protected static extractNumbers(result: InvocationResult): number[] {
+    // eslint-disable-next-line
+    return result.value.map(num);
+  }
```
rpiszczatowski, line 15: You can tell `InvocationResult` what you expect from it as a value property.
```suggestion
  protected static extractNumbers(
    result: InvocationResult<U256Output[]>
  ): number[] {
    return result.value.map(num);
  }
```

packages/fuel-connector/test/prices/prices.spec.ts
```diff
@@ -98,8 +98,8 @@ describe("Prices contract", () => {
 
     try {
       await adapter.getPricesFromPayload(paramsProvider);
-    } catch (e: any) {
-      expect(e).not.toBeNull();
+    } catch (e: unknown) {
```
rpiszczatowski, line 101: `: unknown` is redundant. `} catch (e) {` is the same.

packages/fuel-connector/test/prices/prices.spec.ts
```diff
@@ -98,8 +98,8 @@ describe("Prices contract", () => {
...
+      expect(e as Error).not.toBeNull();
```
rpiszczatowski, line 102: Is this type assertion needed?

packages/starknet-connector/src/prices/StarknetPricesContractAdapter.ts
```diff
@@ -1,20 +1,13 @@
...
-    paramsProvider: ContractParamsProvider
-  ): Promise<number[]> {
-    return this.extractNumbers(
-      await this.contract.call("get_prices", [
-        paramsProvider.getHexlifiedFeedIds(),
-        await paramsProvider.getPayloadData(),
-      ])
-    );
+  protected static extractNumbers(response: Result): number[] {
+    return (response as Result[]).map(getNumberFromStarknetResult);
```
rpiszczatowski, line 10: This is cool. Recursive type definition. `Result` can be an array of `Result`s.

packages/starknet-connector/tsconfig.build.json
```diff
@@ -0,0 +1,17 @@
...
+  ],
+  "references": [
+    {
+      "path": "../protocol/tsconfig.build.json"
+    },
+    {
+      "path": "../sdk/tsconfig.build.json"
+    }
+  ]
+}
```
rpiszczatowski, line 17: new line at the end of file

packages/starknet-relayer/src/start-relayer.ts
```diff
@@ -2,7 +2,7 @@ import { config } from "./config";
 import { ContractConnectorFactory } from "./starknet/ContractConnectorFactory";
 import { startSimpleRelayer } from "redstone-sdk";
 
-(async () => {
+void (async () => {
```
rpiszczatowski, line 5: I think we agreed to use eslint annotation as a cleaner way of disabling those warnings.

Even though my personal preference would be probably to use `void` as shorter and terser, we should stick to comments.

rpiszczatowski (changes requested):
Minor suggestions.

.gitignore
Lukasz2891: ignoring all yarn.locks and yarn.lock-error inside the particular package ("remove that file for `yarn.lock`")

packages/starknet-relayer/src/start-relayer.ts
Lukasz2891: yes, it was changed before the discussion and just remained .

packages/fuel-connector/tsconfig.json
```diff
@@ -1,28 +1,39 @@
...
+      "path": "../protocol/tsconfig.build.json"
     },
     {
-      "path": "../sdk"
-    },
+      "path": "../sdk/tsconfig.build.json"
+    }
   ]
-}
+}
```
cehali, line 39: Please add new line

packages/fuel-connector/tsconfig.build.json
```diff
@@ -0,0 +1,17 @@
...
+  ],
+  "references": [
+    {
+      "path": "../protocol/tsconfig.build.json"
+    },
+    {
+      "path": "../sdk/tsconfig.build.json"
+    }
+  ]
+}
```
cehali, line 17: Please add new line

cehali (approved):

.gitignore
rpiszczatowski: sorry for how I formulated a question. It should be: why do we ignore patterns of files other than the files created by tools we use in a normal development cycle? There shouldn't be any yarn* files created inside packages. If you have such files then:
1. you can delete them if not needed otherwise
2. you can add them to .git/info/exclude if those are your private stuff not relevant to anybody except you.

I consider adding too much stuff to .gitignore a harmful practice in general. (said somebody whose .gitignore is 116 lines long ;) ).

rpiszczatowski (approved):

.gitignore
Lukasz2891: But they are created in a development cycle, but maybe my own is abnormal :p I'll delete the line or will use 'exclude' :)

packages/fuel-connector/tsconfig.json
Lukasz2891: I'll do it soon :)

packages/fuel-connector/tsconfig.build.json
Lukasz2891: Will be added soon ;)

PR #1085, opened 2023-09-13
eslint for oracle-node, on-chain-relayer and cache-service

rpiszczatowski:


.github/workflows/gelato-relayer-tests.yml
```diff
@@ -24,6 +24,8 @@ jobs:
         run: yarn
       - name: Build
         run: yarn build
+      - name: Lint
```
Lukasz2891, line 27: Shouldn't we lint before build?

rpiszczatowski: no, because build delivers typings for referenced projects. Eslint searches for types for e.g. utils in utils' dist directory.

packages/cache-service/scripts/analyze-data-packages.ts
```diff
@@ -84,6 +85,7 @@ async function main() {
       lastTimestamp = timestamp;
 
       const timestampFromId = new Date(
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
```
koonopek, line 88: Omg xD Do we have to list all the possible errors? Maybe we can configure it

packages/cache-service/scripts/compare-dbs-alex.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unnecessary-condition */
```
koonopek, line 1: Why?

packages/cache-service/scripts/wait-for-data-packages.ts
```diff
@@ -19,6 +20,7 @@ async function main() {
   console.log("MongoDB connected");
 
   await checkDataPackagesCount();
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
koonopek, line 23: Why it is missused promise? Is eslint stupid?

packages/cache-service/scripts/analyze-data-packages.ts
rpiszczatowski: This doesn't look good, I admit. But in reality, you click in vscode on the error mark and click `ignore for this line`.
I don't know of a method that allows to disable multiple warnings (but not all of them) with a single annotation

packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -78,74 +84,76 @@ export class DataPackagesService {
...
+        {
+          $sort: { timestampMilliseconds: -1 },
+        },
+      ]);
 
     // Parse DB response
     for (const dataPackage of groupedDataPackages) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
       const { _id, __v, ...rest } = dataPackage;
-      __v;
```
koonopek, line 139: I think that this line was here on purpose, but Im not sure

packages/cache-service/src/utils/IsNumberOrString.ts
```diff
@@ -6,10 +6,12 @@ import {
...
 @ValidatorConstraint({ name: "string-or-number", async: false })
 export class IsNumberOrString implements ValidatorConstraintInterface {
-  validate(text: any, args: ValidationArguments) {
+  // eslint-disable-next-line @typescript-eslint/class-methods-use-this, @typescript-eslint/no-unused-vars
+  validate(text: unknown, args: ValidationArguments) {
     return typeof text === "number" || typeof text === "string";
   }
 
+  // eslint-disable-next-line @typescript-eslint/class-methods-use-this, @typescript-eslint/no-unused-vars
   defaultMessage(args: ValidationArguments) {
```
koonopek, line 15: You can use `_` before unused var to silent warning I think (just tip)

packages/cache-service/src/utils/utils.ts
```diff
@@ -1,14 +1,22 @@
 import { Logger } from "@nestjs/common";
 
 // TODO: this should be in common package
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
koonopek, line 4: xDDDD oh boy maybe we should disable `no-unsafe-member-access`?

packages/cache-service/scripts/compare-dbs-alex.ts
rpiszczatowski: Because this script is based on the idea that on top of it, you have configuration parameters, e.g. 
```
const ANALYZE_ONLY_BIG_PACKAGES = true;
```
and then we have if statements in the code based on that value that typescript considers always truthy.

But your challenging this approach made me think that we can do better :)
Changing the above to
```
const ANALYZE_ONLY_BIG_PACKAGES: boolean = true;
```
changes the type of constant from `true` to `boolean` and eslint stops complaining.

I'll change the code.

packages/on-chain-relayer/src/config.ts
```diff
@@ -29,7 +29,7 @@ export const config = () => {
     relayerConfig.updateConditions.includes("value-deviation")
   ) {
     throw new Error(
-      "Mento adapter does not support the value-deviation update condition"
+      "Mento adapter does not support the value-deviation update condition",
```
koonopek, line 32: If you disabled this `trailing comma` rule there would be a lot of less files to check

koonopek: Also I dont see any advantages of enabling it, it is very opinionated

packages/cache-service/src/data-packages/data-packages.service.ts
rpiszczatowski: What the purpose could be? Maybe it was a way to 'use' the variable to silent the compiler about warning of unused variables. BTW. I disabled warnings about unused variables starting with underscore.

packages/on-chain-relayer/src/core/monitoring/send-healthcheck-ping.ts
```diff
@@ -1,19 +1,20 @@
...
 import { config } from "../../config";
 
 export const sendHealthcheckPing = async () => {
   const { healthcheckPingUrl } = config();
   if (healthcheckPingUrl) {
     try {
       await axios.get(healthcheckPingUrl);
-    } catch (error: any) {
+    } catch (e: unknown) {
+      const error = e as AxiosError;
```
koonopek, line 10: Hacked!

packages/cache-service/scripts/wait-for-data-packages.ts
rpiszczatowski: The idea is that setInterval expects a function that returns `void`; thus, it 'promises' that if you pass the function that returns `Promise` setInterval won't wait for it.

This is an artifact of setInterfal not being fully compatible with async.

packages/cache-service/src/utils/IsNumberOrString.ts
rpiszczatowski: Definitely! Because there were two warnings at the same line, I probably misclicked in vscode.

packages/on-chain-relayer/src/config.ts
rpiszczatowski: But this is a prettier complaining (via eslint plugin). If I changed the prettier rule it would generate multiple changes in different direction.
This is a one-time problem.

packages/cache-service/src/utils/utils.ts
rpiszczatowski: What makes sense in case of this file is to it this way:
```
/* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-return */
// eslint-disable-next-line @typescript-eslint/no-explicit-any
export function stringifyError(e: any) {
  if (e === undefined) {
    return "undefined";
  } else if (e === null) {
    return "null";
  } else if (e.response) {
    return JSON.stringify(e.response.data) + " | " + e.stack;
  } else if (e.toJSON) {
    return JSON.stringify(e.toJSON());
  } else {
    return e.stack || String(e);
  }
}
/* eslint-enable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-return */
```
Now code is more readable. I'll apply such a change.

packages/on-chain-relayer/src/config.ts
rpiszczatowski: Wow, I just realized that for on-chain-relayer our prettier config was not applied because of lack of entry in package.json. This is the reason why the comma was added in this particular place. I'll make sure all the packages uses our prettier config.

.github/workflows/oracle-tests.yml
```diff
@@ -19,6 +25,10 @@ jobs:
         run: yarn
       - name: Build
         run: yarn build
+      - name: Lint
+        run: yarn lint
+      - name: Lint streamr-proxy
+        run: cd ../streamr-proxy && yarn lint
```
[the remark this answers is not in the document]
rpiszczatowski: Because streams-proxy doesn't have a dedicated CI workflow and oracle-node makes use of it and its workflow runs on changes in streamr-proxy. Now at least we check streamr-proxy anywhere. Of course, I can create a separate CI flow for streamr-proxy. I just thought we already had many workflows. This approach does not spin a new machine on GitHub.

koonopek (approved):

Lukasz2891 (approved):

packages/oracle-node/src/fetchers/band/BandFetcher.ts
```diff
@@ -13,7 +13,7 @@ interface BandResponse {
         px: number;
         multiplier: number;
         [otherOptions: string]: unknown;
-      }
+      },
```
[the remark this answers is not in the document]
cehali: I believe in an array trailing comma is require in ES5

cehali (dismissed):

packages/oracle-node/src/fetchers/dex-on-chain/DexOnChainFetcher.ts
```diff
@@ -139,8 +142,8 @@ export const getDefinedResponse = <T>(
 ): T => {
   RedstoneCommon.assert(
     !!responses[dataFeedId],
-    `responses are defined for ${Object.keys(
-      responses
+    `responses are defined for ${Object.keys(responses).join(
+      ","
```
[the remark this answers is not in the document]
rpiszczatowski: I don't know what to say. This is how our formatting works.

packages/on-chain-relayer/src/core/update-conditions/check-value-deviation-condition.ts
```diff
@@ -73,12 +73,13 @@ const calculateDeviation = (
 class ValueDeviationLogTrace {
   private perDataFeedId: Record<
     string,
-    {
-      valueFromContract: number;
-      valuesFromNode: number[];
-      timestamp: number;
-      packagesCount: number;
-    }
+    | {
```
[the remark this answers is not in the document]
rpiszczatowski: This is called union type and this is the way prettier does it. I was supprised at first, but this is 'similar but opposite' to trailing commas.

PR #1086, opened 2023-09-14
fix: adjust deviation check configs in manifests

cehali:
It turned out during my fight with dry-runs that deviation config for a specific token is not taken under consideration, in check it always returns `undefined` for custom deviation config so it uses default one


packages/oracle-node/src/manifest/validate-manifest.ts
```diff
@@ -33,7 +33,7 @@ const CustomUrlsSchema = z.object({
 
 const TokenSchema = z.object({
   source: z.array(z.string()).optional(),
-  deviationCheck: z.string().optional(),
+  deviationCheck: DeviationCheckSchema.optional(),
```
koonopek, line 36: Why this change?
Also why we don't change in node code

koonopek (approved):

rpiszczatowski (approved):

PR #1088, opened 2023-09-14
feat: implementton relayer

Lukasz2891:


packages/ton-relayer/src/start-relayer.ts
```diff
@@ -0,0 +1,11 @@
...
+import { ContractConnectorFactory } from "./ton/ContractConnectorFactory";
+import { startSimpleRelayer } from "redstone-sdk";
+
+// eslint-disable-next-line @typescript-eslint/no-floating-promises -- Disabled for top-level functions
+(async () => {
+  await startSimpleRelayer(
+    config,
+    ContractConnectorFactory.makePriceManagerContractConnector()
+  );
+})();
```
koonopek, line 11: I think you can just: 
startSimpleRelayer(
    config,
    ContractConnectorFactory.makePriceManagerContractConnector()
  );

packages/ton-relayer/src/ton/PriceManagerRelayerTonContractConnector.ts
```diff
@@ -0,0 +1,43 @@
...
+  DATA_FEEDS,
+  DATA_SERVICE_ID,
+  UNIQUE_SIGNER_COUNT,
+} from "../config/data-service-parameters";
+import {
+  TonNetwork,
+  TonPricesContractConnector,
+} from "@redstone-finance/ton-connector";
+
+export class PriceManagerRelayerTonContractConnector
```
koonopek, line 19: PriceManager <3

packages/ton-relayer/src/ton/PriceManagerRelayerTonContractAdapter.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    return {
+      payload_timestamp: timestamp,
+    };
+  }
+
+  async writePrices(_: number): Promise<string> {
+    await this.writePricesFromPayloadToContract(this.paramsProvider);
+
+    return "";
```
koonopek, line 34: What is the sense of this response?

packages/ton-relayer/src/ton/PriceManagerRelayerTonContractAdapter.ts
```diff
@@ -0,0 +1,36 @@
...
+  IPriceManagerContractAdapter,
+  PriceManagerMetadata,
+} from "redstone-sdk";
+import {
+  AnyTonOpenedContract,
+  TonPriceManager,
+  TonPricesContractAdapter,
+} from "@redstone-finance/ton-connector";
+
+export class PriceManagerRelayerTonContractAdapter
```
koonopek, line 12: I don't get why we need this adapter instead of just calling: 
* await this.writePricesFromPayloadToContract(this.paramsProvider);
* const timestamp = await this.readTimestampFromContract();

packages/ton-relayer/src/ton/PriceFeedRelayerTonContractConnector.ts
```diff
@@ -0,0 +1,13 @@
+import { IContractConnector } from "redstone-sdk";
+
+import { TonPriceFeedContractConnector } from "@redstone-finance/ton-connector";
+import { PriceFeedRelayerTonContractAdapter } from "./PriceFeedRelayerTonContractAdapter";
+
+export class PriceFeedRelayerTonContractConnector
```
koonopek, line 6: What is the point of this class?

packages/ton-relayer/src/ton/PriceFeedRelayerTonContractAdapter.ts
```diff
@@ -0,0 +1,23 @@
+import { IPriceFeedContractAdapter, PriceFeedRoundData } from "redstone-sdk";
+import {
+  AnyTonOpenedContract,
+  TonPriceFeed,
+  TonPriceFeedContractAdapter,
+} from "@redstone-finance/ton-connector";
+
+export class PriceFeedRelayerTonContractAdapter
```
koonopek, line 8: What is the point of this class?

packages/ton-connector/src/network/CustomTonNetwork.ts
```diff
@@ -41,13 +42,14 @@ export class CustomTonNetwork implements TonNetwork {
 
     this.walletContract = this.oldApi.open(wallet);
     this.sender = this.walletContract.sender(walletKey.secretKey);
+    this.walletAddress = wallet.address;
   }
 
   async isContractDeployed(address?: Address): Promise<boolean> {
```
koonopek, line 48: Why do you return true when address is undefined?
Why address can be undefined? It seems weired API

packages/ton-relayer/src/scripts/get-price.ts
```diff
@@ -0,0 +1,24 @@
...
+
+(async () => {
+  for (const feedAddress of [priceFeedAddresses.BTC, priceFeedAddresses.ETH]) {
+    try {
+      const connector =
+        ContractConnectorFactory.makePriceFeedContractConnector(feedAddress);
+      const adapter = await connector.getAdapter();
+
+      await adapter.fetchRoundData();
+      await connector.waitForTransaction("");
```
koonopek, line 13: ? what is happening here

packages/ton-connector/src/TonContractConnector.ts
```diff
@@ -38,8 +35,57 @@ export abstract class TonContractConnector<
     return (await this.network.api!.getLastBlock()).last.seqno;
   }
 
-  // eslint-disable-next-line @typescript-eslint/class-methods-use-this -- Interface conformance
-  waitForTransaction(_: string): Promise<boolean> {
-    throw "Not implemented";
+  async waitForTransaction(_: string): Promise<boolean> {
+    console.log(`Transaction sent.`);
```
koonopek, line 39: So we are waiting? Or sending? :)

packages/ton-connector/src/TonContractConnector.ts
```diff
@@ -38,8 +35,57 @@ export abstract class TonContractConnector<
...
+    const { hash } = await this.fetchNetworkValues();
```
koonopek, line 40: This is race condition you are optimisicly asssuming that last transaction sent by this account will be transaction sent by this account.
`openNetwork.call(x)` (or something like this)  doesnt return tx hash?

packages/ton-connector/src/TonContractConnector.ts
```diff
@@ -38,8 +35,57 @@ export abstract class TonContractConnector<
...
+    const seqno = (await api.getLastBlock()).last.seqno;
+    const account = await api.getAccountLite(seqno, walletAddress);
+    const hash = account.account.last!.hash;
+    const transactions = await api.getAccountTransactions(
+      walletAddress,
+      BigInt(account.account.last!.lt),
+      Buffer.from(hash, "base64")
+    );
+
+    const lastTransaction = transactions[0].tx;
```
koonopek, line 80: What if you have send 2 transactions before awaiting?

packages/ton-connector/src/Ton.ts
```diff
@@ -32,34 +33,10 @@ export abstract class Ton {
     body?: Cell,
     sendMode = SendMode.PAY_GAS_SEPARATELY
   ): Promise<void> {
-    await this.wait(async () => {
-      await provider.internal(this.sender, {
-        value: `${coins.toFixed(2)}`,
-        body,
-        sendMode,
-      });
+    await provider.internal(this.sender, {
```
koonopek, line 36: Okey I see now that they don't return hash, however you can propably calculate it by yourself, before even sending transaction.

packages/ton-connector/src/network/CustomTonNetwork.ts
koonopek: Maybe you meant something like is WalletContractDeployed then it would have more sens

packages/ton-connector/tests/unit_tester/TonUnitTesterContractAdapter.ts
```diff
@@ -1,14 +1,9 @@
-import { OpenedContract } from "ton";
-import { SandboxContract } from "@ton-community/sandbox";
 import { TonUnitTester } from "./TonUnitTester";
 import { Cell } from "ton-core";
+import { AnyTonOpenedContract } from "../../src";
 
 export class TonUnitTesterContractAdapter {
```
koonopek, line 5: What it the advantage of this class over simply calling contract?

koonopek (commented):
Left some comments, however I don't know how TON works. So some of them might be inadequate

packages/ton-connector/src/TonContractConnector.ts
Lukasz2891: Nothing returns the tx hash :( that's just ton...

packages/ton-connector/src/TonContractConnector.ts
Lukasz2891: Race :( I didn't find a better solution for that, the SDK doesn't provide methods for that...

packages/ton-connector/src/network/CustomTonNetwork.ts
```diff
@@ -41,13 +42,14 @@ export class CustomTonNetwork implements TonNetwork {
...
-    const seqno = await this.walletContract!.getSeqno();
-
-    return await (address == undefined ||
-      this.api!.isContractDeployed(seqno, address));
+    return (
+      address == undefined ||
+      ((await this.oldApi?.isContractDeployed(address)) ?? true)
```
rpiszczatowski, line 51: I am not sure about using patterns of this class, but it looks to me that:
- both `oldApi` and `api` are initialized at the same point in `setUp` method
- when using one of them we use ! and for other ?.

If setUp is always called before anything else in this class and you know that property will be always set on usage even though it is not initialized in constructor you can use ! on declaration like this.
```
export class CustomTonNetwork implements TonNetwork {
  api!: TonClient4;
  oldApi!: TonClient;
```

packages/ton-relayer/src/ton/PriceManagerRelayerTonContractAdapter.ts
Lukasz2891: The interface requires string (transactionId) or Prices.

packages/ton-relayer/src/start-relayer.ts
Lukasz2891: that was probably related to lint method ;)

packages/ton-relayer/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "dotenv": "^16.0.3",
+    "ethers": "^5.7.2",
+    "redstone-sdk": "^1.8.3",
+    "@redstone-finance/ton-connector": "^0.1.0"
+  },
+  "devDependencies": {
+    "@types/node": "^18.13.0",
+    "typescript": "^4.9.5"
+  },
+  "references": [
```
rpiszczatowski, line 24: Doesn't this belong to tsconfig.json?
Also for references we use tsconfig.build.json instead of directory name.

packages/ton-relayer/src/ton/PriceFeedRelayerTonContractConnector.ts
Lukasz2891: General idea, common with other connectors.
The layer is responsible for creating/opening the contract and returning its adapter.
It hides the whole concrete blockchain implementation and can be reused in an abstract way

packages/ton-relayer/yarn.lock
rpiszczatowski, line 1: We should remove this file.

packages/utils/src/monitoring/send-healthcheck-ping.ts
```diff
@@ -1,19 +1,19 @@
...
 
-export const sendHealthcheckPing = async () => {
-  const { healthcheckPingUrl } = config();
+export const sendHealthcheckPing = async (healthcheckPingUrl?: string) => {
   if (healthcheckPingUrl) {
     try {
       await axios.get(healthcheckPingUrl);
-    } catch (error: any) {
+    } catch (err: unknown) {
+      const error = err as Error & { code?: string };
```
rpiszczatowski, line 8:
```suggestion
    } catch (err) {
      const error = err as AxiosError;
```

packages/ton-relayer/src/ton/PriceManagerRelayerTonContractAdapter.ts
Lukasz2891: what is `this` ;)?
maybe it's not clear in that case but that's a class implementing `IPriceManagerContractAdapter` and it's adapted for a common usage of all relayers, in the same way for other blockchains (fuel/starknet0/starknet1/...).

packages/ton-connector/src/TonContractConnector.ts
Lukasz2891: waiting ;) Maybe the line is not needed.

Lukasz2891:
> Left some comments, however I don't know how TON works. So some of them might be inadequate

They are ok, thanks ;) maybe some solutions are unclear, indeed. But it's also worth to see across all non-evm connectors ;)

packages/ton-relayer/package.json
Lukasz2891: should be removed

packages/ton-connector/src/network/CustomTonNetwork.ts
Lukasz2891: good catch.

rpiszczatowski (dismissed):

koonopek (approved):

hatskier (approved):

packages/ton-relayer/src/scripts/get-price.ts
Lukasz2891: The feed needs to communicate to its adapter and fetch the prices. There is no way to do it synchronously or to wait for it inside the contract.

packages/ton-connector/src/network/CustomTonNetwork.ts
Lukasz2891: `isContractDeployed` is a common name for TonClient/TonClient4 apis (Ton. Wallet contract is a regular contract.

the undefined seems to be a leftover which doesn't occur afters setting the walletAddress above.

rpiszczatowski (approved):

PR #1089, opened 2023-09-14
Robal/syncpack

rpiszczatowski:


packages/evm-connector/test/libs/numeric-array-lib.test.ts
```diff
@@ -12,13 +12,13 @@ describe("SampleNumericArrayLib", function () {
 
   const prepareRandomArray = (arrayLength: number) => {
     return getRange({ start: 0, length: arrayLength }).map(() =>
-      Math.round(Math.random() * 10000)
+      Math.round(Math.random() * 10000),
```
rpiszczatowski, line 15: We have no prettier config in evm-connector yet. I'll be adding it in future PRs. Then those comma will be removed.

cehali (dismissed):
LGTM

PR #1093, opened 2023-09-14
feat: Oracle node telemetry

wrybakiewicz:
UPDATE:
- fixed problem with lack of .git in our dockerfile by removing git last commit usage replaced by env variable

- add node performance data & docker image version telemetry (sent to Influxdb and to be displayed in grafana)
- add telemetry in CI (to special influxdb redstone-test bucket, we empty it every hour so there is not much trash data from CI) - so we can test if telemetry doesn't brake anything in node
- DOCKER_IMAGE_TAG is set during image build
- added batch sending (after node queries series finish, every 10s)
- 
After merging I will add env variables TELEMETRY_URL  and TELEMETRY_AUTHORIZATION_TOKEN  to our terraform for all nodes.


packages/oracle-node/src/utils/performance-tracker.ts
```diff
@@ -46,25 +47,74 @@ export function trackEnd(trackingId: string): void {
...
+  fields: string
+) {
+  const requestData = `${measurementName},${tags} ${fields} ${Date.now()}`;
+
+  const requestConfig = {
+    headers: {
+      Authorization: `Token ${config.telemetryAuthorizationToken}`,
+    },
+  };
+  axios
```
hatskier, line 104: Can you add tests to be sure that node works even if the `post` promise fails. We had problems with this in past, because this is default Node.js behaviour

.github/workflows/long-arbitrum-dry-run-test.yml
```diff
@@ -30,3 +30,6 @@ jobs:
           COINGECKO_API_URL: https://pro-api.coingecko.com/api/v3/simple/price
           COINGECKO_API_KEY: ${{ secrets.COINGECKO_API_KEY }}
           COINMARKETCAP_API_KEY: ${{ secrets.COINMARKETCAP_API_KEY }}
+          DOCKER_IMAGE_TAG: "ci"
```
hatskier, line 33: What do you think about getting this value without env variable, but using some mechanism like this: https://github.com/redstone-finance/redstone-oracles-monorepo/blob/50df3a08ebf3131d3c395160268f68995198e5f5/packages/oracle-node/src/NodeRunner.ts#L156

It would help us to have a single source of truth

hatskier (commented):
Good job! Added 2 ideas for improvements

.github/workflows/long-arbitrum-dry-run-test.yml
koonopek: And in CI we could simply override this

packages/oracle-node/src/utils/performance-tracker.ts
```diff
@@ -46,25 +47,74 @@ export function trackEnd(trackingId: string): void {
...
+      console.error(
+        `Failed saving metric: ${measurementName}: ${e.response.status}, ${e.response.data.message}`
+      )
+    );
+}
+
+function isTelemetryEnabled() {
+  const isTelemetryEnabled =
+    config.telemetryUrl !== "" && config.telemetryAuthorizationToken;
+  if (!isTelemetryEnabled) {
```
koonopek, line 116: 1) we use `logger` for logging
2) I don't think that this log is fully. It can be printed once when node starts. Rest is redundant

packages/oracle-node/src/utils/performance-tracker.ts
```diff
@@ -46,25 +47,74 @@ export function trackEnd(trackingId: string): void {
...
+    .catch((e) =>
+      console.error(
+        `Failed saving metric: ${measurementName}: ${e.response.status}, ${e.response.data.message}`
+      )
+    );
+}
+
+function isTelemetryEnabled() {
+  const isTelemetryEnabled =
+    config.telemetryUrl !== "" && config.telemetryAuthorizationToken;
```
koonopek, line 115: Shouldn't be `||` instead of `&&`?

packages/oracle-node/src/utils/performance-tracker.ts
```diff
@@ -46,25 +47,74 @@ export function trackEnd(trackingId: string): void {
...
+async function saveMetric(label: string, executionTime: number): Promise<void> {
   const evmPrivateKey = config.privateKeys.ethereumPrivateKey;
   const evmAddress = new ethers.Wallet(evmPrivateKey).address;
   const labelWithPrefix = `${evmAddress.slice(0, 14)}-${label}`;
 
-  logger.info(`Metric: ${labelWithPrefix}. Value: ${value}`);
+  logger.info(`Metric: ${labelWithPrefix}. Value: ${executionTime}`);
+
+  if (isTelemetryEnabled()) {
+    const measurementName = "nodePerformance";
```
koonopek, line 85: Can be top level const

packages/oracle-node/src/utils/performance-tracker.ts
```diff
@@ -46,25 +47,74 @@ export function trackEnd(trackingId: string): void {
...
+    const tags = `label=${label},address=${evmAddress}`;
+    const fields = `executionTime=${executionTime}`;
+    sendMetric(measurementName, tags, fields);
+  }
+}
+
+async function sendMetric(
```
koonopek, line 92: Shouldnt we batch telemetry logs?

koonopek (commented):
GJ!

packages/oracle-node/src/utils/performance-tracker.ts
```diff
@@ -49,25 +52,66 @@ export function trackEnd(trackingId: string): void {
...
 
   // Clear the start value
   delete tasks[trackingId];
 
-  // Saving metric using Redstone HTTP endpoint
-  saveMetric(label, value);
+  sendNodePerformanceMetric(label, executionTime);
+}
+
+export async function sendNodeTelemetry() {
```
hatskier, line 64: IMHO it's not very clear from the naming that `sendNodeTelemetry` sends the actual request, but `sendNodePerformanceMetric` does not. Can we rename use the word queue instead of send in this functions and remove `await telemetrySendService.sendMetricsBatch();` from the `sendNodeTelemetry` and just call it where it's needed in the NodeRunner? WDYT?

hatskier (changes requested):

hatskier (approved):
LGTM!

packages/oracle-node/src/telemetry/TelemetrySendService.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const logger = loggerFactory("telemetry/TelemetrySendService");
+
+export default class TelemetrySendService {
+  private static instance: TelemetrySendService | undefined;
+  private metrics: string[] = [];
+
+  private constructor() {}
+
+  static getInstance(): TelemetrySendService {
```
koonopek, line 14: [For future] In JS modules are singleton so to create singleton it is enough to do export `export const telemetrySendService = new TelemetrySendService()` and just import it

packages/oracle-node/src/NodeRunner.ts
```diff
@@ -198,6 +208,8 @@ export default class NodeRunner {
     await this.safeProcessManifestTokens(iterationContext);
 
     printTrackingState();
+    await queueNodeTelemetry();
```
koonopek, line 211: [IMPORTANT] I think we shouldn't block here, we want node to still working properly even if telemetry server is down

packages/oracle-node/src/utils/performance-tracker.ts
```diff
@@ -49,25 +52,65 @@ export function trackEnd(trackingId: string): void {
...
 
   // Clear the start value
   delete tasks[trackingId];
 
-  // Saving metric using Redstone HTTP endpoint
-  saveMetric(label, value);
+  queueNodePerformanceMetric(label, executionTime);
+}
+
+export async function queueNodeTelemetry() {
```
koonopek, line 64: This shouldn't be async I think

packages/oracle-node/test/integration/NodeRunner.spec.ts
```diff
@@ -296,8 +296,8 @@ describe("NodeRunner", () => {
     it("should broadcast fetched and signed price data", async () => {
       await runTestNode();
 
-      // one for /bulk and the sconde one for prices
-      expect(axios.post).toHaveBeenCalledTimes(2);
+      // one for /bulk, one for prices and one for telemetry
+      expect(axios.post).toHaveBeenCalledTimes(3);
```
koonopek, line 300: This test is flaky, we should check paths, but we can leave it for now

koonopek (changes requested):

packages/oracle-node/src/NodeRunner.ts
koonopek: this should be pointing to one line below (212)

packages/oracle-node/src/utils/performance-tracker.ts
wrybakiewicz: there is fetching git commit that is async :/

koonopek: true, sorry didnt notice

packages/oracle-node/src/NodeRunner.ts
```diff
@@ -208,8 +207,9 @@ export default class NodeRunner {
     await this.safeProcessManifestTokens(iterationContext);
 
     printTrackingState();
-    await queueNodeTelemetry();
-    await this.telemetrySendService.sendMetricsBatch();
+    queueNodeTelemetry()
+      .then(() => telemetrySendService.sendMetricsBatch())
+      .catch(() => logger.info("Failed to send telemetry"));
```
koonopek, line 212: This log is redundant, cause we are already printing failure in `sendMetricsBatch`.
What do you think about:
```ts
await queueNodeTelemetry();
this.telemetrySendService.sendMetricsBatch(); // it shouldn't throw because it is in try catch
```

packages/oracle-node/src/utils/performance-tracker.ts
```diff
@@ -49,25 +51,65 @@ export function trackEnd(trackingId: string): void {
...
```
koonopek, line 63: Also can we wrap this in try-catch whole body, this function should never fail cause it is not crucial from node standpoint

packages/oracle-node/src/utils/performance-tracker.ts
```diff
@@ -49,25 +51,65 @@ export function trackEnd(trackingId: string): void {
...
+  );
 }
 
 export function printTrackingState() {
   const tasksCount = Object.keys(tasks).length;
   logger.info(`Perf tracker tasks: ${tasksCount}`, JSON.stringify(tasks));
 }
 
-function saveMetric(label: string, value: number) {
+function queueNodePerformanceMetric(label: string, executionTime: number) {
```
koonopek, line 88: Also can we wrap this in try-catch whole body, this function should never fail cause it is not crucial from node standpoint

koonopek (dismissed):

.github/workflows/deploy-images-to-ecr-public.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           echo TAG=$(echo ${GITHUB_SHA} | cut -c1-7) >> "$GITHUB_ENV"
       - name: Build docker image
         run: |
-          docker build -t "$IMAGE:$TAG" -f packages/oracle-node/Dockerfile .
+          docker build -t "$IMAGE:$TAG" -f packages/oracle-node/Dockerfile --build-arg DOCKER_IMAGE_TAG=$TAG .
```
koonopek, line 28: <3

koonopek (approved):
GJ 🥇

PR #1094, opened 2023-09-14
Setup lerna workflow

koonopek:
After relasing, we should change in all other repos to use this package


packages/cache-service/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@redstone-finance/cache-service",
-  "version": "0.0.1",
+  "version": "1.0.1",
```
hatskier, line 3: Why do you want to bump the major version? Can't we just increase the minor one to have e.g. 0.3.0 everywhere?

hatskier (commented):
Good job! Could you please also add few words about using lerna somewhere (e.g. in README or to our wiki)?

packages/cache-service/package.json
koonopek: we can, I don't why.

How is 0.3.0 better then 1.0.1?

Lukasz2891 (commented):
That's ok for me regarding to the code - more general questions on dscrd

package.json
```diff
@@ -1,40 +1,29 @@
...
-      "packages/oracle-node",
-      "packages/oracles-smartweave-contracts",
-      "packages/protocol",
-      "packages/rpc-providers",
-      "packages/sdk",
-      "packages/starknet-connector",
-      "packages/starknet-relayer",
-      "packages/streamr-proxy",
-      "packages/utils"
+      "packages/*"
```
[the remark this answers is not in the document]
koonopek: yep!

packages/cache-service/test/e2e/data-packages.spec.ts
```diff
@@ -8,11 +8,11 @@ import {
   SignedDataPackagePlainObj,
   consts,
   utils,
-} from "redstone-protocol";
+} from "@redstone-finance/protocol";
 import {
   RedstonePayloadParser,
   convertDataPointToNumericDataPoint,
-} from "redstone-protocol/dist/src/redstone-payload/RedstonePayloadParser";
+} from "@redstone-finance/protocol/dist/src/redstone-payload/RedstonePayloadParser";
```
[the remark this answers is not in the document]
koonopek: Yeah, i think it should be fixed but for sure not in this PR

cehali (approved):
Good job, probably we should wait for eslint PR to merged first

packages/cache-service/package.json
koonopek: Okey it is opinionated but set to 0.3.0 to merge

hatskier (approved):

PR #1098, opened 2023-09-15
fix: adjust calculation of usd in token

cehali:
Internal discussion: https://discord.com/channels/786251205008949258/1149746843983552627/1152259274668519484

### DIFF
--- a/packages/oracle-node/src/fetchers/SlippageAndLiquidityCommons.ts
+++ b/packages/oracle-node/src/fetchers/SlippageAndLiquidityCommons.ts
@@ -5,19 +5,16 @@ export const DEFAULT_AMOUNT_IN_USD_FOR_SLIPPAGE = 10_000; // 10k usd
 
 export const convertUsdToTokenAmount = (
   assetId: string,
-  decimalsMultiplier: number,
   amountInUsd: number
 ) => {
   return new Decimal(amountInUsd)
     .div(getRawPriceOrFail(assetId).value)
-    .mul(decimalsMultiplier)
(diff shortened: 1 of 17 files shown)


packages/oracle-node/src/fetchers/SlippageAndLiquidityCommons.ts
```diff
@@ -5,19 +5,16 @@ export const DEFAULT_AMOUNT_IN_USD_FOR_SLIPPAGE = 10_000; // 10k usd
 
 export const convertUsdToTokenAmount = (
   assetId: string,
-  decimalsMultiplier: number,
```
rpiszczatowski, line 8: This function is supposed to return Token amount in subtokens, i.e. if 1ETH = $2K then this should return 5ETH and 5ETH is 5*10^18 wei. So returned number should be 5*10^18

packages/oracle-node/test/fetchers/balancer.spec.ts
```diff
@@ -110,12 +110,12 @@ describe("balancer multi fetcher", () => {
             {
               direction: "buy",
               simulationValueInUsd: "10000",
-              slippageAsPercent: "100",
```
rpiszczatowski, line 113: Everywhere when I put `spippageAsPercent: "100"` in tests as an expected result test data was specifically set up so the result is 100. We shouldn't change it without good reason.

rpiszczatowski (changes requested):
This is bad path IMO

PR #1102, opened 2023-09-15
fix: fix maverick slippage calculation 2

koonopek:


packages/oracle-node/src/fetchers/maverick/MaverickFetcher.ts
```diff
@@ -265,31 +224,23 @@ class MultiCallHandler {
...
 
-    const bigBuyAmountOutScaled = this.getPrice(
-      bigBuyAmountOut,
-      this.amountIn1Token
-    );
+    const bigSellAmountOutScaled = tokens0Sell.div(tokens1Sell);
+
+    const tokens0Buy = this.token0Scaler.fromSolidityValue(bigBuyAmountOut);
+    const tokens1Buy = this.token1Scaler.fromSolidityValue(this.amountIn1Token);
+    const bigBuyAmountOutScaled = tokens1Buy.div(tokens0Buy);
```
rpiszczatowski, line 240: Probably now we should simply rename in as a price here and in response object. What do you think?

PR #1104, opened 2023-09-18
feat: add additional logs and checks to long dry runs

cehali:
This PR also fix avalanche long dry run


packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -81,14 +82,19 @@ const compareValuesFromSmallPackagesAndLocalCache = (
   dataFeedId: string,
   allFeedObjectsFromProd: DataPackagePlainObj[]
 ) => {
-  const dataPointValueFromLocal =
-    dataPackagesFromLocal[dataFeedId][0].dataPoints[0].value;
+  const dataFeedValueFromLocal =
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    dataPackagesFromLocal[dataFeedId]?.[0]?.dataPoints?.[0]?.value;
```
rpiszczatowski, line 87: We could adjust types instead to be specific about what is optional and what is obligatory.

If this is temporary for debugging purposes then it does not matter.

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -81,14 +82,19 @@ const compareValuesFromSmallPackagesAndLocalCache = (
...
+  if (!dataFeedValueFromLocal) {
+    console.log(`Missing data feed in local for ${dataFeedId}`);
+    return 0;
```
rpiszczatowski, line 90: Is returning 0 here what we want? This will not cause tests to fail in case of missing data.

cehali: You are right, changed to 100 and added comment

packages/integration/src/framework/compare-data-packages.ts
cehali: Changes types and adjusted the code

packages/integration/src/framework/compare-data-packages.ts
```diff
@@ -154,8 +160,13 @@ const compareValuesFromBigPackageAndLocalCache = (
...
         continue;
       }
-      const dataFeedValueFromLocal =
-        dataPackagesFromLocal[dataFeedId][0].dataPoints[0].value;
+      let dataFeedValueFromLocal =
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        dataPackagesFromLocal[dataFeedId]?.[0]?.dataPoints?.[0]?.value;
+      if (!dataFeedValueFromLocal) {
+        console.log(`Missing data feed in local for ${dataFeedId}`);
+        dataFeedValueFromLocal = 0;
```
rpiszczatowski, line 167: Same question as above.

rpiszczatowski (approved):

packages/integration/src/framework/compare-data-packages.ts
cehali: Here situation is a little bit different, 0 is taken to deviation calculation as a new value, which will lead to deviation values set as `createSafeNumber(Number.MAX_SAFE_INTEGER)`

PR #1115, opened 2023-09-19
chore: prettier config cleanup

rpiszczatowski:
- we do not automatically format contracts
- we use only prettier, no .editorconfig or other tools for formatting
- we configure all the relevant tools in top-level package.json


packages/fuel-connector/package.json
```diff
@@ -22,6 +22,11 @@
     "@babel/preset-typescript": "^7.21.0",
     "@types/jest": "^29.5.4",
     "@types/node": "^18.14.6",
+    "@typescript-eslint/eslint-plugin": "^6.6.0",
```
koonopek, line 25: Can't we install all these dependencies once on the level of whole monorepo `/package.json`

koonopek (changes requested):

cehali (approved):
👍

PR #1116, opened 2023-09-19
chore: eslint for evm-connector

rpiszczatowski:


packages/evm-connector/benchmarks/two-phase-approach-benchmark.ts
```diff
@@ -149,45 +150,51 @@ describe("Benchmark", function () {
...
         address: string,
         ...args: Uint8Array[]
-      ) => any;
+      ) => Promise<ContractTransaction>;
 
       switch (benchmarkParams.passedArgumentsCount) {
         case 3:
           sendStructRequestFunction =
-            wrappedStorageStructureModel.sendRequestWith3Args;
+            wrappedStorageStructureModel.sendRequestWith3Args as typeof sendStructRequestFunction;
```
koonopek, line 171: That is why this rules sucks sometimes, this typing doesnt add any value

packages/evm-connector/src/helpers/add-contract-wait.ts
```diff
@@ -9,45 +13,48 @@ export const addContractWait = (
...
-    return wait(confirmations).then((receipt: any) => {
-      receipt.events = receipt.logs.map((log: any) => {
-        let event: any = <Event>deepCopy(log);
-        // let parsed: LogDescription = null;
-        let parsed: any = null;
-        try {
-          parsed = contract.interface.parseLog(log);
-        } catch (e) {}
+    return wait(confirmations).then((receipt: TransactionReceipt) => {
+      (receipt as unknown as { events: Event[] }).events = receipt.logs.map(
```
koonopek, line 17: Why it this better than using @ts-ignore?

packages/evm-connector/src/wrappers/BaseWrapper.ts
```diff
@@ -60,15 +59,15 @@ export abstract class BaseWrapper {
...
-  setContractForFetchingDefaultParams(contract: Contract) {
+  setContractForFetchingDefaultParams(contract: T) {
     this.contract = contract;
   }
 
-  overwriteEthersContract(contract: Contract): Contract {
+  overwriteEthersContract(contract: T): T {
     this.setContractForFetchingDefaultParams(contract);
-    const contractPrototype = Object.getPrototypeOf(contract);
+    const contractPrototype = Object.getPrototypeOf(contract) as unknown;
```
koonopek, line 68: Again this casting introduce more confusion then value.

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -56,23 +58,25 @@ export class DataServiceWrapper extends BaseWrapper {
...
-      const dataServiceId = await this.contract.getDataServiceId();
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      const dataServiceId = (await this.contract.getDataServiceId()) as string;
       return dataServiceId;
-    } catch (e: any) {
+    } catch (e) {
       throw new Error(
         `DataServiceId not provided and failed to get it from underlying contract. Error: ` +
-          e?.message
+          (e as Error).message
```
koonopek, line 67: `e?.message` it is not the same, please leave `?`

koonopek:
Most comments are generally about typescript/eslint rules.
Base on this PR I don't see that we catched some issues thanks to eslint/typescript, I would say we even introduced bug when avoiding `?` because eslint doesn't like it :)

packages/evm-connector/src/WrapperBuilder.ts
```diff
@@ -13,35 +13,33 @@ import {
   SimpleNumericMockWrapper,
 } from "./wrappers/SimpleMockNumericWrapper";
 
-export class WrapperBuilder {
-  constructor(private baseContract: Contract) {}
+export class WrapperBuilder<T extends Contract> {
+  constructor(private baseContract: T) {}
 
-  static wrap(contract: Contract): WrapperBuilder {
+  static wrap<T extends Contract>(contract: T): WrapperBuilder<T> {
```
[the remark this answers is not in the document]
rpiszczatowski: I discussed this with @hatskier. We also agreed that this is a good change. Also, I don't think we need to release it immediately.

packages/evm-connector/benchmarks/two-phase-approach-benchmark.ts
rpiszczatowski: Totally agree. On the other hand, I consider this as a price we need to pay to have all the other places checked for us.

Maybe there is some complicated way of rewriting this code to get rid of those type assertions. But because this is only here I'd leave the code as is.

packages/evm-connector/src/wrappers/BaseWrapper.ts
rpiszczatowski: I do not see how. I see the value in banning almost all the usages of `any`. Unfortunately, I see no way to say what 'almost' means in a way that will ban only what I want.

To me, `Object.getPrototypeOf` should return `unknown` so that unsafe usages are discouraged by default. But it doesn't...

But we could use `as object` instead and get rid of a type assertion below or simply inline the call not introducing a variable we don't want to have type any...
I'll do one of those things.

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
rpiszczatowski: Really, we assume someone can throw `undefined`?

packages/evm-connector/src/helpers/add-contract-wait.ts
rpiszczatowski: This is the result of not doing types correctly from the beginning.

To me this is more verbose that @ts-ingore but I am ok with it. If you have a strong preference I can change the code.

rpiszczatowski: Fixed by using proper type for receipt and removing type assertions.

packages/evm-connector/src/wrappers/DataServiceWrapper.ts
```diff
@@ -56,23 +58,25 @@ export class DataServiceWrapper extends BaseWrapper {
...
   private async getUniqueSignersThresholdFromContract(): Promise<number> {
     try {
-      return await this.contract.getUniqueSignersThreshold();
-    } catch (e: any) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      return (await this.contract.getUniqueSignersThreshold()) as number;
+    } catch (e) {
       throw new Error(
         `UniqueSignersCount not provided and failed to get it from underlying contract. Error: ` +
-          e?.message
```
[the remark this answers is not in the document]
rpiszczatowski: changed

koonopek (approved):

PR #1125, opened 2023-09-19
feat: lower minValidSourcesPercentage

koonopek:


packages/oracle-node/manifests/data-services/test.json
```diff
@@ -0,0 +1,28 @@
+{
```
cehali, line 1: This is unnecessary, I merged it with main on PR removing old dex'es

cehali (dismissed):
One small comment

Lukasz2891 (approved):

PR #1128, opened 2023-09-20
feat: add tokens for Mento

cehali:


rpiszczatowski (dismissed):
+1

packages/oracle-node/manifests/data-services/primary.json
```diff
@@ -355,6 +355,48 @@
...
+        "gate",
+        "huobipro",
+        "kucoin",
+        "latoken",
+        "mexc",
+        "okx"
+      ]
+    },
+    "BRL": {
+      "source": ["ecb", "kaiko", "kaiko-v2", "twelve-data", "yf-unofficial"]
```
hatskier, line 381: Let's not use `yf-unofficial` please on prod nodes

packages/oracle-node/manifests/data-services/primary.json
```diff
@@ -355,6 +355,48 @@
...
+    "EUR": {
+      "source": [
+        "binance",
+        "currencycom",
+        "ecb",
+        "kaiko",
+        "kaiko-v2",
+        "kraken",
+        "twelve-data",
+        "yf-unofficial"
```
hatskier, line 392: Let's not use yf-unofficial please on prod nodes

packages/oracle-node/manifests/data-services/primary.json
```diff
@@ -355,6 +355,48 @@
...
+      ]
+    },
+    "cEUR": {
```
hatskier, line 395: I would change cEUR to CELO/EUR

packages/oracle-node/manifests/data-services/primary.json
```diff
@@ -355,6 +355,48 @@
...
+      "source": ["non-usd-based"]
+    },
+    "cBRL": {
```
hatskier, line 398: I would change cBRL to CELO/BRL

packages/oracle-node/manifests/data-services/main.json
```diff
@@ -1091,6 +1102,12 @@
     "LVL": {
       "source": ["bitget", "coingecko", "coinmarketcap", "mexc"]
     },
+    "cEUR": {
```
hatskier, line 1105: Don't we want to change it as well?

packages/oracle-node/manifests/data-services/primary.json
```diff
@@ -380,6 +380,59 @@
...
+        "huobipro",
+        "kucoin",
+        "latoken",
+        "mexc",
+        "okx"
+      ]
+    },
+    "BRL": {
+      "source": [
+        "binance-api-fetcher",
```
hatskier, line 407: Can we please remove the word fetcher to be consistent and to make it more readable for app.redstone.finance users (especially ones who will check it after our Mento proposal)?

packages/oracle-node/manifests/data-services/primary.json
```diff
@@ -380,6 +380,59 @@
...
+        "coingecko",
+        "gate",
+        "huobipro",
+        "kucoin",
+        "latoken",
+        "mexc",
+        "okx"
+      ]
+    },
+    "BRL": {
```
hatskier, line 405: The number of sources is really impressive now, good job!

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
```diff
@@ -0,0 +1,72 @@
...
+    private apiFetcherConfig: Record<string, ApiFetcherConfig>
+  ) {
+    super(`${apiName}-api-fetcher`);
+  }
+
+  override async makeRequest(
+    dataFeedId: string
+  ): Promise<AxiosResponse<unknown>> {
+    const { url } = this.apiFetcherConfig[dataFeedId];
+    return (await axios.get(url)).data as AxiosResponse<unknown>;
```
rpiszczatowski, line 29: Technically it is better to do it like this:
```suggestion
    return (await axios.get<unknown>(url)).data;
```
without explicit type assertions.

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
```diff
@@ -0,0 +1,72 @@
...
+    const rawPrice = this.extractPriceFromResponse(dataFeedId, response);
+    return this.makeAdditionalCalculations(dataFeedId, rawPrice).toNumber();
+  }
+
+  private extractPriceFromResponse(
+    dataFeedId: string,
+    response: unknown
+  ): Decimal {
+    const { url, jsonpath } = this.apiFetcherConfig[dataFeedId];
+    const [extractedValue] = jp.query(response, jsonpath) as unknown[];
```
rpiszczatowski, line 46: Maybe instead of `unknown[]` we can use `Decimal.Value[]` which is compatible with how we want to use this return value. This way you could remove type assertion below in `Decimal` construction.

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
```diff
@@ -0,0 +1,72 @@
...
+  ): number | undefined {
+    const response = responses[dataFeedId];
+    const rawPrice = this.extractPriceFromResponse(dataFeedId, response);
+    return this.makeAdditionalCalculations(dataFeedId, rawPrice).toNumber();
+  }
+
+  private extractPriceFromResponse(
+    dataFeedId: string,
+    response: unknown
+  ): Decimal {
```
rpiszczatowski, line 44: Is concept of decimals relevant here? I see no handling of it.

rpiszczatowski (changes requested):
minor changes required.

packages/oracle-node/manifests/data-services/main.json
cehali: Yes, probably issue when resolving confilcts

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
cehali: I am not sure if I understand correctly but the casting to decimal value is used to "validate" value from response. Decimals.js will check if the value is a proper numberlike. Also, decimal value is used in `makeAdditionalCalculations`, maybe there aren't big calculations but I wanted to be consistent with other fetchers

rpiszczatowski: Sorry, the question was about decimals like in WETH has 18 decimals and USDC 6, not the decimals.js library.
So the question would be: Does this API operate on ETH or WEI level, etc.?

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
```diff
@@ -0,0 +1,70 @@
...
+    const { url, jsonpath } = this.apiFetcherConfig[dataFeedId];
+    const [extractedValue] = jp.query(response, jsonpath) as Decimal.Value[];
+
+    const isEmptyString =
+      typeof extractedValue === "string" && extractedValue.length === 0;
+    if (isEmptyString) {
+      throw new Error(`Request to ${url} returned empty string`);
+    }
+
+    return new Decimal(extractedValue as string);
```
koonopek, line 52: [can be omitted] Why casting to string instead of Decimal.Value

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  private makeAdditionalCalculations(
```
koonopek, line 55: [can be omitted] Intresting function name :D

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
```diff
@@ -0,0 +1,70 @@
...
+import { getLastPriceOrFail } from "../../db/local-db";
+
+interface ApiFetcherConfig {
+  url: string;
+  jsonpath: string;
+  pairedToken?: string;
+  isReverted?: boolean;
+}
+
+export class ApiFetcher extends MultiRequestFetcher {
```
koonopek, line 17: Nice!

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
```diff
@@ -0,0 +1,70 @@
...
+    const rawPrice = this.extractPriceFromResponse(dataFeedId, response);
+    return this.makeAdditionalCalculations(dataFeedId, rawPrice).toNumber();
+  }
+
+  private extractPriceFromResponse(
+    dataFeedId: string,
+    response: unknown
+  ): Decimal {
+    const { url, jsonpath } = this.apiFetcherConfig[dataFeedId];
+    const [extractedValue] = jp.query(response, jsonpath) as Decimal.Value[];
```
koonopek, line 44: [Important] Should we also very if values is not undefined?

packages/oracle-node/test/fetchers/api-fetcher.spec.ts
```diff
@@ -0,0 +1,54 @@
+import axios from "axios";
```
koonopek, line 1: We should have propably test for each ApiFetcher from api-fetchers-config. To not break something when we will be doing changes to ApiFetcher

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
rpiszczatowski: No casting is required. Casting is generally bad. We should try to avoid it if possible.
This is an argument with `[can be omitted]` part :)

packages/oracle-node/test/fetchers/api-fetcher.spec.ts
rpiszczatowski: Agree. Maybe we can mock a single success response for each of the fetchers and one test for 'expected' failures like `undefined` response.

rpiszczatowski (commented):
.

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
cehali: Currently all values are returned on ETH level

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
cehali: Fixed

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
cehali: Changed

packages/oracle-node/test/fetchers/api-fetcher.spec.ts
cehali: Added

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    if (typeof extractedValue === "undefined") {
```
rpiszczatowski, line 48:
```suggestion
    if (xtractedValue === undefined) {
```
Is simpler, but does the same, right?

packages/oracle-node/test/fetchers/api-fetcher.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+    mockedAxios.get.mockResolvedValue({ data: exampleResponse });
+    await saveMockPriceInLocalDb(0.99, "USDT");
+
+    const fetcher = fetchers["kucoin-api"]!;
+    const result = await fetcher.fetchAll(["BRL"]);
+
+    expect(result).toEqual([
+      {
+        symbol: "BRL",
+        value: 0.20691159205641568,
```
rpiszczatowski, line 49: [nice to have]

I assume this value is `1 / (4.8818 * 0.99)`.
I prefer to use values that make clear what to expect.
You could e.g. introduce the variables `brlPrice = 4.8818` and `usdtPrice=0.99` then use them in mock definition and then in expected value like `(1/(brlPrice*usdtPrice))`. For such a simple logic I find the described approach valuable as it makes clear what the fetcher should calculate

rpiszczatowski (dismissed):
Nice

packages/oracle-node/test/fetchers/api-fetcher.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    expect(result).toEqual([
+      {
+        symbol: "BRL",
+        value: 0.20233085140822274,
+      },
+    ]);
+  });
+
+  test("coinbase api", async () => {
```
koonopek, line 133: Great!

koonopek (approved):

packages/oracle-node/src/fetchers/api-fetcher/ApiFetcher.ts
cehali: image

packages/oracle-node/src/fetchers/api-fetcher/api-fetcher-configs.json
```diff
@@ -0,0 +1,51 @@
+{
+  "kucoin": {
```
hatskier, line 2: Great idea to create a fetcher like this one!

hatskier (approved):
Good job!

hatskier (approved):

PR #1129, opened 2023-09-20
chore: improve getEnv function type

rpiszczatowski:


packages/on-chain-relayer/src/FilesystemConfigProvider.ts
```diff
@@ -7,12 +7,21 @@ import {
 import fs from "fs";
 import { makeConfigProvider } from "./make-config-provider";
 
-const getFromEnv = (name: string, optional: boolean = false) => {
+type GetFromEnvType = {
```
Lukasz2891, line 10: It would be nice to move it to utils and reuse in all other places ;)

Lukasz2891 (approved):

packages/on-chain-relayer/src/FilesystemConfigProvider.ts
rpiszczatowski: There is no 'other place'. Other place uses a little different signature.

PR #1132, opened 2023-09-20
fix: docker images build

rpiszczatowski:


cehali (approved):

Lukasz2891:
As mentioned on daily - it would be nice to have it upgraded for ton-relayer too ;)

PR #1133, opened 2023-09-20
chore: small test fix

rpiszczatowski:
ERROR that this PR fixes:

✕ /data-packages/latest (GET) - should serve cached result (3[74](https://github.com/redstone-finance/redstone-oracles-monorepo/actions/runs/6249210971/job/16965915096?pr=1134#step:6:75) ms)
      ✓ /data-packages/latest (GET) - should be called twice for different data-service-id (379 ms)

  ● Data packages (e2e) › cache › /data-packages/latest (GET) - should serve cached result

    RangeError [ERR_OUT_OF_RANGE]: The value of "value" is out of range. It must be >= 0 and <= 4294967295. Received -1

      at Function.Object.<anonymous>.ObjectId.generate (../../../node_modules/bson/src/objectid.ts:168:12)
      at new ObjectId (../../../node_modules/bson/src/objectid.ts:73:28)
      at NativeCollection.<computed> [as createIndex] (../../../node_modules/mongoose/lib/drivers/node-mongodb-native/collection.js:108:18)
      at create (../../../node_modules/mongoose/lib/model.js:1915:22)
      at ../../../node_modules/mongoose/lib/model.js:1870:7


koonopek (approved):
Nice!

PR #1135, opened 2023-09-20
v0.3.3-alpha.0

koonopek:
* Testing if publishing is working correctly (with alpha version) - works
* Had to add in every package.json publishConfig: access: public, because packages starting from `@org-name` are by default private


packages/starknet-connector/package.json
```diff
@@ -12,13 +12,16 @@
...
   },
-  "devDependencies": {},
   "files": [
     "dist/**/*"
-  ]
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "gitHead": "c7b70dfbd7450063c8aabb891b2fc46e75df440a"
```
rpiszczatowski, line 26: Should we really have 'gitHead' property in our repo.
I haven't found lots of documentation on it, but according to this: https://github.com/lerna/lerna/issues/1880 'gitHead' should be a transient property only set for publishing time.

packages/starknet-connector/package.json
```diff
@@ -12,13 +12,16 @@
...
     "ethers": "^5.7.2",
     "starknet": "^5.19.5"
   },
-  "devDependencies": {},
   "files": [
     "dist/**/*"
-  ]
+  ],
+  "publishConfig": {
+    "access": "public"
```
rpiszczatowski, line 24: Do we really want to publish starknet-connector, starknet-relayer and ton-connector?

packages/fuel-connector/package.json
```diff
@@ -24,5 +24,9 @@
   "files": [
     "dist/src/**/*",
     "src/autogenerated/prices.bin"
-  ]
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "gitHead": "c7b70dfbd7450063c8aabb891b2fc46e75df440a"
```
rpiszczatowski, line 31: Do we want to publish this connector?

packages/eth-contracts/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@redstone-finance/eth-contracts",
-  "version": "0.3.0",
+  "version": "0.3.2-alpha.0",
```
rpiszczatowski, line 3: Really 0.3.2? Other packages have 0.3.3.

rpiszczatowski (changes requested):
Couple of questions.

packages/starknet-connector/package.json
koonopek: Why not? @Lukasz2891

packages/fuel-connector/package.json
koonopek: Why not? @Lukasz2891

packages/starknet-connector/package.json
koonopek: So it wasn't cleared by lerna :(

packages/starknet-connector/package.json
koonopek: Also it is already published. I haven't configured what should be published in this PR it was done on first PR on lerna

packages/eth-contracts/package.json
koonopek: fixed! Good catch

packages/starknet-connector/package.json
koonopek: Removed

rpiszczatowski (approved):

PR #1140, opened 2023-09-21
feat: use redstone memoize for cache

koonopek:
* improve type inference in Memoize function
* cache
* tests for cache


packages/utils/src/common/memoize.ts
```diff
@@ -1,47 +1,50 @@
...
       cacheKey.length < EXPECTED_MAX_CACHE_KEY_LENGTH_PER_FN,
-      `Assumed cache key will not be longer then ${EXPECTED_MAX_CACHE_KEY_LENGTH_PER_FN}`
+      `Assumed cache key will not be longer than ${EXPECTED_MAX_CACHE_KEY_LENGTH_PER_FN}`
     );
 
     // to avoid caching results forever
     cleanStaleCacheEntries(cache, ttl);
 
-    if (!cache[cacheKey] || Date.now() - cache[cacheKey]!.lastSet > ttl) {
+    if (!cache[cacheKey]) {
```
rpiszczatowski, line 36: `cleanStaleCacheEntries` takes care of stale cache entries, so this check was redundant.

packages/utils/src/common/memoize.ts
```diff
@@ -1,47 +1,50 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function memoize<F extends (...args: any[]) => Promise<unknown>>({
   functionToMemoize,
   ttl,
-  cacheKeyBuilder = JSON.stringify,
-}: MemoizeArgs<T, A>): (...args: A) => Promise<T> {
-  const cache: Partial<Record<string, MemoizeCache<T>>> = {};
+  cacheKeyBuilder = stringifyAndJoin,
+}: MemoizeArgs<F>): F {
+  const cache: Partial<Record<string, MemoizeCache<ReturnType<F>>>> = {};
```
rpiszczatowski, line 23: >>>> (4 levels of template), nice :)

packages/utils/src/common/memoize.ts
koonopek: True

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -202,3 +214,18 @@ const convertHexToNumber = (hex: string): number => {
   }
   return number;
 };
+
+/**
+ * It takes fields of transaction which might have change output
+ * of call method
+ */
+const txCacheKeyBuilder = async (
```
koonopek, line 222: This is important one

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -11,6 +11,8 @@ import {
 } from "./ProviderWithFallback";
 
 const BLOCK_NUMBER_TTL = 200;
+// 10 min
+const AGREED_RESULT_TTL = 600_000;
```
koonopek, line 15: It could be one hour, but it would increase memory consumption, and would nothing because our "twaps" are no longer than 10 mins

cehali (approved):
LGTM

packages/utils/src/common/memoize.ts
```diff
@@ -1,26 +1,32 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
```
rpiszczatowski, line 1: Is this necessary?

rpiszczatowski (approved):

packages/rpc-providers/src/ProviderWithAgreement.ts
```diff
@@ -202,3 +214,18 @@ const convertHexToNumber = (hex: string): number => {
...
+ */
+const txCacheKeyBuilder = async (
+  transaction: Deferrable<TransactionRequest>,
+  blockTag: BlockTag
+) =>
+  String(await transaction.chainId) +
+  String(await transaction.to) +
+  String(await transaction.from) +
+  String(await transaction.data) +
+  String(await transaction.customData) +
```
hatskier, line 230: For very rare cases it can be possible to have collisions between different calls, if we don't have a separator (e.g. `#`) between `transaction.customData` and `blockTag`. Rest should be fine, thanks to `0x` separator

hatskier (approved):
Good job!

packages/rpc-providers/src/ProviderWithAgreement.ts
koonopek: Will add in next PR

PR #1142, opened 2023-09-21
chore: disable deviation checks for bybit due to inconsistent results

rpiszczatowski:


packages/integration/src/framework/check-sources-deviations.ts
```diff
@@ -2,18 +2,24 @@ import { SourceDeviationsPerDataFeed } from "./run-long-price-propagation-core-t
...
       if (deviation > maxPercentageValueDifference) {
+        const ignoreString = sourcesToSkip.includes(source)
+          ? "(source ignored)"
+          : "";
         console.log(
-          `Source deviation for ${dataFeedId} from ${source} is bigger than maximum (${maxPercentageValueDifference}%) - ${deviation}%`
+          `Source deviation for ${dataFeedId} from ${source} is bigger than maximum (${maxPercentageValueDifference}%) - ${deviation}% ${ignoreString}`
         );
-        deviationsBiggerThanAllowed += 1;
+        if (!ignoreString) {
```
koonopek, line 20: [can be ignored] `ignoreString === ""` weird that eslint is not yelling here

packages/integration/src/arbitrum-long-price-propagation-core.ts
```diff
@@ -1,5 +1,7 @@
 import { runLongPricePropagationCoreTest } from "./framework/run-long-price-propagation-core-test";
 
+const SOURCES_TO_SKIP = ["bybit"]; // bybit returns inconsistent results for GMX, skip for now
```
koonopek, line 3: Shouldn't this be global for all integration tests not only for arbitrum?
I think this can be `process.env`

cehali: I agree

packages/integration/src/framework/check-sources-deviations.ts
```diff
@@ -2,18 +2,24 @@ import { SourceDeviationsPerDataFeed } from "./run-long-price-propagation-core-t
...
+  maxPercentageValueDifference: number,
+  sourcesToSkip: string[] = []
 ) => {
   let deviationsBiggerThanAllowed = 0;
   const dataFeeds = Object.keys(deviationsPerDataFeed);
   for (const dataFeedId of dataFeeds) {
     const deviationsPerSource = deviationsPerDataFeed[dataFeedId];
     for (const [source, deviation] of Object.entries(deviationsPerSource)) {
       if (deviation > maxPercentageValueDifference) {
+        const ignoreString = sourcesToSkip.includes(source)
```
cehali, line 14: [can be ignored] I would do this a little bit differently. This string comparing below is a little bit odd 😄 
```js
const isSourceIgnored = sourcesToSkip.includes(source);
const logText = `Source deviation for ${dataFeedId} from ${source} is bigger than maximum (${maxPercentageValueDifference}%) - ${deviation}%`;
console.log(`${logText} ${isSourceIgnored ? (source ignored) : ""}`);
if (!ignoreString) {
      deviationsBiggerThanAllowed += 1;
}
```

cehali (dismissed):
LGTM, 1 minor comment

packages/integration/src/arbitrum-long-price-propagation-core.ts
rpiszczatowski: we have separate workflows for arbitrum, avalanche, primary. Fir each of them need to specify `process.env` separately.

cehali (approved):
👍

PR #1143, opened 2023-09-22
Integrtion tests relayer conditions

koonopek:


packages/cache-service/.env-1
```diff
@@ -0,0 +1,14 @@
+MONGO_DB_URL=mongodb://127.0.0.1:49386/
```
rpiszczatowski, line 1: Isn't this file accidentally commited?

packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -48,8 +48,6 @@ const portNumberForInstance = (
 export const buildCacheLayer = async () =>
```
rpiszczatowski, line 48: Do we need this function anymore?

packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -199,13 +197,16 @@ export const waitForDataPackages = async (
   );
 };
 
-export const waitForDataAndDisplayIt = async (instance: CacheLayerInstance) => {
+export const waitForDataAndDisplayIt = async (
+  instance: CacheLayerInstance,
+  expectedDataPackageCount: number = 1
+) => {
   // Waiting for data packages to be available in cache service
   const ALL_FEEDS_KEY = consts.ALL_FEEDS_KEY as string;
```
rpiszczatowski, line 205: You can remove this cast.

packages/integration/tsconfig.json
```diff
@@ -9,6 +9,9 @@
     },
     {
       "path": "../protocol/tsconfig.build.json"
+    },
+        {
```
rpiszczatowski, line 13: indentation

packages/on-chain-relayer/.env-1
```diff
@@ -0,0 +1,13 @@
+RELAYER_ITERATION_INTERVAL=../integration/relayerManifest.json
```
rpiszczatowski, line 1: Probably accidentally committed. This problem will disappear when we switch to passing env by variable to not by file.

packages/on-chain-relayer/test/monorepo-integration-tests/scripts/verify-mock-prices-in-adapter.ts
```diff
@@ -19,11 +18,18 @@ import { PriceFeedsAdapterWithoutRounds } from "../../../typechain-types";
...
   const bytes32Symbols = Object.keys(pricesToVerify).map(formatBytes32String);
-  const expectedPrices = Object.values(pricesToVerify);
-
   const oracleValues = await contract.getValuesForDataFeeds(bytes32Symbols);
-  expect(oracleValues.length).to.eq(expectedPrices.length);
-  for (let i = 0; i < oracleValues.length; i++) {
-    expect(oracleValues[i]).to.eq(BigNumber.from(expectedPrices[i] * 10 ** 8));
+
+  let oracleValuesIndex = 0;
+  for (const symbol of Object.keys(pricesToVerify)) {
```
rpiszczatowski, line 24: You have an object you want to iterate through, having access to key, value and index of objects property. More idiomatic way to do this would be

```suggestion
  for (const [oracleValuesIndex, [symbol, price]] of Object.entries(
    pricesToVerify
  ).entries()) {
    const expectedPrice = price * 10 ** 8;
    // ...
  }
```

packages/oracle-node/mock-prices.json
```diff
@@ -0,0 +1 @@
+{"BTC":16000,"__DEFAULT__":42}
```
rpiszczatowski, line 1: new line

packages/utils/src/common/time.ts
```diff
@@ -18,3 +21,6 @@ export const timeout = <T>(
     ),
   ]).finally(() => clearTimeout(timer));
 };
+
+export const sleep = (ms: number) =>
+  new Promise((resolve, rejects) => setTimeout(resolve, ms));
```
rpiszczatowski, line 26: `rejects` --> `_rejects`

koonopek: That is the job of eslint, if he is not complaining then I dont see purpose of doing this :)

packages/oracle-node/mock-prices.json
koonopek: This file has to be removed

packages/on-chain-relayer/test/monorepo-integration-tests/scripts/verify-mock-prices-in-adapter.ts
koonopek: "more idomatic" seems like very opinionated thing, lets focus on important issues :)

packages/on-chain-relayer/.env-1
koonopek: yep, will remove

packages/integration/src/framework/cache-layer-manager.ts
koonopek: I havent touch this code, as far as I see. I m not refactoring castings in this PRs :) Can be done in seperate one

packages/integration/src/framework/cache-layer-manager.ts
koonopek: Yes long dry runs are still using them, will eliminate this in next pr

packages/cache-service/.env-1
koonopek: will remove

packages/on-chain-relayer/package.json
```diff
@@ -24,7 +24,8 @@
     "cron-parser": "^4.8.1",
     "dotenv": "^16.0.3",
     "ethers": "^5.7.2",
-    "toad-scheduler": "^2.2.0"
+    "toad-scheduler": "^2.2.0",
+    "zod": "^3.22.2"
```
rpiszczatowski, line 28: You can give a try to `yarn syncpack:list` and `yarn syncpack:fix-mismatches` utilities we have in top-level `package.json` to avoid multiple versions of `zod` in our codebase.

koonopek: Can do in separate PR

yarn.lock
```diff
@@ -5255,19 +5255,32 @@
...
-"@ton-community/func-js@0.6.3-tvmbeta.1", "@ton-community/func-js@^0.6.2":
+"@ton-community/func-js@0.6.3-tvmbeta.1":
   version "0.6.3-tvmbeta.1"
   resolved "https://registry.yarnpkg.com/@ton-community/func-js/-/func-js-0.6.3-tvmbeta.1.tgz#90ec7987f341d1a56862a72b6b0415eb42c04bbb"
   integrity sha512-6/Iz3bbDQFMzyFUc0tOLukY2l+ItXZJt7q5Jcs8edHlBsYhffHUIgXinPE7E70Bb0E3Cbs3p+8RxPSLLrUDGhQ==
   dependencies:
     "@ton-community/func-js-bin" "0.4.5-tvmbeta.1"
     arg "^5.0.2"
 
+"@ton-community/func-js@^0.6.2":
```
rpiszczatowski, line 5276: I don't understand where this change comes from.
Also when I perform `yarn install` on your PR I get some changes regarding ton in yarn.lock.

packages/utils/src/common/time.ts
rpiszczatowski: But it does complain. The only thing is that unused things are marked as warnings not errors and build does not fail. You should see warnings in your IDE.

We have also unused imports in this file.

Probably we should switch all problems we care about to error level. I can do it in a separate PR. But you can fix these particular instances of warnings as we don't want to introduce new issues

rpiszczatowski (approved):
Minor comments.

yarn.lock
koonopek: in next PR

packages/utils/src/common/time.ts
koonopek: next pr

PR #1144, opened 2023-09-22
feat: integration tests - allow for different prices being returned f…

rpiszczatowski:
…rom each node


cehali (approved):
LGTM just shorten title or when merging make it to fit in commit title

PR #1153, opened 2023-09-25
Sign all data points

rpiszczatowski:


packages/integration/src/framework/run-long-price-propagation-core-test.ts
```diff
@@ -83,13 +83,12 @@ export const runLongPricePropagationCoreTest = async (
...
   const iterationsCount =
     nodeWorkingTimeInMilliseconds / nodeIntervalInMilliseconds;
-  const timestampsRange = [...Array(iterationsCount).keys()];
-  const timestampsCountToAnalyze = timestampsRange.slice(
-    0,
-    iterationsCount - coldStartIterationsCount
-  );
   const fetchDataPackagesPromises = [];
-  for (const timestampDiffNumber of timestampsCountToAnalyze) {
+  for (
```
rpiszczatowski, line 87: To me it looked strange that we generated a list of numbers from 0 to n and iterated through that list with 'foreach' instead of using plain old 'for' loop

rpiszczatowski, line 87: To me, it looked strange that we generated a list of numbers from 0 to n and iterated through that list with 'foreach' instead of using plain old 'for' loop

packages/oracle-node/test/dry-run/dry-run-test.spec.ts
```diff
@@ -125,18 +120,14 @@ describe("Main dry run test", () => {
...
-      ) {
+      if (!pricesPerDataFeedId[token]) {
         console.log(
           `Missing token ${token} in broadcaster during dry run test`
         );
       }
       if (config.additionalCheck) {
         config.additionalCheck(token, currentDataFeedPrice?.value);
       }
-      checkTokensBroadcasted(pricesPerDataFeedId, tokensWithoutSkipSigning);
```
cehali, line 139: Why did you decide to remove this check?

cehali: After this change we check all sources but if we make the wrong change to aggregation or sanitizing process we won't have any information about it

rpiszczatowski: You are right.

rpiszczatowski: I've moved the call above the loop.

packages/oracle-node/src/aggregated-price-handlers/DataPackageBroadcastPerformer.ts
```diff
@@ -55,20 +55,10 @@ export class DataPackageBroadcastPerformer
     pricesService: PricesService,
     iterationContext: IterationContext
   ): Promise<void> {
-    // Excluding "helpful" prices, which should not be signed
-    // "Helpful" prices (e.g. AVAX_SPOT) can be used to calculate TWAP values
-    const pricesForSigning =
-      pricesService.filterPricesForSigning(aggregatedPrices);
-
-    // When nothing to broadcast
-    if (pricesForSigning.length === 0) {
```
koonopek, line 64: Why do we remove this check?

packages/oracle-node/src/fetchers/PricesService.ts
```diff
@@ -382,10 +383,10 @@ export default class PricesService {
     }
   }
 
-  filterPricesForSigning(
-    prices: PriceDataAfterAggregation[]
-  ): PriceDataAfterAggregation[] {
-    return prices.filter((p) => !this.manifest.tokens[p.symbol]!.skipSigning);
+  filterDataPointsForBigPackage(dataPoints: DataPoint[]): DataPoint[] {
```
koonopek, line 386: This is general comment, but this class has already 400+ lines, and we are writning here methods just because it has access to manifest. It smells :)

koonopek: It should be seperate method in my opinion which accepts additional param manifest

koonopek (approved):

koonopek (approved):
So in new behaviour we will sign all feeds, and skipSigning will be only used for BigDataPackage?
If so then it also requires adjusting in monitoring (maybe also some dry runs), we also have their logic for skipSgining :)
Also, maybe we could also rename this variable:
https://github.com/redstone-finance/redstone-oracles-monorepo/blob/acddfb4dabbfd22d82c48a57cd250b689e779fca/packages/oracle-node/src/types.ts#L30
From `skipSigning` to `excludeFromBigDataPackage`

packages/oracle-node/src/aggregated-price-handlers/DataPackageBroadcastPerformer.ts
rpiszczatowski: Because it is being done in NodeRunner. This check effectively would now check if there is at least one price fetched for the whole iteration.

packages/oracle-node/src/fetchers/PricesService.ts
rpiszczatowski: agreed.

rpiszczatowski:
I agreed with @hatskier that for now, the name skipSigning stays.

PR #1170, opened 2023-09-28
chore: pass extraEnv to each component instead of creating specific .…

rpiszczatowski:
…env file per instance


packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -90,28 +91,17 @@ export const startAndWaitForCacheLayer = async (
...
+  const extraEnv: Record<string, string> = {
+    MONGO_DB_URL: memoryMongoDbUrl,
+    API_KEY_FOR_ACCESS_TO_ADMIN_ROUTES: "hehe",
+    ENABLE_DIRECT_POSTING_ROUTES: "true",
+    ENABLE_STREAMR_LISTENING: "false",
+    USE_MOCK_ORACLE_STATE: "true",
+    ENABLE_HISTORICAL_DATA_SERVING: String(enableHistoricalDataServing),
+    MAX_ALLOWED_TIMESTAMP_DELAY: "20000",
+  };
+  printDotenv(getLogPrefix(instance), DOTENV_PATH);
```
koonopek, line 103: I think it doesnt have sens any more

packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -188,7 +178,10 @@ export const waitForDataPackages = async (
     ],
     `Waiting ${feedId}`,
     CACHE_SERVICE_DIR,
-    { DOTENV_CONFIG_PATH: instance.dotenvPath! }
+    {
+      ...instance.extraEnv,
+      DOTENV_CONFIG_PATH: DOTENV_PATH,
```
koonopek, line 183: Can't we resign from using `DOTENV_CONFIG_PATH` if we are using extraEnv?

packages/integration/src/framework/integration-test-utils.ts
```diff
@@ -27,6 +27,15 @@ export const printDotenv = (label: string, filePath: string) => {
   console.groupEnd();
 };
 
+export const printExtraEnv = (
+  label: string,
+  extraEnv: Record<string, string>
+) => {
+  console.group(`${label} extra env`);
```
koonopek, line 34: Nice thing!

packages/integration/src/framework/oracle-node-manager.ts
```diff
@@ -77,29 +80,14 @@ const populateEnvVariables = (
...
-    dotenvPath
-  );
-  updateDotEnvFile("ECDSA_PRIVATE_KEY", HARDHAT_MOCK_PRIVATE_KEY, dotenvPath);
-  updateDotEnvFile(
-    "MOCK_PRICES_URL_OR_PATH",
-    getMockPricesPath(instance),
-    dotenvPath
-  );
-  printDotenv(getLogPrefix(instance), dotenvPath);
+  const extraEnv: Record<string, string> = {
```
koonopek, line 83: 22 lines -> 7 :)

packages/integration/src/framework/relayer-manager.ts
```diff
@@ -48,61 +48,44 @@ export const startRelayer = (
...
-    dotenvPath
-  );
-  updateDotEnvFile(
-    "HISTORICAL_PACKAGES_GATEWAYS",
-    JSON.stringify(cacheServiceUrls),
-    dotenvPath
-  );
-
-  printDotenv("on chain relayer", dotenvPath);
+  printDotenv(getLogPrefix(instance), dotenvPath);
```
koonopek, line 77: Is it still needed

koonopek (changes requested):

packages/integration/src/framework/cache-layer-manager.ts
rpiszczatowski: We still use defaults from .env.example. Because we support dotenv in our codebase it may be useful to test here whether it still works. But I won't insist if you think otherwise.

packages/integration/src/framework/oracle-node-manager.ts
rpiszczatowski: mostly because of our line length setting, but I so much agree that this is better :)

packages/integration/src/framework/cache-layer-manager.ts
koonopek: okey

koonopek (approved):

cehali (approved):

PR #1171, opened 2023-09-28
chore: fix gelato relayer tests

Lukasz2891:


packages/gelato-relayer/DEPLOYED_redstone
```diff
@@ -0,0 +1,2 @@
+QmXgAsKYKRBBmeGceeB61tXnUrNHDrhdFwdiEtHjDZT62d
```
rpiszczatowski, line 1: What is this file for?

rpiszczatowski (approved):

packages/gelato-relayer/DEPLOYED_redstone
Lukasz2891: > What is this file for?

That's an id of the function deployed to gelato. I haven't found another place to have it stored ;)

PR #1173, opened 2023-09-28
feat: add integrations test for gateway

koonopek:


packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -145,21 +167,35 @@ export const stopDirectAndPublicCacheServices = (
...
-  await startAndWaitForCacheService(instance, false);
+  await startAndWaitForCacheService(instance, { direct: false }, env);
 };
 
 export const startDirectAndPublicCacheServices = async (
   instance: CacheLayerInstance
 ) => {
-  await startDirectCacheService(instance);
-  await startPublicCacheService(instance);
+  const commonGatewayEnv: Record<string, string> = {
```
rpiszczatowski, line 189: What is the reason behind getting rid of extraEnv and preparing it from scratch here? And why here we have different variables set than above?

packages/integration/src/gateway-history-test.ts
```diff
@@ -0,0 +1,74 @@
...
+  await startAndWaitForOracleNode(oracleNodeInstance, [cacheLayerInstance]);
+  await waitForDataInMongoDb(cacheLayerInstance);
+  await verifyPricesInCacheService([cacheLayerInstance], { BTC: 42 });
+
+  const latestResponse = await axios.get<{
+    AAVE: { timestampMilliseconds: number }[];
+  }>(
+    `http://localhost:${cacheLayerInstance.directCacheServicePort}/data-packages/latest/mock-data-service`
+  );
+  console.log(latestResponse.data);
```
rpiszczatowski, line 46: Maybe you could add more info to this log message like
```
console.log(`latest data for mock service ${latestResponse.data}`);
```
or something along these lines

koonopek: I would rather remove this log tottaly

rpiszczatowski (commented):
one question left

packages/integration/src/framework/cache-layer-manager.ts
koonopek: Assigning env to instance had a bug `instance.extraEnv= env`, because we were then adding in startCache service thing like: `instance.extraEnv ={...env, port: 123}` which can lead to hazards, if two relayers are started simultaneously. I would recommend always minimizing the global state.

And this part I don't understand: `And why here we have different variables set than above?`

koonopek: I don't like that we have to ways, of starting cache-services, that is why this code is duplicated. But I didn't want to refactor it.

rpiszczatowski: I don't think there was a bug.
The basic idea behind the two methods is:
- startAndWaitForCacheLayer - starts the whole gateway including mongo. Should be called once at startup
- startAndWaitForCacheService - only starts cacheService. It exists because we want to simulate cache-service unavailability

The actual start of the cache service is always done via startAndWaitForCacheService function. I see no hazards here.

Additionally, the port that was once set will be read from the instance config and set to the same value.

rpiszczatowski: > And this part I don't understand: And why here we have different variables set than above?

Now we prepare env twice and in one place we set some variables that we do not set in the other.

rpiszczatowski: Because the required order of calls is to start the whole cache layer using `startAndWaitForCacheLayer` function first and only then use stop/start on cache service only we can simply add an assertion in `startAndWaitForCacheService` that would verify if the whole layer has been started already.

rpiszczatowski (approved):
approved as discussed.

packages/integration/src/framework/cache-layer-manager.ts
koonopek: Will be done in next PR

PR #1190, opened 2023-10-03
feat: replace data slicing with decoding

cehali:


packages/oracle-node/src/fetchers/evm-chain/arbitrum/evm-fetcher/sources/level-finance/LevelFinanceTokensRequestHandlers.ts
```diff
@@ -77,12 +77,15 @@ export class LevelFinanceTokensRequestHandlers implements IEvmRequestHandlers {
     trancheValuesResponses: MulticallParsedResponse[],
     address: string
   ) {
+    const contractInterface = new Interface(LevelFinanceLiquidityCalculatorAbi);
     return trancheValuesResponses.find((trancheValueResponse) => {
       if (trancheValueResponse.data) {
-        const llpTokenAddress = trancheValueResponse.data.slice(
-          ...INDEXES_OF_LLP_TOKEN_ADDRESS_IN_MULTICALL_DATA
+        const functionData = contractInterface.decodeFunctionData(
```
rpiszczatowski, line 83: Is this ok to call `decodeFunctionData` here and not `decodeFunctionResult` like everywhere else?

packages/oracle-node/src/fetchers/evm-chain/avalanche/evm-fetcher/sources/steak-hut/SteakHutTokensRequestHandler.ts
```diff
@@ -108,4 +100,35 @@ export class SteakHutTokensRequestHandlers implements IEvmRequestHandlers {
...
+      GET_UNDERLYING_ASSETS_FUNCTION_NAME
+    );
+
+    const contractInterface = new Interface(abi);
+    const underlyingAssets = contractInterface.decodeFunctionResult(
+      GET_UNDERLYING_ASSETS_FUNCTION_NAME,
+      underlyingAssetsResult
+    );
+    const firstAssetValueAsHex = (
+      underlyingAssets.totalX as BigNumber
```
rpiszczatowski, line 125: Just to make sure. While working with multicall library I learned that I don't get `BigNumber` instances from contracts directly, but rather JSON objects that can be passed as parameters to the `BigNumber.from` factory method. Are you sure we get here a real BigNumber instance and not a JSON object without `toHexString` method?

rpiszczatowski (approved):
LGTM

packages/oracle-node/src/fetchers/evm-chain/arbitrum/evm-fetcher/sources/level-finance/LevelFinanceTokensRequestHandlers.ts
cehali: Yes, in this custom implementation, there isn't enough information regarding which contract you check tranche, which is why you need to get it from input data.

packages/oracle-node/src/fetchers/evm-chain/avalanche/evm-fetcher/sources/steak-hut/SteakHutTokensRequestHandler.ts
cehali: Probably multicall library parses responses. Here I don't use it and receive BigNumber instance with `toHexString` 
image

PR #1191, opened 2023-10-03
feat: test update prices on chain on single node price change

rpiszczatowski:


packages/cache-service/src/data-packages/data-packages.service.ts
```diff
@@ -241,9 +239,12 @@ export class DataPackagesService {
   static isSignerAddressAlreadyInDbResponseForDataFeed(
     signerAddress: string,
     fetchedPackagesForDataFeed: CachedDataPackage[] | undefined
-  ) {
-    return fetchedPackagesForDataFeed?.some((dataPackage) =>
-      Object.values(dataPackage).includes(signerAddress)
+  ): boolean {
```
koonopek, line 242: Nice!

packages/integration/src/framework/cache-layer-manager.ts
```diff
@@ -1,4 +1,5 @@
 import { consts } from "@redstone-finance/protocol";
+import redstoneSDK, { DataPackagesRequestParams } from "@redstone-finance/sdk";
```
koonopek, line 2: If importing as a namespace (which is great), maybe lets stick to using it.

packages/integration/src/framework/relayer-manager.ts
```diff
@@ -198,7 +205,8 @@ export const verifyPricesNotOnChain = async (
   }
 };
 export const deployMockAdapter = async (
-  rpcUrl?: string
+  rpcUrl?: string,
+  initialPricesCacheInstances?: CacheLayerInstance[]
```
koonopek, line 209: I think this function is doing too much already (and this param is really weired, you have to read code to understand taht), what do you think about just calling after `deployMockAdapter`
  ```
    const currentPrices = await fetchDataPackages(initialPricesCacheInstances);
    await updateValuesInAdapterContract(contract, currentPrices);
  ```
Otherwise functions like this will grow and grow, by adding next params like this.

packages/integration/src/relayer-conditions/relayer-one-rpc-down.ts
```diff
@@ -60,12 +60,14 @@ const main = async () => {
   const rpcUrls = [secondHardhatInstance.url!, firstHardhatInstance.url!];
 
   // deploy on first instance
-  const adapterContract = await deployMockAdapter();
+  const adapterContract = await deployMockAdapter(undefined, [
```
koonopek, line 63: We don't need here inital prcies, also rpcUrl: undefiend looks pretty smelly

packages/integration/src/relayer-conditions/relayer-value-and-time-condition.ts
```diff
@@ -48,7 +48,9 @@ const main = async () => {
   await waitForDataAndDisplayIt(cacheLayerInstance);
   await startAndWaitForHardHat(hardhatInstance);
 
-  const adapterContract = await deployMockAdapter();
+  const adapterContract = await deployMockAdapter(undefined, [
```
koonopek, line 51: Same

packages/integration/src/relayer-conditions/relayer-value-deviation-condition.ts
```diff
@@ -47,7 +47,9 @@ const main = async () => {
   await waitForDataAndDisplayIt(cacheLayerInstance);
   await startAndWaitForHardHat(hardhatInstance);
 
-  const adapterContract = await deployMockAdapter();
+  const adapterContract = await deployMockAdapter(undefined, [
```
koonopek, line 50: Same

packages/integration/src/relayer-conditions/simple-price-propagation-classic.ts
```diff
@@ -63,6 +74,22 @@ const main = async () => {
...
+    {
+      BTC: 18000,
+      ETH: 1500,
+      __DEFAULT__: 42,
+    },
+    oracleNodeInstance1
+  );
+
+  // prices should get updated in adapter even if only single node delivers deviating prices
+  await verifyPricesOnChain(adapterContract, priceFeedContract, {
```
koonopek, line 87: We are breaking the rule, that one test, tests only one thing.
Also name of the test doesnt say anything about that it is also testing something else then only simple-rpice-rpopagation

packages/integration/relayerManifestSample.json
```diff
@@ -4,7 +4,7 @@
     "id": 31337
   },
   "updateTriggers": {
-    "deviationPercentage": 0,
+    "deviationPercentage": 0.1,
```
koonopek, line 7: Okey, I see now why we need inital prices, every where, however this changed how tests were behaving because most of the tests expected that this is set to 0. And for example now tests which were testing cron conditions or time condtion can be giving false positvies (passing because of value deviation, where time or cron could stop working)

koonopek: You can just pass conditions like this in `propagation-classic` instead of editing it globally
```
  startRelayer(relayerInstance, {
    cacheServiceInstances: [cacheLayerInstance],
    adapterContractAddress,
    intervalInMs: 5_000,
    updateTriggers: {
      timeSinceLastUpdateInMilliseconds: 60_000,
      deviationPercentage: 10,
    },
    isFallback: false,
  });
```

koonopek (changes requested):

packages/integration/src/framework/relayer-manager.ts
cehali: I agree, we could use `updateValuesInAdapterContract` as separate function used after deploying mock adapter

packages/integration/relayerManifestSample.json
cehali: I started to wonder if we have a check for `deviationPercentage` equal to 0. I see something like that: 
`js 
const shouldUpdatePrices = maxDeviation >= minDeviationPercentage!;
`
Does this mean that we upload every time the relayer runs its iteration?

packages/integration/src/framework/relayer-manager.ts
rpiszczatowski: I thought about the proposed solution. I decided to do what I did because it is actually obligatory to set initial prices whenever we have deviation conditions defined (which is currently 6/8 uses).
I'll create a separate method with a nicer interface that will combine deploying with setting initial prices.

packages/integration/relayerManifestSample.json
rpiszczatowski: Understood. I'll revert globally enabling deviation conditions and stick to setting it only when required.

packages/integration/src/relayer-conditions/simple-price-propagation-classic.ts
rpiszczatowski: I'll rename the test as simple price propagation was the initial example test. Now, its functionality is being tested along the way by all classic tests.

.github/workflows/monorepo-integration-tests.yml
```diff
@@ -26,10 +26,10 @@ jobs:
     uses: ./.github/workflows/integration-test-template.yml
     with:
       command: yarn ts-node src/prices-updates-node-failures.ts
-  integration-tests-classic:
+  integration-tests-classic-not-all-prices-change:
     uses: ./.github/workflows/integration-test-template.yml
     with:
-      command: yarn ts-node src/relayer-conditions/simple-price-propagation-classic.ts
+      command: yarn ts-node src/relayer-conditions/relayer-not-all-prices-change.ts
```
koonopek, line 32: I think that it should be relayer-not-all-nodes-deviates or something about that, this test is not testing if one price deviates then the other should be updated too

packages/integration/relayerManifestSample.json
koonopek: I think it is still set globally, at least github is saying so:
image
I can't diff from my last review cause you forced pushed and it shows, changes from other commits too

rpiszczatowski: I think you look at the wrong version. I've reverted the original change and only specified the deviation trigger in one relevant test.

Also, I force-pushed because I had conflicts and I really don't like merges :)

rpiszczatowski: > I started to wonder if we have a check for `deviationPercentage` equal to 0. I see something like that:
> 
> ```js
> const shouldUpdatePrices = maxDeviation >= minDeviationPercentage!;
> ```
> 
> Does this mean that we upload every time the relayer runs its iteration?

I think we handle it properly. While reading the manifest file we have this code:
```
  if (deviationPercentage) {
    updateConditions.push("value-deviation");
  }
```
And then we only check conditions contained in the `updateConditions` list.

packages/sdk/src/index.ts
```diff
@@ -103,9 +103,7 @@ const getDataPackagesSortedByDeviation = (
   }
 
   if (dataFeedId === consts.ALL_FEEDS_KEY) {
-    throw new Error(
```
koonopek, line 106: Weird wasn't it introduced lately by you?

PR #1195, opened 2023-10-04
chore: update hubble aylin testnet relayer manifest

cehali:


packages/oracle-node/manifests/data-services/primary.json
```diff
@@ -519,6 +519,30 @@
     },
     "EUROC/EUR": {
       "source": ["non-usd-based"]
+    },
+    "AVAX": {
```
cehali, line 523: I have to add it to primary manifest as we will use one data service to update all price feeds

cehali: By the way, I updated AVAX sources slightly

rpiszczatowski (approved):

PR #1199, opened 2023-10-04
feat: New docker files

koonopek:


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+COPY . .
+RUN cd "packages/${PACKAGE}" && echo $PWD && yarn install  --frozen-lockfile --non-interactive  && yarn build && yarn bundle 
+
+FROM node:16
+ARG PACKAGE
+
+COPY --from=build "/app/packages/$PACKAGE/package.json" yarn.lock
+COPY --from=build "/app/packages/$PACKAGE/package.json" package.json 
+
+RUN yarn install --pure-lockfile --non-interactive --production=true
```
koonopek, line 23: In best scenario we would avoid this step tottaly and everything would be bundle in single js file, however I have problem with native files `*.node`

rollup.config.mjs
```diff
@@ -0,0 +1,21 @@
...
+        file: 'dist/index.js',
+        format: 'cjs',
+    },
+    plugins: [
+        typescript({ composite: false }),
+        commonjs({
+            ignoreGlobal: false,
+        }),
+        json(),
+        obfuscator(),
```
koonopek, line 19: Smae obfuscator

packages/oracle-node/src/fetchers/kaiko-v2/KaikoV2Fetcher.ts
```diff
@@ -16,7 +16,7 @@ const KAIKO_PRICES_URL =
 const KAIKO_REQUEST_HEADERS = {
   "X-Api-Key": config.kaikoApiKey,
   Accept: "application/json",
-};
+} as Record<string, string>;
```
rpiszczatowski, line 19: Out of curiosity: what was the reason behind this change?

rollup.config.mjs
```diff
@@ -0,0 +1,21 @@
...
+    ]
+};
```
rpiszczatowski, line 21: empty line

.github/workflows/deploy-image-template.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Push docker image to ECR public
+        if: ${{ github.ref_name == 'main' }}
+        run: |
+          aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws
+          docker push $IMAGE:$TAG
+      - name: Print image details
+        run: |
+          docker history --no-trunc $IMAGE:$TAG
+          IMAGE_SIZE=$(docker inspect -f "{{ .Size }}" $IMAGE:$TAG | numfmt --to=iec --format='%.2f')
+          echo "image size ${IMAGE_SIZE}" >> $GITHUB_STEP_SUMMARY
```
rpiszczatowski, line 43:
```suggestion
          echo "image: $IMAGE:$TAG, size ${IMAGE_SIZE}" >> $GITHUB_STEP_SUMMARY
```

.dockerignore
```diff
@@ -4,6 +4,6 @@ node_modules
 packages/starknet-connector/build
 **/dist
 **/.DS_Store
-Dockerfile
+Dockerfile*
```
rpiszczatowski, line 7: Why?

rpiszczatowski (commented):
I really like the idea.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+RUN yarn install --frozen-lockfile --non-interactive --production=false
+
+COPY . .
+RUN cd "packages/${PACKAGE}" && echo $PWD && yarn install  --frozen-lockfile --non-interactive  && yarn build && yarn bundle 
+
+FROM node:16
+ARG PACKAGE
+
+COPY --from=build "/app/packages/$PACKAGE/package.json" yarn.lock
```
rpiszczatowski, line 20: typo. Should be 'yarn.lock' not 'package.json'

Dockerfile
```diff
@@ -0,0 +1,29 @@
+ARG PACKAGE="oracle-node"
```
rpiszczatowski, line 1: I'd get rid of a default value.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+FROM node:16 AS build
+ARG PACKAGE
+
+WORKDIR /app
+
+COPY package.json .
+COPY yarn.lock .
+COPY .yarnrc .
+
+RUN yarn install --frozen-lockfile --non-interactive --production=false
```
rpiszczatowski, line 12: What does this install actually without access to package.json from the component? DevDependencies only?

.github/workflows/deploy-image-template.yml
koonopek: done

.dockerignore
koonopek: Not needed, good catch!

Dockerfile
koonopek: Yep dev deps to use rollup and plugins

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+COPY . .
+RUN cd "packages/${PACKAGE}" && echo $PWD && yarn install  --frozen-lockfile --non-interactive  && yarn build && yarn bundle
```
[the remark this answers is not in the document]
koonopek: good point

Dockerfile
```diff
@@ -0,0 +1,29 @@
+ARG PACKAGE
```
cehali, line 1: Do we need this initial `ARG PACKAGE`?

koonopek: Nope ;)

rollup.config.mjs
```diff
@@ -0,0 +1,30 @@
...
+        // bundle only @redstone-finance modules
+        // rest import from node_modules
+        if (id.includes('node_modules') && !id.includes('@redstone-finance')) {
+            return true;
+        }
+        return false;
+    },
+    input: process.env['ENTRYPOINT'],
+    output: {
+        file: 'dist/index.js',
```
rpiszczatowski, line 18: Not sure how it works, but here it looks to me that our single output file is `dist/index.js` but in Dockerfile we have:
```
ENTRYPOINT [ "node", "index.js" ]
```
Do those work together?

packages/oracle-node/package.json
```diff
@@ -8,8 +8,9 @@
   "prettier": "redstone-prettier-config",
   "scripts": {
     "build": "rm -rf dist && tsc --build tsconfig.build.json",
+    "bundle": "ENTRYPOINT=index.ts rollup -c ../../rollup.config.mjs",
     "start": "ts-node index.ts",
-    "start:prod": "node dist/index.js",
+    "start:prod": "node index.js",
```
rpiszczatowski, line 13: Should this change? It looks to me that `yarn build` does the same thing as previously, so index.js is probably in dist folder, right?

rpiszczatowski (dismissed):
a couple of questions left.

rollup.config.mjs
koonopek: In dockerfile we copy everything from dist, but without dist

rpiszczatowski (approved):

cehali (approved):
Looks very good 👍

koonopek:
I want to test it on tf before merging

PR #1200, opened 2023-10-04
chore: add decimals for YY_WOMBEX tokens

cehali:
This change is at the request of Delta Prime. It should be deployed on the nodes with close collaboration as they need to revert some changes on their side.


rpiszczatowski (approved):

koonopek:
We have hardcoded 8 decimals in many places, this change requires testing for sure before using in production

koonopek (approved):

PR #1202, opened 2023-10-04
tsconfig cleanup

rpiszczatowski:
We want to:
- specify all the .ts and .js and optionally .json files included in tsconfig.json so that all of them are nicely linted in IDE while writing the code
- not use .eslintignore files as the are tricky to be handled by IDE (problems with files on different levels of directory tree)
- exclude everything not required in runtime in tsconfig.build.json

If not specified:
- include section contains [**/*]
- exclude section contains [node_modules, $outDit]

'*' expands to all .ts and .js files (and variations)
Paths are relative to tsconfig.json file directory.

The default behavior is good enough usually.
The only problem is that *.json files if required need to be added separately.


packages/data-analytics/tsconfig.json
```diff
@@ -3,13 +3,5 @@
...
-  "ts-node": {
-    "files": true
-  },
-  "include": [
-    "src/**/*",
-    "src/**/*.json",
-    "./package.json",
-  ],
-  "exclude": ["node_modules"],
+  "include": ["**/*", "package.json"]
```
koonopek, line 6: so we will aslo include node_modules?

packages/eth-contracts/tsconfig.json
```diff
@@ -3,6 +3,5 @@
   "compilerOptions": {
     "outDir": "dist"
   },
-  "include": ["./src", "./test", "./typechain-types", "package.json"],
-  "files": ["./hardhat.config.ts"]
+  "include": ["**/*", "package.json"]
```
koonopek, line 6: 1) so we will aslo include node_modules?
2) **/* doesn't include package.json already?

packages/fuel-connector/tsconfig.build.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "./tsconfig.json",
-  "exclude": ["node_modules", "test", "dist", "**/*spec.ts"],
+  "exclude": ["node_modules", "dist", "test", "**/*spec.ts", "jest.config.ts"],
```
koonopek, line 3: Why we are sometimes excluding something twice like test but also **/*spec.ts

packages/fuel-connector/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "dist"
   },
-  "include": ["./src", "./test", "./src/autogenerated/*.bin"],
+  "include": ["**/*", "src/autogenerated/*.bin"],
```
koonopek, line 6: "**/*", 
   ^ this set already includes "src/autogenerated/*.bin" or am I wrong?

yarn.lock
```diff
@@ -19802,6 +20019,49 @@ web3-utils@^1.3.6:
...
+  resolved "https://registry.yarnpkg.com/web3-validator/-/web3-validator-2.0.2.tgz#c0d411371c5ab75806c8ad958276171678b8ddb9"
+  integrity sha512-9sQ5owd2UldTsva3o3htj2fTPpbmUwb4TfBXhjIkew8FyT0ss3DPI+j3p6XrfdxIVBEQ5r17YUakElDV99aW+A==
+  dependencies:
+    ethereum-cryptography "^2.0.0"
+    util "^0.12.5"
+    web3-errors "^1.1.2"
+    web3-types "^1.2.0"
+    zod "^3.21.4"
+
+web3@^4.1.2:
```
koonopek, line 20043: Why we are isntalling this whole web3 thingy?

koonopek (commented):
Left some questions/comments most of them can be applied to every tsconfig, but I didnt want to repeat mysefl

packages/data-analytics/tsconfig.json
cehali: I think we shouldn't include `node_modules` in both `tsconfig` and `tsconfig.build`

packages/evm-connector/tsconfig.json
```diff
@@ -3,15 +3,7 @@
...
-  "include": [
-    "./src",
-    "./test",
-    "./typechain-types",
-    "package.json",
-    "./contracts",
-    "./hardhat.config.ts",
-    "benchmarks"
-  ],
+  "include": ["**/*", "package.json"],
```
cehali, line 6: Maybe we can extract `"include": ["**/*"]` to `tsconfig-common.json`  as it seems to be enough for most of the cases.

cehali: Or maybe we can narrow it down to `["src", "test", "**/*.json", "hardhat.config.ts", "typechain-types"]` in `tsconfig-common.json`. I don't know why we should include f.g. contracts

packages/data-analytics/tsconfig.json
rpiszczatowski: `include` by default (and also here) includes 'node_modules'. But 'node_modules' is excluded by default `exclude` value.

Changed behavior with regards to 'node_modules' inclusion is expected (it is not included) and it is the same as if there was no 'include' or 'exclude' properties defined.

packages/evm-connector/tsconfig.json
cehali: Even better: `"include": ["**/*.ts", "**/*.json", "**/*.js"]`

packages/eth-contracts/tsconfig.json
rpiszczatowski: See my comment to this review. * wildcard for file names and also directory names only include .ts, .js and relatives (source maps, etc.), not .json, .bin or other.

packages/fuel-connector/tsconfig.build.json
rpiszczatowski: There exists a well-known approach of placing test files along with the source files that they test. On the other hand 'test' directory may contain other files than *spec.ts files.
Excluding both supports both approaches. If we are sure we always put spec.ts files in test folder we can simplify exclude section further.

Should I do it?

packages/fuel-connector/tsconfig.json
rpiszczatowski: see above.

yarn.lock
rpiszczatowski: Because it is used in data-analytics package. Without it data-analytics is red in vscode.

packages/evm-connector/tsconfig.json
rpiszczatowski: We cannot because there is no way to use "include" from parent and add something to it.
Also, whenever we don't need anything more than default we simply skip 'include' entry altogether.

When we want to add something we need to repeat the default value.
This is very similar to 'exclude' entry. If you want to exclude something non-standard you have to repeat 'node_modules' and $outDir.

packages/fuel-connector/tsconfig.build.json
koonopek: nope, fair point

koonopek (approved):

packages/fuel-connector/tsconfig.build.json
cehali: To be honest I think we have a consensus that we add tests only in the test folder, not along with source files. Also, even if we have additional files in the test folder, they shouldn't be in the build version. I think we should stick to the test folder in exclude just to have in as cleanest as possible

packages/evm-connector/tsconfig.json
cehali: Ok, I think I get it now. One more question. Could we have `["**/*.ts", "**/*.json", "**/*.js"]` in all tsconfigs instead of `**/*`?

PR #1205, opened 2023-10-05
feat: allow for up to 3H old prices while calculating slippage

rpiszczatowski:


packages/oracle-node/test/fetchers/uniswap-v2-like.spec.ts
```diff
@@ -88,4 +95,22 @@ describe("UniswapV2Like", () => {
       },
     ]);
   });
+
+  test("Should throw on stale base price", async () => {
```
cehali, line 99: I don't think this is the best place to test it. We have `local-db.spec.ts` for testing local db

packages/oracle-node/test/fetchers/uniswap-v3-on-chain.spec.ts
```diff
@@ -142,4 +142,21 @@ describe("uniswap V3 fetcher", () => {
       },
     });
   });
+
+  test("Should throw on stale base price", async () => {
```
cehali, line 146: I don't think this is the best place to test it. We have `local-db.spec.ts` for testing local db

packages/oracle-node/test/fetchers/uniswap-v2-like.spec.ts
rpiszczatowski: I agree when we talk about unit testing the function that throws an error if the price is too old, but honestly, it isn't complicated enough to deserve its own tests. I intentionally wanted to test it in two contexts in which it is used.

Uniswap v2 calls that method in `calculateSlippage` and other DEX fetchers call it in `makeRequest` already. I wanted to verify that in both situations we correctly won't return the price.

cehali: But this test has nothing to do with `uniswap-v2`, it could be named `should return an empty array if an error happened`, we don't check what error happened. It would be really hard to understand in the future why this test is here and what it checks.

rpiszczatowski: Really? Maybe I can adjust the title, but even now it says that the price shouldn't be returned if the available token price is staled.

packages/oracle-node/src/db/local-db/index.ts
```diff
@@ -169,29 +170,27 @@ export const clearLastPricesCache = () => {
...
-  getPrices,
-  clearPricesSublevel,
-  closeLocalLevelDB,
-  getLastPrice,
-  clearLastPricesCache,
-};
+export const getRawPrice = (symbol: string) =>
+  getRawPriceNotOlderThan(symbol, MAX_PRICE_IN_DB_TIME_DIFF);
+
+export const getRawPricePotentiallyStalled = (symbol: string) =>
```
koonopek, line 195: I think that is function is meant to be used only for calculating slippage maybe lets put it in slippage_commons (we have such a file already). And rename it to `getRawPriceForSlippage`. Let's be specific.

packages/oracle-node/test/fetchers/uniswap-v2-like.spec.ts
koonopek: I agree with Cezary we are coupling concrete fetcher with general fetchers logic.
What if we remove uniswap-v2-like-fetcher, this means that we also remove this feature?
Maybe we can assure this on OnChainDexFetcher level?
Or add such a test in every class which implements OnChainDexFetcher

rpiszczatowski: I'll just add tests for SlippageAndLiquidityCommons.

koonopek (approved):
Nice!

PR #1208, opened 2023-10-05
feat: Add timeouts for single operations

koonopek:


cehali (approved):

packages/rpc-providers/src/ProviderWithFallback.ts
```diff
@@ -12,7 +12,8 @@ import { RedstoneCommon } from "@redstone-finance/utils";
 
 const logger = Logger.globalLogger();
 
-export const PROVIDER_OPERATION_TIMEOUT = 30_000;
+export const PROVIDER_OPERATION_TIMEOUT = 20_000;
```
cehali, line 15: We want to decrease it?

koonopek: yep

PR #1210, opened 2023-10-05
feat: first check for nonce error then for price

koonopek:
There exists rare case:
1. Broadcast tranaction [nonce: 1]
2. Wait 7 seconds [nonce: 1]
3. Check if nonce increase [nonce: 1] => not increads
4. Bump fees [nonce: 2]
5. Send transaction wrong nonce should be 3 

In this edge case updates happens just between 3 and 5 step.
Surprisingly, this case is happening very often on kava (9/15 cases) chain don't know why. On bnb it is 0.
Maybe because 7 seconds is too close to the average block size which is 6.2 seconds.

Nevertheless, these changes are still relevant.
I will also increase EXPECTED_DELIVERY_TIME to 10seconds


packages/rpc-providers/src/TransactionDeliveryMan.ts
```diff
@@ -230,7 +230,6 @@ export class TransactionDeliveryMan {
       (e.message.includes("maxFeePerGas") ||
         e.message.includes("baseFeePerGas") ||
         e.code === ErrorCode.INSUFFICIENT_FUNDS ||
-        e.code === ErrorCode.SERVER_ERROR ||
```
koonopek, line 233: This was a little bit to broad. It may happen that we will have to edit this lists when adding new chains, which have different error messages like kava do

rpiszczatowski (approved):

hatskier (approved):

PR #1214, opened 2023-10-06
chore: unify dotenv usage

rpiszczatowski:
`import "dotenv/config"` is a better way of using dotenv as it allows for external configuration, e.g. specifying .env file path in env variables. We used this feature e.g. in integration tests.


koonopek (approved):
Good change!

cehali (approved):

PR #1218, opened 2023-10-06
chore: disable sources consistency test in oracle node tests

cehali:
We want to use it in GitHub workflow, there is no need to run it twice


koonopek (approved):

packages/oracle-node/jest.config.ts
```diff
@@ -9,7 +9,11 @@ const config: Config.InitialOptions = {
   testEnvironment: "node",
   modulePaths: ["<rootDir>"],
   setupFiles: ["<rootDir>/.jest/set-env-vars.js"],
-  testPathIgnorePatterns: ["<rootDir>/test/dry-run/*", "<rootDir>/dist/*"],
+  testPathIgnorePatterns: [
+    "<rootDir>/test/dry-run/*",
+    "<rootDir>/dist/*",
```
rpiszczatowski, line 14: Why is dist required here? @koonopek ? We shouldn't have tests in dist in the first place.
On the other hand, we should probably add 'node_modules' as `["node_modules"]` is the default value for this variable. At the moment, thankfully, there are no stray tests in our node_modules inside oracle_node, but there are some in top-level node_modules

```suggestion
    "<rootDir>/node_modules/*",
```

cehali: Jest wouldn't run node_modules tests even when it is not specified in `testPathIgnorePatterns`: https://github.com/jestjs/jest/issues/2145

PR #1235, opened 2023-10-12
feat: do dot deliver prices to mento sorted oracles when deviation ex…

rpiszczatowski:
…ceeds defined threshold


packages/on-chain-relayer/src/types.ts
```diff
@@ -61,6 +64,7 @@ export interface RelayerConfig {
   cacheServiceUrls?: string[];
   isAuctionModel?: boolean;
   historicalPackagesGateways?: string[];
+  mentoMaxDeviationAllowed?: number;
```
cehali, line 67: How are we going to provide `mentoMaxDeviationAllowed` value if it isn't declared in `FilesystemConfigProvider`?

rpiszczatowski: good point.

cehali: It is more our safety tool, not a commitment for the classic model so I would say env variables

PR #1241, opened 2023-10-13
chore: simplify ssh access configuration to public monorepo

rpiszczatowski:


.github/workflows/sync-to-public-repo.yml
```diff
@@ -14,16 +14,12 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-
-      - name: Sync to target repo respecting .gitignore
-        env:
-          PRIVATE_KEY: ${{ secrets.PUBLIC_REPO_KEY }}
+      - name: Setup public monorepo access
+        uses: webfactory/ssh-agent@v0.8.0
```
cehali, line 18: I would be careful with external tools using our secrets. I am wondering if leak of the public repo key could be harmful for us

koonopek: I Agree with cehali, also In my opinion introducing an external tool instead of 4 lines of bash is complicating stuff.

koonopek (changes requested):

.github/workflows/sync-to-public-repo.yml
rpiszczatowski: It isn't complicating anything in my opinion. And I've read the code of that tool (it is less than 100 lines). It adds one feature over what we had ability to add keys for more than one github repo + couple of configuration options like specifying git and ssh binaries

rpiszczatowski: We are going to use accessing private repositories more now and I think it is nice not to write bash voodoo in all the places. Not that I'm not a fan of bash voodoo (because I am a little) but do we really want everybody to understand why `eval $(ssh-agent -s)` is required in this 'simple' bash script? Or do we want people to copy-paste the code without understanding what it is doing?

rpiszczatowski: I've imported action code to our repository, so that in can be reused. I agreed to this with Alex

.github/workflows/sync-to-public-repo.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - robal/**
```
rpiszczatowski, line 7: for testing, to be deleted before merge

koonopek (dismissed):
Nicee!

